PR #28, opened 2014-12-09
Replaced exterior_edges() to boundary_facets() as suggested in the comment

stefanbrugger:


alecjacobson:
Sorry this took so long. Thanks again.

PR #32, opened 2015-01-22
Removed flipping of texture coordinates in writeOBJ(...)

stefanbrugger:
Thats a bug, isn't it? At least it is not consistent with readOBJ(...).


alecjacobson:
This bug fix (and it is a bug fix) will likely break code using textures currently, though at least it's happening upon writing and not reading. Thanks for finding this.

PR #34, opened 2015-01-26
Added overload that returns barycentric coordinates of hit

stefanbrugger:
Added overload that returns barycentric coordinates of hit instead of vid of closest vertex of hit


alecjacobson:
Thanks. Good idea.

PR #53, opened 2015-04-03
Example makefiles now use eigen global configuration variables.

louen:
Some examples' makefiles would ignore eigen directories configuration variables  defined in Makefile.conf. These commit use the global Makefile.conf variables if they are defined, and resort to the previous behaviour if they are not defined.


alecjacobson:
Looks like we both tried to make the same change at the same time. Please `git pull` and try the new makefiles, they'll use `../buildMakefile.conf` to define things like eigen3_inc

Issue #62, opened 2015-04-17
warning: using integer absolute value function 'abs' when argument is of floating point type


chrisidefix:
I am receiving warnings when compiling the tutorials, which might not cause any problems, but allow me to commit a fix anyway.

Inside the `trackball.cpp` my compiler complains that `abs` is used with floating point input, where you probably want to use `std::abs`.

Maybe the cast to `int` is intended here, but it should then be done explicitly. Otherwise, please feel free to merge my pull request.

PR #63, opened 2015-04-17
using std::abs() instead of abs() for double input (issue #62)

chrisidefix:
c math.h abs() function will cast input to int, you probably want to use std::abs() for floating point input 
as discussed in issue #62

### DIFF
--- a/include/igl/trackball.cpp
+++ b/include/igl/trackball.cpp
@@ -18,6 +18,8 @@
 #include <algorithm>
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


alecjacobson:
Thanks for catching that. I've pushed a fix. I cannot merge yours because we _never_ use `using namespace` in the global scope.

chrisidefix:
Great - makes sense - I wasn't aware this would be global scope, but you are absolutely right.

PR #67, opened 2015-04-21
FindLIBCOMISO include directory fix

chrisidefix:
FindLIBCOMISO.cmake re-defined (expanded) the variable LIBCOMISO_INCLUDE_DIR every time it was called by adding a sub-directory. This leads to a "recursive" expansion of LIBCOMISO_INCLUDE_DIR, which was probably not intended. Instead I would recommended to write such additional include directories into a new variable (LIBCOMISO_INCLUDE_DIRS) and use this instead. I also added additional search paths for finding the library.


danielepanozzo:
Thank you very much for finding this bug and for the fix!

Issue #60, opened 2015-04-17
AntTweakBar: OpenGL Core Profile functions cannot be loaded.


chrisidefix:
Let me first say that I would like to use the "original" AntTweakBar version, which I installed through homebrew. If you think this is not a good idea, please share your thoughts.

I am compiling some of the tutorials and received this error from AntTweakBar:

    AntTweakBar: OpenGL Core Profile functions cannot be loaded.
    ERROR(AntTweakBar) >> Cannot load OpenGL library dynamically

It is apparently due to the OpenGL library not being found and can be fixed manually by adding the path to your OpenGL library to your linker path. In my case under Mac OS X this can be accomplished like this:

    export DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:/System/Library/Frameworks/OpenGL.framework/Versions/Current/

However, it may be a good idea to include the OpenGL lib path into the `link_directories` in your `CMakeLists.shared` file.

Credit for this blog-post to point me into the right direction: [http://www.pixelnerve.com/v/2015/01/22/anttweakbar-with-core-gl-osx](http://www.pixelnerve.com/v/2015/01/22/anttweakbar-with-core-gl-osx/)

chrisidefix:
[UPDATE]
I just saw that you already reference `${OPENGL_LIBRARIES}` in your `SHARED_LIBRARIES` variable, which you then write into the `target_link_libraries`. If I do `otool -L` on the compiled executable I can also see the OpenGL library with full path. I guess the issue is solely with `AntTweakBar` wanting to load OpenGL dynamically at run-time and can not be fixed through CMake.

It is weird, though, since I also ran `otool -L` on my `AntTweakBar` library and it correctly references OpenGL. This is strange.

    otool -L 102_DrawMesh_bin:
    /System/Library/Frameworks/OpenGL.framework/Versions/A/OpenGL (compatibility version 1.0.0, current version 1.0.0)
    /System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa (compatibility version 1.0.0, current version 21.0.0)
    /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
    /System/Library/Frameworks/CoreVideo.framework/Versions/A/CoreVideo (compatibility version 1.2.0, current version 1.8.0)
    /System/Library/Frameworks/AGL.framework/Versions/A/AGL (compatibility version 1.0.0, current version 1.0.0)
[...]

PR #70, opened 2015-04-28
adding AntTweakBar src to LIBIGL include directories

chrisidefix:
`igl/viewer/TextRenderer.h` contains `#include <TwOpenGLCore.h>`, which is not distributed inside the AntTweakBar include directory. To use FindLIBIGL.cmake in other projects, this allows them to find the required header. (issue ref: #60 )

### DIFF
--- a/tutorial/cmake/FindLIBIGL.cmake
+++ b/tutorial/cmake/FindLIBIGL.cmake
@@ -22,7 +22,10 @@ FIND_PATH(LIBIGL_INCLUDE_DIR igl/readOBJ.h
 
 if(LIBIGL_INCLUDE_DIR)
    set(LIBIGL_FOUND TRUE)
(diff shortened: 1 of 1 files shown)


alecjacobson:
The AntTweakBar dependency in the Viewer is a mess right now that we're
trying to sort out. While this would be a temporary solution, I fear it
would prolong these troubles.

chrisidefix:
Sure, this fix might be temporary, but for now it allows people to use `CMake` with `libigl`, which simply doesn't work right now, unless you manually include that AntTweakBar `src` directory inside your own `CMake`.

alecjacobson:
As of roughly f03c6f1 the tutorial uses nanogui instead of AntTweakBar and the TextRenderer.h is obsolete.

PR #71, opened 2015-04-29
adding writeOBJPoly aligned with readOBJPoly

chrisidefix:
Added a writeOBJPoly function to support writing out mixed element meshes (triangles, quads, polygons, ..), which is comparable to readOBJPoly for reading such meshes.


alecjacobson:
Thanks, but wait a second. This seems like it will only work if all of the polygons are the same degree (all triangles or all quads or all pentagons etc.) Since `readOBJpoly` supports arbitrary an polygon mesh, probably `writeOBJpoly` should, too.

chrisidefix:
Yes, sorry it was a bit late last night and I guess I didn't commit everything - I'll fix it asap.

chrisidefix:
Ok, so I didn't change the simplified code you checked-in the other day for `writeOBJ(V,F)`, but I thought theoretically it would be easier to maintain, if the `writeOBJ` functions just became wrappers for `writeOBJPoly`. As you can see I am using `matrix_to_list` to convert the relevant input inside `writeOBJ(V,F,CN,FN,TC,FTC)` to `std::vector` objects and then pass it along to `writeOBJPoly(V,vF,CN,vFN,TC,vFTC)`.

Lastly, I haven't looked at how to generate the explicit instantiations properly. That's something left to do.

alecjacobson:
Thanks for trying again. I hesitate to agree. Using `matrix_to_list` makes the eigen code potentially slower, but you're right about simplifying the code...

I can't merge this until you test it though. For example, just quickly looking it over this line looks suspicious:

    for(int c =0;c<(int)F.size();++c) {

should probably be

    or(int c =0;c<(int)F[i].size();++c) {

Finally, I think we can drop the `poly` from `writeOBJpoly` (and `readOBJpoly`, but I'll take care of that) since the arguments disambiguate those from the others.

chrisidefix:
Sure, I guess the question is - what do you prefer - redundant code that you have to edit twice in case you would like to make a change or faster execution time...

I did test the code, but my test case contained 4 faces ... :1234: and therefore worked, despite of this obvious mistake.

alecjacobson:
For now, let's use the matrix_to_list version. I did a quick benchmark and
it doesn't make _that_ much of a difference. I guess ideally we could write
a functional version that provides a "matrix" (either Eigen:Matrix* or
vector<vector> ) as an input argument and an `at(i,j)` function handle as a
_template_ argument. But that's a little complicated for now.

Please do test on many example meshes beforehand.

Thanks!

chrisidefix:
Yes, I will do that - thanks for checking my code all the time - I just tested with a generic mesh containing 6'000'000 faces (712 MB).

`readOBJ` ~ 53 seconds
`writeOBJ` ~ 28 seconds
`readOBJPoly` ~ 49 seconds
`writeOBJPoly` ~ 20 seconds

alecjacobson:
This will be affected/fixed by an upcoming major release.

PR #74, opened 2015-05-10
write object bug: need newline between end of v and start of f

fmardini:
Hi,
Small bug with serialization into obj format


alecjacobson:
Ah. Thanks, I found this in writeOFF but didn't think to also fix it in writeOBJ.

PR #101, opened 2015-08-31
Exterior vertex/edge/facet

qnzhou:
Added ``exterior_vertex()``, ``exterior_edge()`` and ``exterior_facet()`` support:

* ``exterior_vertex()`` returns a vertex reachable from infinity without crossing any face.
* ``exterior_edge()`` returns a edge reachable from infinity without crossing any face.
* ``exterior_facet()`` returns a face reachable from infinity without crossing any face.

The goal of ``exterior_facet()`` is similar to ``outer_facet()``, but using a different algorithm.  The algorithm used in ``outer_facet()`` returns incorrect result for certain cases ([Counter example](https://www.dropbox.com/s/zaa2nh5q4g1bvo3/pipe.obj)).  ``exterior_facet()`` uses a more rigorous algorithm that handles this case.  It also comes with two sibling methods ``exterior_vertex()`` and ``exterior_edge()`` just in case they are needed by other projects.


alecjacobson:
This name might get confused with the existing `exterior_edges` which
computes something entirely different. Probably these new functions are
more correctly named, but perhaps rename them `outer_*` and replace the old
outer_facet.

qnzhou:
name changed to outer*.

PR #116, opened 2015-09-15
Unit tests

qnzhou:
OK, here is another pull request.  This pull request contains my unit test setup for libigl.  It is similar to the setup in [PyMesh](https://github.com/qnzhou/PyMesh).  To compile and run all unit tests:

    cd $LIBIGL_PATH/tests
    mkdir build
    cd build
    cmake ..
    make # make automatically runs unit test after compilation

I put all static data needed by unit tests in ``tests/data/``.  The compiled unit test executables are located in ``tests/bin/``.


alecjacobson:
Depending on how big the `tests/data` directory gets, it might make sense to have this as a separate repo (then non-developers don't have to clone a huge amount of test data).

jdumas:
Just a quick note about gtest, you can still use the command add_subdirectory with arbitrary folders as long as you specify a destination folder for the binaries to be built in. For example, if  the gtest sources are in folder pointed by `$ENV{GTEST_ROOT}`, you can use the following line :

```
Add_Subdirectory($ENV{GTEST_ROOT} ${CMAKE_CURRENT_BINARY_DIR}/gtest)
```

qnzhou:
@jdumas I see.  I was using ``${CMAKE_BINARY_DIR}`` as the destination folder, and this generates the following warning:

```
CMake Warning (dev) at Settings.cmake:50 (ADD_SUBDIRECTORY):
  Policy CMP0013 is not set: Duplicate binary directories are not allowed.
  Run "cmake --help-policy CMP0013" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
```

Using ``${CMAKE_CURRENT_BINARY_DIR}/gtest)`` or ``${CMAKE_BINARY_DIR}/gtest`` seems to fix this problem.

PR #118, opened 2015-09-16
Outer facet bug fix

qnzhou:
As discussed in pull request #116 , google unit test submodule is moved to the external folder.

Also fixed the bug due to wrong ``flip`` value in outer facet.


alecjacobson:
Could you make sure that the old ay module folder is deleted and verify
that the scripts/clone_and_build.sh executes correctly? The last pull
request broke the submodules...

qnzhou:
I have verified that the old google test directory is deleted, but I am
having a lot of trouble getting embree (required by the tutorial) to
compile properly using the script.  I got errors about invalid gcc-4.9.3
compiler option ``-stdlib=libc++``.  When I switch to clang-3.6.0, the
compilation failed because undefined symbol ``xgetbv`` not defined.

alecjacobson:
eek. Is this a new problem? Is it related to GTest? If it's unrelated, then
please post a github issue with the errors, perhaps Daniele will have some
ideas.

qnzhou:
Do you compile and install Embree beforehand or use the version checked out from github?

It is not related to GTest, maybe we can merge it first.

PR #132, opened 2015-10-10
Enable out of source tutorial build

decrispell:
If the tutorials are built out of source, the paths to the data files in tutorial/shared will be incorrect.  This pull request fixes that by using CMake's "configure_file" command to store the correct absolute location of the directory.   (Note that this may cause issues if the tutorial binaries are ever packaged up for distribution.)


alecjacobson:
I like the idea, but I'm not wild about defining `tutorial_shared_path` in an _automatically generated_ header file. I think a preprocessor macro would be more appropriate: `-DTUTORIAL_DIR=/path/to/libigl/tutorial` and then as a fallback inside each `*/main.cpp` something like:

```cpp
#ifndef TUTORIAL_DIR
// assume default build from within tutorial/build/
#  define TUTORIAL_DIR "../"
#endif
```

decrispell:
What is it about the automatically generated header file that you don't like?  Would you feel better about it if TUTORIAL_SHARED_PATH was declared as a cmake "cache" variable, enabling users to set it to an alternative location if they needed to?  One advantage of the auto-generated file approach is that the configuration is isolated to one location, and you don't have to repeat the preprocessor code in every tutorial main.cpp.

Thanks for the fantastic library by the way - very useful and the tutorial is great.

alecjacobson:
I agree it's a little sloppy to repeat the preprocessor code in each cpp.
Instead we could include a _static_ (as in not automatically generated)
header containing the default define of the path. (I'm still slightly in
favor of just duplicating the code since it keeps the examples more stand
alone).
Generating new files is adding another layer to the compilation process and
getting farther away from  straight C++, putting a heavy dependence on
cmake. Unless it were absolutely necessary I'd like to avoid that.

decrispell:
Fair enough - I've added a commit to the pull request that removes the auto-generated include file, opting for the preprocessor definition and (non auto-generated) include file as per your suggestion.  Converting the preprocessor token into a string constant within the source code requires a little bit of macro ugliness, but at least it's isolated to the single header file.

alecjacobson:
Cool. This is getting there. I don't think the `STRINGIFY` macros are really needed or the extra `const std::string tutorial_shared_path` variable.

In `tutorial_shared_path.h`, you could just have:

```cpp
#ifndef TUTORIAL_SHARED_PATH
#define TUTORIAL_SHARED_PATH "../shared"
#endif
```

And then in the `*/main.cpp` files, something like (notice the space in between rather than `+`):

```cpp
igl::readOFF(TUTORIAL_SHARED_PATH "/cube.off", V, F);
```

Defining a string literal from the command line (or `cmake`) should be no problem if the quotes are escaped.

For example, here's a tiny test program you could save in `test.cpp`:

```cpp
#include <string>
#include <iostream>
int main()
{
  std::cout<<PATH "mesh.obj"<<std::endl;
}
```

If you compile (on Mac OS X with clang or gcc) with:
```bash
g++ -o test test.cpp -DPATH="\"/path/to/libigl/tutorial/shared/\""
```

Then if you run with `./test`, you should see as output:
```bash
/path/to/libigl/tutorial/shared/mesh.obj
```

Thanks again.

decrispell:
Done.  The tutorial shared path is now a #define, so no stringify macros or static string variable.  I didn't do it this way initially because I didn't like having to add the quotes around the path passed to the -D argument, but I agree this way is still a little cleaner and probably the lesser of two evils.

alecjacobson:
Thanks, Daniel!

decrispell:
No problem - Thanks again to all the devs for a great library!

PR #149, opened 2015-11-12
Winding number propagation

qnzhou:
Pull request summary:

* Bug fix in order facets around edge.
* Bug fix in resolving intersection due to CDT ambiguities.
* Add function to extract manifold patches from mesh.
* Add function to extract cells partitioned by the input mesh.
* Add function to extract non-manifold edge curves from input mesh.
* Add function propagate winding number from infinity into the cells.
* Add function to peel winding number layers.
* Update boolean operations to be based on winding number.


alecjacobson:
What's the difference between `extract_manifold_patches` and the existing `orientable_patches`?

qnzhou:
I guess there is no difference besides the interface.  We could keep just
one of them.

alecjacobson:
If the mesh is a moebius strip what will extract_manifold_patches do?

qnzhou:
It would extract just a single patch.

alecjacobson:
I think orientable_patches will do the same. I'll make a "to do" to merge these. For now I'm still plugging away at getting this pull request to compile...

qnzhou:
Ahhh sorry :(  I thought I have it compiling.  Is it the tutorial that
failed to compile?

alecjacobson:
Yes, the version with the static library. I'll fix it this time and I'm creating instructions for future pull requests.

alecjacobson:
I've started combing out these implementations a bit. Some documentation is still missing and there are many "style" issues (see https://github.com/libigl/libigl/blob/master/style-guidelines.md): floating helper functions, classes and naming conventions, etc.

qnzhou:
ok, I can go through the files and clean up the style issues.

If encapsulation in a separate file is not possible or does not make sense,
> then avoid crowding the namespace by creating lambda functions within the
> function implmentation.


So the usage of lambda functions within a function is also discouraged?

alecjacobson:
Thanks.

No, that means that we'd rather have lambdas than floating single-use
helper functions. Though, neither are great.

PR #150, opened 2015-11-12
Start of Polygonal Support

avaxman:
so far:

1. reading and writing polygonal files with the (D,F) scheme, meaning that there is a column vector D with the degrees of faces, and matrix F (|F| x max(D)) with the vertices themselves. everything F(i,j) where j>D(i)-1 is -1. 

2. general_edge_topology to complement edge_topology, just for this representation.


alecjacobson:
Thanks, Amir. I'll try to review this soon. Probably we need to discuss internally best practices with general polygonal meshes.

BTW, I didn't even know that .off supported polygon meshes...

avaxman:
Hi Alec,

Yes, of course, OFF has a face structure which is exactly this (D,F) thing.

reading and writing is not a problem, and also I intend to put some
auxiliary functions (triangulation, measure planarity, cross-ratios,
concyclity, etc). Finally, the ideal plan is to put the Mobius paper
(SIG2015) and the affine paper (SGP2012) in, at least. I'm not sure what
you guys are currently doing with quad remeshing papers for representation.

We need to discuss the integration to the viewer. What I am currently doing
is triangulating in my code and handing triangles to the viewer. This is a
bit disadvantageous since there is the wireframe problem: the user should
probably see the true wireframe as the regular option, and the triangulated
one as a secondary option because who cares. I suggest a special
"set_data(V,D<F)" instant that puts the proper edges in.

Amir.

alecjacobson:
Sounds like a good plan. I'm always weary of beefing up the viewer since it's really only exists in libigl as a means to create tutorials. Which is to say, if this feature's necessary for showing off a cool new feature in the main library then great. But if it's just a nice rendering feature then I'd be against adding it and complicating the viewer.

avaxman:
I guess pure wireframe should be default behavior, and showing
triangulation is really optional if at all needed.

PR #151, opened 2015-11-15
Style fix and refactor

qnzhou:
Summary:

* Style fix for mesh_boolean.*
* Bug fixed in face tracking in boolean.
* Refactor ``resolve_duplicated_faces()`` into its own file.
* Changed tutorial ``CMakeLists.txt`` to not build embree from source due to bugs in Embree's cmake system.  Instead, use existing embree library in the system if available, otherwise skip embree dependent tutorials.


alecjacobson:
Somehow the merge didn't register your commits. Can you verify that your changes were added?

alecjacobson:
Tried it again. Seemed to work. Let's independently run checks to see that it went through OK.

PR #154, opened 2015-11-19
Removing two cases of deprecated DynamicSparseMatrix

thedrakes:


alecjacobson:
We've been purposefully keeping the use DynamicSparseMatrix in these
functions because the newer Eigen implementations are slower. When I get
a sec, I'll review this again with the most current eigen. However,
if DynamicSparseMatrix still exists and is faster, we're going to keep it
(despite the warnings).

thedrakes:
I see, but then I wonder whether one should do a feature request for building a sparse matrix from a generator witch returns triplets presorted.

I guess I should rewrite it using reserve(nnzOfCol) if setFromTriplets is really less efficient even then DynamicSparseMatrix

PR #188, opened 2016-01-06
assert issue

romp13:
T are the faces so I think it should be 3 as documented in the header file.


romp13:
I actually think I am mistaking the usage of face_areas(). Is it for tet mesh only or is it for meshes?
Should I use doublearea for meshes? If yes then just you can just delete this pull request and fix the documentation in face_areas.h

Thanks

alecjacobson:
Yeah, this is for tet-meshes. You should use `igl::doublearea` and I've fixed the doc for `igl::face_areas` in 37f4f21a0b45b185dce82896b4b52f267b558c93 (poor naming). Thanks.

PR #237, opened 2016-03-31
python build instructions.

bdon:
First time using cmake - am I doing this the right way?


danielepanozzo:
Yes, thank you for the pull request. The python bindings are still experimental, it would be great if you could provide any feedback!

PR #249, opened 2016-04-25
Build fix for Microsoft Visual C++

e4lam:
When ACTIVE_SET_CPP_DEBUG is defined, the #warning preprocessor directive is used which
is not supported by Microsoft Visual C++.


danielepanozzo:
Thank you for the fix!

PR #253, opened 2016-05-04
Add material support for OBJ files

breiting:
This brings in support for reading and writing mtl information for OBJ files.


alecjacobson:
Cool! Give us a sec to look this over, but I'm excited to add this kind of
support.

breiting:
Hmm, it seems that the python stuff does not build anymore :-(

alecjacobson:
I've never really used .obj's material support. But can't a .mtl contain multiple materials? Is this not a common use case? I'm not saying we need perfect complaince, but would it make sense for `usemtl` to be split into a list of material names and a list of indices into `F` revealing the starting index of that material?

breiting:
You are right. It can contain multiple materials. It may make sense to keep this general. One other quick question. If the OBJ faces are negative, is there a function which reorders to positive face indices? I am currently struggle with remove duplicates with negative face indices :-(

alecjacobson:
Is there a correct behavior for negative OBJ face indices? Do you mean the
actual vertex indices are negative (less than zero?) or that the faces are
incorrectly oriented (normals point the wrong?).

On Thu, May 5, 2016 at 1:16 PM, Bernhard Reitinger <notifications@github.com

breiting:
The indices in the face list are negative e.g.

```
f -10/-10 -14/-14 -13/-13
```

alecjacobson:
I pushed a change to handle negative indices.

On Fri, May 6, 2016 at 1:50 AM, Bernhard Reitinger <notifications@github.com

metorm:
For obj file, I think [tinyOBJLoader](https://github.com/syoyo/tinyobjloader) is a good choice. The project is simple and provides good material support for OBJ files.

PR #255, opened 2016-05-10
Added missing templates to make static build work with gcc

s-koch:


alecjacobson:
Thanks! Could you possibly separate out the changes to the CMakeLists.txt ? Seems those comments are unrelated.

s-koch:
Hi Alec, they are not really unrelated. I added those as a note on how to fix these build problems, should they be moved somewhere else, e.g. before-submitting-pull-request.md?

alecjacobson:
Ah I see, now I understand. Do you think you could switch the CMakeLists.txt to use an `option(...)` flag so that by default these commands are not called, but at the command line the user can decide to flip them on? This would hopefully be a bit cleaner than commenting things in and out. Thanks.

s-koch:
Sure, that's better. Changed with latest commit.

PR #264, opened 2016-05-20
Interruptability for active_set

e4lam:
This pull requests add support for interruptability for active_set. By default, it does nothing but callers can pass in a Progress class to make active_set interruptable (ie. abort).


alecjacobson:
I think this is a good idea, but the Progress class seems a bit heavyweight/not-libigl-style. Couldn't it just be a function handle that gets called each iteration returning true/false whether to abort?

e4lam:
That sounds like a better idea. I'll try redoing these changes.

e4lam:
I thought about this some more but while we can make it simpler, it really can't be that much simpler if we want to ensure that the check is completely compiled out in the default case. The reason for igl::was_interrupted() is to ensure that in the default implementation, there will be no checks generated for interruptability. So we still need to past down some functor defaulted to nullptr and then have a wrapper function that specializes on the functor type to check whether the solver should abort.

If the concern is that there's too much machinery in Progress.h, then what do you think about something like this?

    struct NullAborter
    {
      bool operator() { return false; }
    };
    template <typename T>
    IGL_INLINE bool should_abort(T* i)
      { return i && (*i)(); }
    template <>
    IGL_INLINE bool should_abort<NullAborter>(T*)
      { return false; }```

Then active_set() is more or lass like before have a new template parameter that defaults to NullAborter, and takes an optional AborterT pointer.

alecjacobson:
This kind of rotted. Meanwhile I think the active_set code should probably be rewritten/revamped at some point. There's a lot of code in there that could be split off into smaller more interchangeable parts.

PR #269, opened 2016-05-25
Further python bindings, bugfixes

s-koch:
Tutorial 704 is now working in python (except the mesh overlay, see function append_mesh).


danielepanozzo:
Works for me in release mode, but an assert fails when I run the python script in debug mode.
Does it also happen for you on linux? If not let me know and I will debug it on mac.

Thanks for the contribution!

s-koch:
The problem should be fixed with my newest commit...

PR #272, opened 2016-05-30
minor fix to instruction string in viewer.cpp

evouga:
Current code has #ifdef inside string literal which does not get evaluated by the preprocessor


danielepanozzo:
Thank you for the fix!

PR #281, opened 2016-06-07
Restructured python bindings, cleaned up and added dependency checks to python tutorials

s-koch:


danielepanozzo:
This is really annoying:

g++-4.8: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.

Any idea of what could cause it?

s-koch:
It works with g++-4.8.5 for me. It seems like the machine is running out of memory when building with g++. Maybe switching of some optimizations could help. I'm just trying this.

s-koch:
Ok, so it looks like the memory problem comes from link time optimization. Can we increase the swap size of the server or omit lto by default?

PR #286, opened 2016-06-15
Convert the file string_to_mesh_booean_type.cpp to UTF8-BOM

metorm:
The special characters "∪","∩","∆" hinder the file to be compiled with msvc. For UTF-8 files which contain such special characters, only UTF-8 with BOM is allowed.


alecjacobson:
Thanks!

PR #290, opened 2016-06-20
Tutorial 609: Output model information (number of triangles) in order to estimate efficiency.

metorm:
The operation is relatively slow (5~6 seconded on normal computer). It is better to let the user know how many triangles the are working with.


alecjacobson:
Personally, I think a few seconds is OK to run without needing to clutter the output. The default models are small exactly for this purpose. Thanks anyway for the suggestion.

PR #299, opened 2016-07-11
add option for uniform laplacian operator in harmonic mapping function

hankstag:
- added a boolean `use_cotan_weight` to indicate use the cotangent weights or uniform weights laplacian operator. (`true` means use cotangent weights)

- also modified function calls in tutorial 401 402 501 503

### DIFF
--- a/include/igl/harmonic.cpp
+++ b/include/igl/harmonic.cpp
@@ -9,6 +9,9 @@
 #include "cotmatrix.h"
 #include "massmatrix.h"
 #include "invert_diag.h"
+#include "adjacency_matrix.h"
+#include "sum.h"
+#include "diag.h"
 #include "min_quad_with_fixed.h"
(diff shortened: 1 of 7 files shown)


alecjacobson:
This is a reasonable functionality to have but when `use_cotan_weights` is false then the vertex positions `V` is ignored.

So, how about instead of adding yet another parameter, rather add a new function overload with the prototype:

    igl::harmonic(F,b,bc,k,W);

since only `F` is an argument then it's clear from context that this is using the uniform laplacian.

hankstag:
Oh I see, it is more reasonable to use an overloading function. I will modify the code.

alecjacobson:
Hmmm. Seems the new code is ignoring the `k` parameter rather than "raising
the uniform laplacian to the kth power". I will have to clean it up.

PR #303, opened 2016-07-14
Minor speed up of ``unique_rows``

qnzhou:
Trying to squeeze some performance gain from the ``unique_rows`` method.  Summary of changes:

* Avoid call ``.size()`` ``.rows()`` ``.cols()`` repeatedly.
* Use lambda instead of comparison functor, which also avoids copying the array.
* Avoid range checking in comparison method.
* Avoid reversing the array, get it sorted correctly in the first run.

The performance gain is around 25% measured on my macbook air.


alecjacobson:
Nice. I had forgotten about how I originally implemented this. I think it was before I even knew about lambda functions. Thanks !

PR #322, opened 2016-08-11
Added loop subdivision functionality.

odedstein:
This is a modified version of the igl::upsample code that performs loop subdivision instead of simply upsampling. The coefficients and rules are taken directly from https://graphics.stanford.edu/~mdfisher/subdivision.html.


danielepanozzo:
Great, thank you!

PR #331, opened 2016-09-01
Updated matching code

olkido:
+ small other changes (old...)
Matching code has been adapted to match between non-line fields (i.e. fields where there are not necessarily any pairwise collinear points).
This mostly affects the recently added streamlines code.


alecjacobson:
Thanks for the pull request including the new tutorial entry.

- This is currently breaking the builds. Could you please take a look at the error messages above to see if they can be resolved before merging?
- Could you please add some text and hopefully a figure to tutorial.md to complement the new 511 tutorial example?

olkido:
Ok, corrected a couple things and added the tutorial text. Please check.

alecjacobson:
Excellent. Thanks!

PR #334, opened 2016-09-07
*fix bug: stl file is closed twiced when reading ascii

atismer:
I think the STL file was closed twiced. This occurs an errors in my case.


alecjacobson:
Good catch. Thanks

PR #340, opened 2016-09-20
Added R2016b to MATLAB versions

odedstein:
There is an error when running cmake for the MATLAB extensions with MATLAB R2016b at the moment. This should fix it.


alecjacobson:
Thanks, if only we knew what they were going to call the 2017 versions we could put them there in advance, too ;-).

PR #345, opened 2016-09-26
change "w" to "wb" in fopen to fix the binary dmat writing error

Zcynical:
i found an error in writing binary dmat file.
i tried change "w" to "wb" in fopen and it works.


Zcynical:
btw, my platform is windows 10, and compiled with vs2015 64bit

PR #349, opened 2016-09-29
seam_edges.cpp fixed duplicate boundary edges bug and removed helper

yig:
seam_edges was returning duplicate boundary edges. It also has a helper function in an anonymous namespace, which is against the style guide. That has been moved to a lambda.


yig:
I also modified point_simplex_squared_distance to optionally return the barycentric coordinates it computes.

alecjacobson:
Thanks !

PR #354, opened 2016-10-07
Change "bary.resize( DIM , 1 )" to "bary.resize( DIM )"

n-taka:
With "bary.resize( DIM , 1 )", bary is assumed Eigen::VectorXd. But sometimes bary is Eigen::RowVectorXd (e.g. void igl::AABB<DerivedV,DIM>::leaf_squared_distance in AABB.cpp) and which causes compile error with signed_distance function(s). So I remove "1".


n-taka:
I made a mistake in the comment above. The error caused by this bug is not a compile error but a runtime error. (eigen assertion)

alecjacobson:
This is an annoying feature/bug of Eigen. If `bary` is a VectorXd then `bary.resize(DIM,1)` will work correctly, but if it's a RowVectorXd then it will not... On the other hand, as you write, `bary.resize(1)` will work for both VectorXd and RowVectorXd, but then it will fail for MatrixXd (which is not necessarily unreasonable to use). Probably the right thing to use is something like:
`bary.resize( Derivedbary::RowsAtCompileTime == 1 ? 1 : DIM,  Derivedbary::ColsAtCompileTime == 1 ? 1 : DIM)` but then of course this could be fooled by passing a Matrix3x3. Maybe the best thing to do is to force bary to be a column vector type and hope/expect that if you pass a row vector the compiler avoids the copy. (This would be better served as a question on the Eigen Forum).

n-taka:
Thank you for your comment!
I change the resize function according to your advice. And I also add assertion statement `assert((Derivedb::RowsAtCompileTime == 1 || Derivedb::ColsAtCompileTime == 1) && "bary must be Eigen Vector or Eigen RowVector");` to check bary is Eigen Vector or RowVector explictly (avoid Eigen Matrix3x3 as you mentioned).

I agree that the best choice is to force bary to be a column vector. But this cause modification in some source codes (for example, AABB.cpp as I found). So, This might be laborious.

alecjacobson:
This is probably OK for now. However, assert should be a `static_assert` in this case, these are known at compile time (will also help catch the template issues with `Eigen::Matrix`)

PR #358, opened 2016-10-12
Use edge_lengths_squared to improve robustness

jmespadero:
I have written a edge_lengths_squared function to compute the squared lengths of sides of every triangle/tetrahedron. The reason is to improve the robustness (and sometimes the performance too) of several algorithms that need to use the squared norm of the vectors.

The most obvious result is that it removes the roundoff errors in the igl::cotmatrix() function even for obviously simple case tests as two triangles in unit square.

```
#include <igl/cotmatrix.h>
#include <igl/cotmatrix_entries.h>
#include <igl/internal_angles.h>
#include <Eigen/Dense>
#include <Eigen/Sparse>
#include <iostream>
using namespace std;
int main()
{
  Eigen::MatrixXd V(4,2);
  V<<0,0,
     1,0,
     1,1,
     0,1;
  Eigen::MatrixXi F(2,3);
  F<<0,1,2,
     0,2,3;
     
  Eigen::MatrixXd C(2,3);
  igl::cotmatrix_entries(V,F,C);
  std::cout<<"Cotangents: "<<std::endl<<Eigen::MatrixXd(C)<<std::endl;

  Eigen::MatrixXd A(2,3);
  igl::internal_angles(V,F,A);
  std::cout<<"Angles: "<<std::endl<<A*180./M_PI<<std::endl;
     
  Eigen::SparseMatrix<double> L;
  igl::cotmatrix(V,F,L);
  std::cout<<"cotmatrix: "<<std::endl<<Eigen::MatrixXd(L)<<std::endl;
  
  return 0;
}
```


alecjacobson:
A quick review of this checks out. But sense this is touching some very heavily used functions, we should run all of the examples and make sure that things are behaving correctly. (Here's where unit tests would be great).

jmespadero:
You are right. Please let me know if some test went wrong. 

Also, I would sugest a function that compute a matrix with the (robust) cosines and sines of every internal angle of the triangles (similar to cotmatrix_entries), so every other trigonometric function (as angles, tangent and cotangets) will be trivial.

jmespadero:
Hi, 

Can I do some more to help this pull-request go ahead? Do you want me to write some kind of unit test?
I have found another function (project_isometrically_to_plane()) that could benefit the usage of edge_lengths_squared() to avoid to have to compute edges and then have to square them.

alecjacobson:
yes, in fact, this morning I revived the

https://github.com/libigl/libigl-unit-tests

The documentation there is not great, but if you're willing to hunt around
and see how things are set up you could add some unit tests for the
functions (previously) depending on edge_lengths().

jmespadero:
Done. I have added unit tests for the functions depending on edge_lengths_squared()

- edge_lengths()
- cotmatrix_entries()
- cotmatrix()

If you accept this pull-request, I will change project_isometrically_to_plane() to use edge_lengths_squared()

alecjacobson:
Could we change `edge_lengths_squared` to `squared_edge_lengths`? Since that's how we'd say it in English. 

Beyond that the change to `internal_angles` seems like the only problematic one. The  issue is that if someone's code is currently calling `internal_angles(L,A)` where `L` is a list of edge lengths (not squared). Then after merging this change, their code will silently breaking and `L` will be interpreted as squared edge lengths. This is because the prototype of the function didn't change, but the interpretation of the input parameters did change.

It seems that internally we're only using the `internal_angles(V,F,A)` version and eventually we'd want people to use the squared edge lengths version. Maybe we should _temporarily_ replace the `internal_angles(L,A)` prototype with `internal_angles_using_edge_lengths(L,A)` and `internal_angles_using_squared_edge_lengths(L,A)` where `internal_angles_using_edge_lengths(L,A) = internal_angles_using_squared_edge_lengths(L.squared(),A);`.

This will cause peoples cause to break at compile time rather than silently give the wrong answer at run time. Eventually we can assume that everyone has made the change and remove `internal_angles_using_edge_lengths` and rename `internal_angles_using_edge_lengths` back to `internal_angles`.  Not ideal, but I'm not sure of a better way to do this.

jmespadero:
Of course we can change the name, it will match the criterion used in "double_areas()". 

I agree to you plan or using temporal names `internal_angles_using_edge_lengths(L,A)` and `internal_angles_using_squared_edge_lengths(L,A)`, and mark both as deprecated. Also I would add a comment in the documentation of `internal_angles_using_edge_lengths(L,A)`explaining that the `internal_angles_using_squared_edge_lengths(L,A)` is preferred.

I will make the changes next monday.

alecjacobson:
Excellent. Thanks. Sorry this pull request was painful.

PR #363, opened 2016-10-16
Triangulation

qnzhou:
Update summary:

* Refactor exact predicate from CGAL into ```igl::copyleft::cgal::ExactPredicate<Scalar>```
* Implemented single edge flip: ```igl::flip_edge(...)```
* Implemented lexicographic triangulation: ```igl::copyleft::cgal::lexicographic_triangulation(...)```
* Implemented edge flipping Delaunay triangulation: ```igl::copyleft::cgal::delaunay_triangulation(...)```


alecjacobson:
Cool!

Why use a class for the exact predicate? It'd be cool to eventually have a version in the main libigl that could use fast inexact predicates or free exact predicates (from Triangle) or predicates taking CGAL exact rationals. If there's only a few predicates needed (orientation, incircle, insphere) then I'd just as soon pass these as explicit function handle input parameters. Then we don't end up with bloated, unmanageable classes and we fit into the libigl style better.

(also seems to be some issues with the build checks)

qnzhou:
OK, the code is refactored to have predicates as functions.

alecjacobson:
Awesome! So now if we want a purely CGAL-free delaunay implementation for floating point numbers we could suck out the incircle and orient2d predicates from Jonathan Shewchuk's _public domain_ [predicates.c](http://www.cs.cmu.edu/afs/cs/project/quake/public/code/predicates.c)

alecjacobson:
(We should add some unit tests for this on https://github.com/libigl/libigl-unit-tests

qnzhou:
Yes, a version of the predicates are embedded in his triangle.c file.  However, its signature is different from the ones defined in predicates.c.  Maybe add a ```external/predicates``` dependency?

alecjacobson:
since its public domain we might as well just drop the code into the main
libigl

jdumas:
Hi there. I you want a CGAL-free implementation of predicates I can also suggest to look into [Geogram](http://alice.loria.fr/software/geogram/doc/html/index.html), and especially the `Predicates_psm_*.zip` in the download page, which contains only 1 couple of .cpp/.h files. You can check the list of available predicates [here](http://alice.loria.fr/software/geogram/doc/html/predicates_8h.html).

qnzhou:
@alecjacobson I don't think predicates.c can be used in a header-only way...  right?

@jdumas Good to know, will checkout its Delaunay modules.

alecjacobson:
This page http://alice.loria.fr/software/geogram/doc/html/geogram_license.html of Geogram makes it seem like their using Shewchuk's predicates. @jdumas do you know if this is the case?

Maybe the Grogram project has already done the work of getting rid of Shewchuk's global variables and `#define` macros. Otherwise, it does seem that converting predicates.c will be a chore. 

Skimming through `predicates.c`, one way to do this would be to 

- (automatically) convert all `#define` macros to inline functions (in theory without incurring a performance hit)
- manually add the 15 global variables as extra params to functions that use them (100 or so use in the whole file), maybe this can be done automatically by adding all 15 to every function regardless
- add REAL as a template rather than `#define`
- finally, to implement `igl::delaunay_triangulation`, use lambdas and capturing:
```cpp
igl:: delaunay_triangulation(V,F)
{
  DerivedV::Scalar epsilon, splitting ... // Shewchucks 15 "globals"
  predicates_exact_init<DerivedV::Scalar>(epsilon, splitting, ... )
  const auto & orient2d = [&epsilon,&splitting,&...](pa,pb,pc)
  {
     return predicates_orient2D<DerivedV::Scalar>(pa,pb,pc,epsilon,splitting,...);
  }
  const auto & incircle = ... same thing for incircle
  return delaunay_triangulation(V,oreint2D,incircle,F);
}
```

jdumas:
@alecjacobson Yes it a reimplementation of Shewchuk's predicates. Quoting the documentation:

> PCK (Predicate Construction Kit) implements a set of geometric predicates. PCK uses arithmetic filters (Meyer and Pion), expansion arithmetics (Shewchuk) and simulation of simplicity (Edelsbrunner).

I'll ask Bruno Levy about the globals when I see him, but you still need to call a `initialize()` function before using any predicate, at it is needed to compute the machine epsilon. And I don't think it is meant to be used in a header-only way, but you can use a trick similar to [nanosvg](https://github.com/memononen/nanosvg) if you want, i.e. define a macro `LIBIGL_IMPLEMENTATION` to populate only one compilation unit with the actual implementation.

alecjacobson:
I haven't seen the nanosvg thing
 Is that also threadsafe?

jdumas:
@alecjacobson I finally got around to ask Bruno about his precidate code. It's a complete reimplementation that is different from Shewchuk's code. The predicates should also be thread-safe, because expansions are allocated on the stack (via `alloca`).

Also, the call to `GEO::expansion::initialize` is only necessary if the target processor does not support the `fma` operation. See this comment in the `Predicates_psm.cpp` source code:

        // If the target processor supports the FMA (Fused Multiply Add)
        // instruction, then the product of two doubles into a length-2
        // expansion can be implemented as follows. Thanks to Marc Glisse
        // for the information.
        // Note: under gcc, automatic generations of fma() for a*b+c needs
        // to be deactivated, using -ffp-contract=off, else it may break
        // other functions such as fast_expansion_sum_zeroelim().

alecjacobson:
Very cool. Do you know the license of that code? (i.e., do we need to put
it in igl/copyleft )

jdumas:
I believe it is the same as the rest of Geogram (i.e. [3-clauses BSD License](http://alice.loria.fr/software/geogram/doc/html/geogram_license.html)), since there is no external code in the `Predicates_psm.cpp`. But if you have any doubt you can drop a mail to Bruno.

PR #365, opened 2016-10-17
Remove semiperimeter unused code in doublearea.cpp

jmespadero:
I have found two sections of unnecessary/unused code in include/igl/doublearea.cpp.

First, there is a sort of side lengths which is not necessary for the Heron's formula to work:
`
  Eigen::Matrix<typename Derivedl::Scalar, Eigen::Dynamic, 3> l;
  MatrixXi _;
  sort(ul,2,false,l,_);
`

Then, the semiperimeter of the triangle is computed but never used:
`
  // semiperimeters
  Matrix<typename Derivedl::Scalar,Dynamic,1> s = l.rowwise().sum()*0.5;
  assert((Index)s.rows() == m);
`
Each of these sections forces a complete traversal of the side lengths array.

### DIFF
--- a/include/igl/doublearea.cpp
+++ b/include/igl/doublearea.cpp
@@ -141,26 +141,20 @@ IGL_INLINE void igl::doublearea(
   assert(ul.cols() == 3);
   // Number of triangles
   const Index m = ul.rows();
-  Eigen::Matrix<typename Derivedl::Scalar, Eigen::Dynamic, 3> l;
-  MatrixXi _;
-  sort(ul,2,false,l,_);
-  // semiperimeters
-  Matrix<typename Derivedl::Scalar,Dynamic,1> s = l.rowwise().sum()*0.5;
-  assert((Index)s.rows() == m);
   // resize output
-  dblA.resize(l.rows(),1);
+  dblA.resize(m,1);
   parallel_for(
     m,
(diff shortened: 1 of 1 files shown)


jmespadero:
By the way, a small question...

Why to compute the side lengths (which need a sqrt for each side) and then the double_area (which need an extra sqrt) when you could just use the norm of the cross product of two edges to compute the double_area?

alecjacobson:
Removing the semi-perimeter is OK.

But this commit also removes the sort which is necessary for robust computation, see "Miscalculating Area and Angles of a Needle-like Triangle"
https://people.eecs.berkeley.edu/~wkahan/Triangle.pdf
"§2. How to compute ∆ :"
(also applies to question about cross products)

jmespadero:
Ok. Many thanks for the response and the  link. I would suggest to include a small comment in the code to point the reason of sorting.

alecjacobson:
Good suggestion. This is added now to my branch, which I'll merge shortly.

jmespadero:
Also, I would suggest to change this line:
`sort(ul,2,false,l,_);`
to 
`igl::sort(ul,2,false,l,_);`

Which is the same... but will help to understand the code. First time I had read it, I believed (and also my IDE when I pressed) it was a strange call to std::sort or Eigen::sort.

PR #371, opened 2016-10-21
Add support for read/write .off files with color-per-vertex info

jmespadero:
Hi,

I have tuned the readOFF and writeOFF methods to support color-per-vertex information

An minimal example of writeOFF() usage is just adding a line of code to the end of tutorial 104

```cpp
#include <igl/readOFF.h>
#include <igl/viewer/Viewer.h>
#include <igl/jet.h>
#include "tutorial_shared_path.h"

Eigen::MatrixXd V;
Eigen::MatrixXi F;
Eigen::MatrixXd C;

{
  // Load a mesh in OFF format
  igl::readOFF(TUTORIAL_SHARED_PATH "/screwdriver.off", V, F);

  // Plot the mesh
  igl::viewer::Viewer viewer;
  viewer.data.set_mesh(V, F);

  // Use the z coordinate as a scalar field over the surface
  Eigen::VectorXd Z = V.col(2);

  // Compute per-vertex colors
  igl::jet(Z,true,C);

  // Add per-vertex colors
  viewer.data.set_colors(C);

  // Launch the viewer
  viewer.launch();

  // Save our colored mesh
  igl::writeOFF("screwdriverColor.off", V, F, C);
}
```

Which allow to save our data and see it in meshlab or coindesigner.

fromigltomeshlab

readOFF() allows us to export a mesh in meshlab using .off format, check the Vert->color option, and then visualize it in IGL. To load a mesh with color we just have to add an extra array to readOFF() and then check if the color information is valid or empty:

```cpp
#include <igl/readOFF.h>
#include <igl/viewer/Viewer.h>
#include <iostream>

{
  Eigen::MatrixXd V, C;
  Eigen::MatrixXi F;

  // Load a mesh in OFF format and add to a viewer
  igl::viewer::Viewer viewer;
  igl::readOFF("screwdriverColor.off", V, F, C);
  viewer.data.set_mesh(V, F);

  //Check colors.
  if (C.rows() == V.rows())
  {
      std::cout << "Mesh has color per vertex information." << std::endl;
      viewer.data.set_colors(C);
  }
  else
  {
      std::cout << "Mesh has no colors." << std::endl;
  }

  viewer.launch();
}
```
Here is the output of Filter-> Color Creation -> Discrete Curvatures -> RMS Curvature for the lion.off model, exported and visualized with IGL.

frommeshlabtoigl


alecjacobson:
Is it possible to separate the squared_edge_lengths stuff from this pull request? I think then we can merge it immediately.

PR #374, opened 2016-10-24
squared_edge_lengths (clean)

jmespadero:
I have written a `squared_edge_lengths()` function to compute the squared lengths of sides of every triangle/tetrahedron. The reason is to improve the robustness (and sometimes the performance too) of several algorithms that need to use the squared norm of the vectors.

The most obvious result is that it removes the roundoff errors in the igl::cotmatrix() function even for obviously simple case tests as two triangles in unit square.

```cpp
#include <igl/cotmatrix.h>
#include <igl/cotmatrix_entries.h>
#include <igl/internal_angles.h>
#include <Eigen/Dense>
#include <Eigen/Sparse>
#include <iostream>
using namespace std;
int main()
{
  Eigen::MatrixXd V(4,2);
  V<<0,0,
     1,0,
     1,1,
     0,1;
  Eigen::MatrixXi F(2,3);
  F<<0,1,2,
     0,2,3;
     
  Eigen::MatrixXd C(2,3);
  igl::cotmatrix_entries(V,F,C);
  //Should show values without roundoff errors
  std::cout<<"Cotangents: "<<std::endl<<Eigen::MatrixXd(C)<<std::endl;

  Eigen::MatrixXd A(2,3);
  igl::internal_angles(V,F,A);
  std::cout<<"Angles: "<<std::endl<<A*180./M_PI<<std::endl;
     
  Eigen::SparseMatrix<double> L;
  igl::cotmatrix(V,F,L);
  std::cout<<"cotmatrix: "<<std::endl<<Eigen::MatrixXd(L)<<std::endl;
  
  return 0;
}
```


jmespadero:
There is a companion pull-request in libigl-unit-tests renaming functions. It is possible that travis check will fail if not applied.

PR #379, opened 2016-10-25
Input filename as argument

jmespadero:
Hi,
This PR changes some demos in the tutorial to accept an input mesh filename as
argument, instead of working with a fixed input.  Usually the change consist in replacing the line:

``` cpp
igl::readOFF(TUTORIAL_SHARED_PATH "/cow.off", V, F);
```

by this code:

``` cpp
  // Set input mesh filename
  std::string filename(TUTORIAL_SHARED_PATH "/cow.off");
  if (argc > 1)
      filename = std::string(argv[1]);

  // Try to load the input mesh
  if (igl::read_triangle_mesh(filename, V, F) == false)
      return -1;
```

This makes possible to try the demos with several input meshes without recompile.

In some demos, I have added an extra check if the mesh needs some property. For example:

``` cpp
  if (E.rows() < 1)
      std::cerr << "error: Mesh has no boundary"<<std::endl;
```

Laplacian:
laplacian_lion

ARAPParam:
arapparam_lion
Marching cubes:
marching_lion

Picking:
picking_lion


jmespadero:
I am aware that this has also a negative side. Each demo has become a few lines longer, and may be seen as "less focused in geometry". Please feel free to accept or discard this Pull Request.

jmespadero:
I will close this pull request next monday. It makes no sense to keep it open forever.

PR #392, opened 2016-11-08
Support for older GCC

dukecyto:
I had problem compiling AABB with GCC 4.6.3. It seems to be related to this:
http://stackoverflow.com/questions/16007855/gcc-compilation-error-when-using-template-class-method-inside-lambda-expression


alecjacobson:
This is an easy, harmless change. Though in general we're not supporting older versions of GCC so you may run into other issues down the line.

PR #394, opened 2016-11-10
Make usample work like loop subdivision

ikoruk:
Allows for returning a sparse matrix that can then be multiplied with properties


alecjacobson:
Thanks, during the merge I also caught a bug in the loop implementation (which permeated into the new upsample).

**_Public service announcement_**:

In Eigen it is _**not safe**_ to do:

    NV = S*NV;

Instead, use:

    NV = (S*NV).eval();

Issue #391, opened 2016-11-07
Monkey test for compilation/include order


alecjacobson:
Here's a small bash  script that will construct a no-op c++ file that includes **_all_** libigl headers in a random order:

    echo "`ls -1 $LIBIGL/include/igl/*.h | sed -e "s/.*igl\/\(.*\)/#include <igl\/\1>/" | perl -MList::Util=shuffle -wne 'print shuffle <>;'`
    int main(){ }" | clang++ -xc++ - -I$LIBIGL/include -I/usr/local/include/eigen3 -std=c++11

Because of the random order (this is a [monkey test](https://en.wikipedia.org/wiki/Monkey_testing) ), it may take multiple executions to root out missing namespaces/includes.

I tested this on my Mac + clang and found 6 or 7 compilation errors (fixed in 6cfb5a48fdc8b93fb7deda81030e17c00b3ba068).

Try it out on your *nix machines. I'm curious if it finds any more errors on Linux + gcc.

jmespadero:
This is my version of the script for Linux+gcc:
```bash
#!/bin/bash
while true
do
    #Create a name for this test based on current date and time
    testName=`date +%Y%m%d%H%M%S`
    echo Create file $testName.cpp
    echo "`ls -1 ../include/igl/*.h | sed -e "s/.*igl\/\(.*\)/#include <igl\/\1>/" | perl -MList::Util=shuffle -wne 'print shuffle <>;'`
    int smain(){ }" > $testName.cpp
    echo Compile file $testName.cpp
    g++ -S $testName.cpp -I../include -IEigen -std=c++11 -fpermissive -Wno-deprecated-declarations 
    #Check if compile was ok (ignoring warnings)
    ls $testName.s || (echo TEST $testName FAILED && break )
    #Clean files for this test
    echo Test OK. Clean $testName and repeat new test
    rm  $testName.* 
done
    echo Check files *.cpp
```

I get this warning **on every** execution:
```
Compile file 20161110130455.cpp
In file included from 20161110130455.cpp:39:0:
../include/igl/serialize.h:191:25: warning: inline function ‘virtual void igl::Serializable::InitSerialization()’ used but never defined [enabled by default]
     inline virtual void InitSerialization() = 0;
                         ^
```

And sometimes, I get also this warning:
```
In file included from Eigen/Eigen/StdVector:23:0,
                 from ../include/igl/column_to_quats.h:13,
                 from 20161110130350.cpp:73:
[...]

PR #396, opened 2016-11-10
Bugfix. Remove warnings and errors

jmespadero:
Remove warnings detected on issue #391 

-  Remove ```inline``` keyword of the pure virtual method InitSerialization() 

- Remove all ```#include <Eigen/StdVector>``` lines, which are no necessary if compiler is C++11 ( and C++11 is mandatory to libigl). See  http://eigen.tuxfamily.org/bz/show_bug.cgi?id=829 .


jmespadero:
Added bugfix to ```error: ‘NULL’ was not declared in this scope``` in igl/Timer.h

PR #414, opened 2016-12-08
Removed some PlainObjectBase declarations for Eigen 3.3 compatibility.

yig:
I removed some more problematic uses of PlainObjectBase. In one place, the local type is now declared as the Derived type directly. In the other, the output parameter is now templates as MatrixBase.


yig:
Should I commit the change that declares `bary` and `b` as `PlainObjectBase` instead of `MatrixBase` in `point_simplex_squared_distance.h`?

alecjacobson:
That'd be great.

PR #424, opened 2016-12-21
Some miscellaneous minor changes

qnzhou:
Change summary:

* Minor update to `mesh_boolean` so it handles empty mesh better.
* Change template type for `point_template_squared_distance(...)` method.
* Avoid bound check in `flip_edge(...)`


include/igl/flip_edge.cpp
```diff
@@ -46,10 +46,10 @@ IGL_INLINE void igl::flip_edge(
   assert(c2 < 3);
 
   assert(f1 != f2);
-  const size_t v1 = F(f1, (c1+1)%3);
-  const size_t v2 = F(f1, (c1+2)%3);
-  const size_t v4 = F(f1, c1);
-  const size_t v3 = F(f2, c2);
+  const size_t v1 = F.coeff(f1, (c1+1)%3);
```
alecjacobson: Why is `.coeff` needed here?

include/igl/flip_edge.cpp
```diff
@@ -72,20 +72,20 @@ IGL_INLINE void igl::flip_edge(
   assert(E(e_32, 0) == v3);
   assert(E(e_32, 1) == v2);
 
-  const size_t ue_24 = EMAP[e_24];
-  const size_t ue_41 = EMAP[e_41];
-  const size_t ue_13 = EMAP[e_13];
-  const size_t ue_32 = EMAP[e_32];
+  const size_t ue_24 = EMAP.coeff(e_24);
```
alecjacobson: Why can't this just be `EMAP(e_24)` (with curved parentheses) ?

include/igl/flip_edge.cpp
qnzhou: `.coeff(...)` and `.coeffRef(...)` skips range checking...

alecjacobson:
Is this just to avoid assets in debug mode?

qnzhou:
Ah, I didn't realize it is disabled in release...  I thought it is only
disable when `EIGEN_NO_DEBUG` is defined.

include/igl/flip_edge.cpp
alecjacobson: I think these EMAP[...] should all be curved parenthesis. We should use the regular () with bound checking so we can get asserts in debug mode.

qnzhou:
Pushed my fixes.  Let's see if the auto compiles will pass.

alecjacobson:
Thanks. This will probably pass the tests. Then I'll merge. @yig , I think Jame's commit will make the point_simplex_squared_distance part of https://github.com/libigl/libigl/pull/423 obsolete (and thus the PR merge-able).

PR #443, opened 2017-01-11
Fixed order of operations bug when computing model transform

rkevingibson:
A minor bug, where Eigen doesn't like casting from a temporary. Without the proper brackets, the multiply by 0.5 results in an Eigen temporary type, which doesn't cast well. Changing the order of operations before the eval makes centroid a Vector3f type which can be cast. Also means we can get rid of the extra evals.

Apparently this manifests only in the debug mode in Visual Studio, and is fine in release.


alecjacobson:
Ewww. Thanks for the fix.

Issue #270, opened 2016-05-26
libigl's boolean take very large memory


hohogpb:
libigl's boolean take very large memory. i suggest if you can dealing with it, it will be better.

alecjacobson:
Can you post an example?

hohogpb:
code: a little change of 609_boolean
`#define REMESH_INTERSECTIONS_TIMING
#define EXTRACT_CELLS_DEBUG
#define SMALL_CELL_REMOVAL 0
#define MESH_BOOLEAN_TIMING
#define IGL_SELFINTERSECTMESH_DEBUG

#include <igl/readOFF.h>
#include <igl/readstl.h>
#include <igl/writestl.h>



//#define IGL_NO_CORK
//#undef IGL_STATIC_LIBRARY
#include <igl/copyleft/cgal/mesh_boolean.h>
#include <igl/viewer/Viewer.h>

#include <Eigen/Core>
#include <iostream>

#include "tutorial_shared_path.h"

Eigen::MatrixXd VA,VB,VC,N;
Eigen::VectorXi J,I;
Eigen::MatrixXi FA,FB,FC;
igl::MeshBooleanType boolean_type(
  igl::MESH_BOOLEAN_TYPE_MINUS);

const char * MESH_BOOLEAN_TYPE_NAMES[] =
{
  "Union",
  "Intersect",
  "Minus",
  "XOR",
  "Resolve",
};

void update(igl::viewer::Viewer &viewer)
{
  igl::copyleft::cgal::mesh_boolean(VA,FA,VB,FB,boolean_type,VC,FC,J);
  Eigen::MatrixXd C(FC.rows(),3);
  for(size_t f = 0;f<C.rows();f++)
  {
    if(J(f)<FA.rows())
    {
      C.row(f) = Eigen::RowVector3d(1,0,0);
    }else
    {
      C.row(f) = Eigen::RowVector3d(0,1,0);
    }
  }

  igl::writeSTL("C:/Users/Administrator/Desktop/newBooleanTest/testdebugBool/result110.stl", VC, FC, false);
  viewer.data.clear();
  viewer.data.set_mesh(VC,FC);
  viewer.data.set_colors(C);
}

bool key_down(igl::viewer::Viewer &viewer, unsigned char key, int mods)
{
  switch(key)
  {
    default:
      return false;
    case '.':
      boolean_type =
        static_cast<igl::MeshBooleanType>(
[...]

PR #450, opened 2017-01-20
Memory opt

qnzhou:
Incorporated the following suggestion in https://github.com/libigl/libigl/issues/270#issuecomment-273701288

> Also, I think storing pairs of triangle indices instead of copying the boxes would save a lot of memory.

Based on a single example test [(thing 1368052)](https://ten-thousand-models.appspot.com/detail.html?file_id=1368052), the peak memory usage is reduced from >15G to ~7G.


alecjacobson:
Great!

PR #459, opened 2017-02-03
Avoid precision issue in AABB::init()

puzzlepaint:
I ran into an issue where ```AABB::init()``` (with Scalar == float) crashed for a valid but large mesh. It seems that float was not sufficient to store the median element used in the construction, therefore the LI and RI vectors were not filled correctly. Instead of defaulting to double for the median, I would propose to avoid precision issues by directly finding the largest element for LI, which is an int, instead of computing the average of the two middle elements, in case of an even number of elements.

### DIFF
--- a/include/igl/AABB.cpp
+++ b/include/igl/AABB.cpp
@@ -148,22 +148,14 @@ IGL_INLINE void igl::AABB<DerivedV,DIM>::init(
         {
           SIdI(i) = SI(I(i),max_d);
         }
-        // Since later I use <= I think I don't need to worry about odd/even
         // Pass by copy to avoid changing input
-        const auto median = [](VectorXi A)->Scalar
+        const auto median = [](VectorXi A)->int
         {
-          size_t n = A.size()/2;
+          size_t n = (A.size()-1)/2;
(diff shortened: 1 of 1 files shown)


puzzlepaint:
The failing check seems to be due to this error which looks unrelated:
```
wget -P external/nanogui/ext/glfw/include/GL http://www.opengl.org/registry/api/GL/glcorearb.h
[...]
ERROR: no certificate subject alternative name matches
       requested host name `www.khronos.org'.
[...]
The command "wget -P external/nanogui/ext/glfw/include/GL http://www.opengl.org/registry/api/GL/glcorearb.h" exited with 5.
```

danielepanozzo:
I pushed a fix, but I prefer if Alec double check this PR before accepting it.

alecjacobson:
I want to merge this, but I don't want to break the build for this wget error. Is that fixed now?

danielepanozzo:
Yes it is fixed. I still don't know why this certificate check fails and I cannot reproduce it on my virtual machine --- I simply disabled it in the autobuild and it seems fine now.

PR #462, opened 2017-02-10
Fix for compilation with Visual C++ 2015

enricodetoma:
To be able to compile with Visual C++ 2015, template parameters `<int, double>` must be explicitly specified to be able to call the overloaded `mosek_quadprog` template function (otherwise the compiler tries to recursively call the same caller function).

In addition, `MSK_initenv` has been removed since it is no longer present in Mosek 8 and is probably not necessary even with Mosek 7 and previous: `MSK_makeenv` should do everything that is needed.

### DIFF
--- a/include/igl/mosek/mosek_quadprog.cpp
+++ b/include/igl/mosek/mosek_quadprog.cpp
@@ -110,6 +110,10 @@ IGL_INLINE bool igl::mosek::mosek_quadprog(
   //  printf("%s",str);
   //}
   //mosek_guarded(MSK_linkfunctoenvstream(env,MSK_STREAM_LOG,NULL,printstr));
+#if MSK_VERSION_MAJOR <= 7
+  // initialize mosek environment
+  mosek_guarded(MSK_initenv(env));
+#endif
   // Create the optimization task
   mosek_guarded(MSK_maketask(env,m,n,&task));
   verbose("Creating task with %ld linear constraints and %ld variables...\n",m,n);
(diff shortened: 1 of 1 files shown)


danielepanozzo:
Thanks for the fix!

Why has this line been removed? 
``
mosek_guarded(MSK_initenv(env));
``

After a quick look at the Mosek documentation this seems to be necessary.

enricodetoma:
Mosek 8 does not contain `MSK_initenv` any more (and if that line is not removed, it won't compile): http://docs.mosek.com/8.0/capi/search.html?q=MSK_initenv&check_keywords=yes&area=default

In addition, if you look at the examples in Mosek 7 documentation, for example http://docs.mosek.com/7.0/capi/Quadratic_optimization.html you see that `MSK_makeenv` is necessary, but no call to `MSK_initenv` is present.

enricodetoma:
Actually, I didn't check if `MSK_initenv` is necessary for Mosek 6, so I guess my last commit is safer, please review it: I only removed `MSK_initenv` with Mosek 8 where it has to be removed to compile.

enricodetoma:
Here's an additional example which shows the need for removing `MSK_initenv` with Mosek 8 https://projects.coin-or.org/Osi/changeset/2090

danielepanozzo:
All looks good to me, I want to wait for @alecjacobson to confirm since he is the author of most (or maybe all) of the functions that use mosek in libigl.

enricodetoma:
I close this pull request because it doesn't apply correctly to the new commits in master branch

PR #470, opened 2017-02-20
Multi-mesh viewer support

schuellc:
* add viewer support for multiple mesh objects (separated mesh object
  data properties from core)
* add support for viewer windows size state
* move viewer state serialization to new file ViewerSerialization.h
* extend cmake options to expose new viewer options
* adjust tutorials to work with new multi mesh viewer
* other small improvments and fixes


alecjacobson:
Moving `show_lines` etc. from core to data indeed seems like the _right_ thing to do. but this will likely break a lot of projects. I wonder if we chew on these changes in a dev branch where we also 

 - move viewer to the glfw dependency directory, and 
 - factor out viewer plugins like nanogui 

This way we break projects only once.

schuellc:
Yes, makes sense.

alecjacobson:
I think this is ready to go, but let's compile and run each tutorial just in case (if you or somebody else does it, I'll believe you; otherwise I might not get to it for a day or two)

danielepanozzo:
I tried the tutorials and on mac everything seems fine, except a major difference in the shading. This is what happens for 201:

Before:
before

After:
after

It is not a problem of inverted normals, it seems to be related with the new camera. @schuellc: do you know what it could be? Do you also get the same behavior on windows?

schuellc:
This was an issue with the fixed light position which was to close to the mesh. Previously, this worked because we had a model_zoom factor scaling the mesh appropriately. For multiple meshes we need to see their actual dimensions. Therefore, we should adjust the light positions rather than the model scale. I fixed this.

danielepanozzo:
This fixed the problem, thanks! I think it is ready to be merged as is. @alecjacobson do you want to move this into a separate branch or can I merge in the master?

alecjacobson:
I'm working on this refactor now. The viewer code is very messy, but I'm hoping to clean it up. Merging this PR will require some care afterward. 

This PR contains a lot of changes that seem to be unrelated to supporting multiple meshes in the viewer. Is it possible to separate out just the changes that support multiple meshes? Other changes (e.g., serialization, cmake, where `show_lines` lives) should each have their own small PR. This makes them much easier to merge.

include/igl/read_triangle_mesh.cpp
```diff
@@ -43,7 +43,7 @@ IGL_INLINE bool igl::read_triangle_mesh(
     bool success = readOBJ(str,V,TC,N,F,FTC,FN);
     for(auto & v : V)
     {
-      v.resize(std::min(v.size(),(size_t)3));
+      v.resize((std::min)(v.size(),(size_t)3));
```
alecjacobson: I don't understand the parenthesis around `std::min`. Is this a windows thing?

include/igl/viewer/Viewer.cpp
```diff
@@ -355,39 +468,148 @@ namespace viewer
...
+    return active_data_id;
+  }
+
+  IGL_INLINE bool Viewer::set_active_mesh(unsigned int data_id)
+  {
+    assert(data_buffer.size() > data_id && "data_id out of range");
+
+    data_buffer[active_data_id] = ViewerData(data);
+    active_data_id = data_id;
+    data = data_buffer[active_data_id];
```
alecjacobson: Is this copying the entire mesh+data every time the active mesh id is changed?

include/igl/read_triangle_mesh.cpp
schuellc: Right, that bypasses the macro definition in windef.h

include/igl/viewer/Viewer.cpp
schuellc: That's correct, that's the only way to not have to change the current "interface" of the viewer where a lot of code is using the member data directly. Or do you have another suggestion?

include/igl/read_triangle_mesh.cpp
alecjacobson: Wow, what a mess. So every time we use `std::min()` we should be using `(std::min)()`?

include/igl/viewer/Viewer.cpp
alecjacobson: I wonder if we should just bite the bullet and change it to a member function (e.g., `.current_data()`) that returns a reference. All old code will have to change lines like:

    viewer.data.face_based = true;

to 

    viewer.current_data().face_based = true;

But all old code is going to break with the namespace changes anyway, so if it's done at the same time maybe that's OK.

danielepanozzo: Performance-wise it will not make a big difference --- we still have to convert our doubles to float every time the mesh is sent to the VBO (double support was really shaky on ATI drivers, I could not get it to work reliably 2 years ago). If we go for the a reference, then it would probably make sense to try again to switch the shader to doubles and avoid all the conversion that are currently happening in OpenGL state (and simplify the viewer code considerably since there is no need to buffer the matrices). Overall, I prefer Alec's solution, it will break the current code (but fixable with a search&replace) but it is more clean.

schuellc: I completely agree. If we can break with old code than it is the much better solution!

include/igl/read_triangle_mesh.cpp
schuellc: I guess so, that's the safest way crossplatform-wise. Or you ask all the windows people to define NOMINMAX before including there windows.h in their projects.

alecjacobson:
@danielepanozzo , @schuellc 

Take a look at https://github.com/libigl/libigl/tree/viewer-refactor

Viewer components are moved into folders according to required dependencies. For example,

 - `igl::viewer::Viewer` --> `igl::opengl::glfw::Viewer`
 - `igl::viewer::ViewerCore` --> `igl::opengl::ViewerCore`
 - `igl::viewer::OpenGL_State` --> `igl::opengl::State`
 - `igl::viewer::ViewerData` --> `igl::ViewerData`

I've added a `data_list` and `opengl_list` to store each model's "data" and "OpenGL state" respectively.  Any old lines can easily be replaced:

`viewer.data.set_face_based(true)` --> viewer.selected_data().set_face_based(true)`

Here's a little snippet to get started:

```cpp

{
  igl::opengl::glfw::Viewer v;
  for(int m = 1;m<argc;m++) v.load_mesh_from_file(argv[m]);
  v.launch();
}
```

I find the current organization of viewer data a bit confusing. It's not clear what lives in "ViewerData", "ViewerCore" or "State". Perhaps this should also be reorganized before merging into the main branch.

danielepanozzo:
I think the current organization is quite sensible:
ViewerData --- Contains all the data that we want to visualize in "libigl" format
State --- Contains all the data in the opengl format and keep track of what is allocated on the GPU
ViewerCore --- Contains all the viewer-specific settings (camera, rendering options, zoom, etc.)
Viewer --- the glfw application, that uses the 3 modules above.

One possible improvement would be to merge ViewerData and State, it will improve in performances but it will be less organized and reusable. Let me know if you have another proposal on how to reorganize this.

Before this can be committed, the python bindings needs also to be fixed, and also all the python tutorials.

alecjacobson:
There's a bit of a confusion about attributes and switches: `face_based` lives in `ViewerData` but `show_lines` lives in `ViewerCore`.

`State` contains some seemingly static things (compiled shader code) and also per-model vertex arrays.

If we have a `ViewerData` instance for each "model", should each model have "its own" overlays? Sometimes overlays are model specific (vertex labels, cross fields), but other times they might be between two models (correspondence lines).

(`State` also duplicates data in `ViewerData`. This serves a couple purposes: getting it in float matrices with RowMajor ordering and also preparing per-face/per-corner data. I'm OK with this duplication for the same reasons @danielepanozzo writes.)

_We might also use this as an opportunity to give these objects more meaningful names._

Also, I didn't even fix the main C++ tutorials yet. I won't bother until you guys are happy with the layout. Let me know if you are, or if you have suggestions. (Also feel free to push/PR onto this branch).

Javier12:
Is there a way I can help with this?

josefgraus:
Same here, I'm interested in getting this feature in its final form (I fake multi-mesh by merging meshes and re-indexing, and it's becoming cumbersome -- I'd like to replace it), so I'm willing to help test or refactor where needed.

PR #474, opened 2017-02-23
Small Fix, separate the implementation and add documentation for HalfEdgeIterator

jiangzhongshi:
The first commit fix two small bugs in the implementation.

The second separate it to .h/.cpp, and add some documentation.


jiangzhongshi:
I'll add some explicit instantiation and re-pr later.

PR #479, opened 2017-02-24
Tetrahedral Analogue for HalfEdgeIterator and triangle_triangle_adjacency.

jiangzhongshi:
In this pull request, I have created new class `TetTupleIterator`: Tetrahedral Analogue for `HalfEdgeIterator` (which should be more properly called Cell-Tuple)

and the helper function `tetrahedron_tetrahedron_adjacency` , also as a tet analogue of  `triangle_triangle_adjacency`.

I've submit a pull request to libigl/libigl-unit-tests#10

### DIFF
--- a/include/igl/tetrahedron_tetrahedron_adjacency.cpp
+++ b/include/igl/tetrahedron_tetrahedron_adjacency.cpp
@@ -1,0 +1,134 @@
+// This file is part of libigl, a simple c++ geometry processing library.
+//
+// Copyright (C) 2017 Zhongshi Jiang <zhongshi@cims.nyu.edu>
+//
+// This Source Code Form is subject to the terms of the Mozilla Public License
+// v. 2.0. If a copy of the MPL was not distributed with this file, You can
+// obtain one at http://mozilla.org/MPL/2.0/.
+#include "tetrahedron_tetrahedron_adjacency.h"
+#include "all_edges.h"
+#include "unique_simplices.h"
+#include "parallel_for.h"
+#include "unique_edge_map.h"
+#include "slice.h"
+#include <Eigen/Core>
+#include <algorithm>
+#include <iostream>
+#include <array>
+namespace igl
+{
+  struct tetrahedron_tetrahedron_adjacency_tet_cell
+  {
+    int v1;
+    int v2;
+    int v3;
+    int t;
+    int f;
+  };
+
+// Extract the face adjacencies
+  template<typename DerivedT, typename TTT_type, typename DerivedTT,
+           typename DerivedTTif, typename DerivedTTie>
+  IGL_INLINE void tetrahedron_tetrahedron_adjacency_extractTTi(
+      const Eigen::MatrixBase<DerivedT> &F,
+      std::vector<TTT_type> &TTT,
+      Eigen::MatrixBase<DerivedTT> &TT,
+      Eigen::MatrixBase<DerivedTTif> &TTif,
+      Eigen::MatrixBase<DerivedTTie> &TTie)
+  {
+    TT.derived().setConstant((int) (F.rows()), 4, -1);
+    TTif.derived().setConstant((int) (F.rows()), 4, -1);
+    TTie.derived().setConstant((int) (F.rows()), 4 * 3, -1);
+
+    for(int i = 1; i < (int) TTT.size(); ++i)
+    {
+      auto &r1 = TTT[i - 1];
+      auto &r2 = TTT[i];
+
+      // if get same face.
+      if((r1.v1 == r2.v1) && (r1.v2 == r2.v2) && (r1.v3 == r2.v3))
+      {
+        TT(r1.t, r1.f) = r2.t;
+        TT(r2.t, r2.f) = r1.t;
+
+        // inverse for face
+        TTif(r1.t, r1.f) = r2.f;
+        TTif(r2.t, r2.f) = r1.f;
+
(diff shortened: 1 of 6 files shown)


alecjacobson:
This is a cool feature. Some of the documentation in the headers could be cleaned up. I don't think you mean [adjacent matrix](https://en.wikipedia.org/wiki/Adjacency_matrix), it's more like a neighbor list or ordered adjacency list. 

The capitalization is inconsistent `NextTE` vs `isBorder`. Actually both of these should be renamed to something without capitals, more like: `next_edge` , `is_border` according to the [libigl style guidelines](https://github.com/libigl/libigl/blob/master/style-guidelines.md). (and most of the other member function names should change analogously).

Also, I wonder if `flip*` is the right terminology in the `TetTupleIterator` (or `HalfEdgeIterator` for that matter). My understanding is that these are setting the iterator to the "opposite" (half-)simplex. They are not "flipping" the simplex in the sense that they are not chaining the mesh combinatorics.

I worry a bit that these HalfEdgeIterator and TetTupleIterator are encouraging deviation from [libigl style](https://github.com/libigl/libigl/blob/master/style-guidelines.md)

There are already some instances where HalfEdgeIterator is unnecessarily used. This is a slippery slope toward having it as an input parameter and accidentally creating a "libigl mesh class".

jiangzhongshi:
Thanks for the comment. 

Terminology-wise, I was trying to conform with `HalfEdgeIterator`. Now I think it's better to rewrite both of them.

As for the 'class' style, do you think it's good to write something like slim?

alecjacobson:
Agreed on all parts. In general, libigl strives to have function-based API
with clearly defined matrix-based inputs.

jiangzhongshi:
@alecjacobson 
I have changed the interfaces in a very aggressive way. Now each function takes the form
```
   IGL_INLINE void tet_tuple_switch_edge(
        TetTupleIterator &ttit,
        const Eigen::MatrixBase<DerivedT> &T,
        const Eigen::MatrixBase<DerivedTT> &TT,
        const Eigen::MatrixBase<DerivedTTif> &TTif,
        const Eigen::MatrixBase<DerivedTTie> &TTie)
```
where `TetTupleIterator` only contains info of four indices `t,f,e,r `. (In this way, maybe it can simply be named `TetTuple`).
If it's OK, I will do some further cleanup and also commit the triangle version, as well as the unit tests.

@danielepanozzo 
I think this type of interface, although longer to write (user can always create ad hoc lambda wrapper), is more conform to libigl style and easier to control than 


- 'storing copy of T,TT,TTif,TTie': expensive for multiple iterators on same mesh and need to re-assign when topology is changed.
- 'storing reference to matrices': unsafe when Matrices go out of scope
- 'storing shared pointer to matrices': result in unintuitive and redundant copy when initialized.

danielepanozzo:
I think this is a good solution. Every function should be on its own file to adhere strictly with the libigl guidelines, but I think for this case it makes sense to keep them all grouped together to reduce clutter.  @alecjacobson do you want to review this further or can we merge?

jiangzhongshi:
I rewrote the history of the pull request as it was getting messy.

Also, I would like to know your opinion on the following things:
1.  I have also implemented a `triangle_tuple_edge_flip` utility based on this structure.  Should I put it also in the same file as other `triangle_tuple_`'s, or a separate file makes more sense?
2. Some more advanced topological editing (also for 3D), involves dynamic change in the size of `F`/`T`,  so it would be better to have the `std::vector` representation. Therefore, which one of `std::vector<std::vector<int>>` and `std::vector<Eigen::RowVector3/4i>` would look better in libigl?

danielepanozzo:
1. A separate file would be best.
2. Definitely std::vector<Eigen::RowVector3/4i>. Nesting std::vectors is a very bad idea, since you need two memory accesses to reach an element, and in addition to this the data is not stored in a single chunk. std::vector<Eigen::RowVector3/4i> is a good choice, as long as the size of the RowVector is fixed (i.e. don't do std::vector<Eigen::RowVectorXi>).

danielepanozzo:
This looks great to me, ready to be merged.

@alecjacobson since this is a superior alternative to HalfEdgeIterator, I wonder if we should just kill HalfEdgeIterator and replace it with the new one. This will break compatibility, but not many functions are using it. What do you think?

include/igl/tetrahedron_tetrahedron_adjacency.cpp
```diff
@@ -0,0 +1,134 @@
...
+    int f;
+  };
+
+// Extract the face adjacencies
+  template<typename DerivedT, typename TTT_type, typename DerivedTT,
+           typename DerivedTTif, typename DerivedTTie>
+  IGL_INLINE void tetrahedron_tetrahedron_adjacency_extractTTi(
+      const Eigen::MatrixBase<DerivedT> &F,
+      std::vector<TTT_type> &TTT,
+      Eigen::MatrixBase<DerivedTT> &TT,
```
alecjacobson: The convention in libigl is to use MatrixBase for input types and PlainObjectBase for output types.

This avoids the need for `.derived()` calls and also prevents outputting into places that can't be resized properly.

jiangzhongshi: Ok, that's very reasonable. I'll fix that.

jiangzhongshi:
I would do some further polishing/refactoring while proceeding with my current project and re-open this PR afterward.

Issue #454, opened 2017-01-26
writeOBJ can take an arbitrary number of dimensions for vertices, but readOBJ can't


yig:
I can call `writeOBJ` on an matrix which stores any number of dimensions for vertices, and it will write every dimension out to the OBJ file. However, `readOBJ` doesn't match.

swannyPeng:
I think libigl can not draw more than 3 dimensions model

swannyPeng:
I do not think libigl can draw more than 3 dimensions model

alecjacobson:
libigl is more than a mesh viewer, so I'm not too concerned whether the viewer can handle them.

I don't see a reason why not to load these objs (except that they're obviously non-standard so we're encouraging files that may crash/fail to read in other apps). A PR for this would be merged.

PR #483, opened 2017-03-09
n-dimensional OBJ vertex support

josefgraus:
Extend readOBJ to support n-dimensional vertex lines, n >= 3

Modification in response to issue #454 enhancement request. readOBJ vertex matrix will contain columns equal to the number of components on the "v ..." .obj file line as encountered.

### DIFF
--- a/include/igl/readOBJ.cpp
+++ b/include/igl/readOBJ.cpp
@@ -14,6 +14,8 @@
 #include <iostream>
 #include <cstdio>
 #include <fstream>
+#include <sstream>
+#include <iterator>
 
 template <typename Scalar, typename Index>
 IGL_INLINE bool igl::readOBJ(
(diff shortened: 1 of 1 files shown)


include/igl/readOBJ.cpp
```diff
@@ -77,22 +79,18 @@ IGL_INLINE bool igl::readOBJ(
...
       if(type == v)
       {
-        double x[4];
-        int count =
-        sscanf(l,"%lf %lf %lf %lf\n",&x[0],&x[1],&x[2],&x[3]);
-        if(count != 3 && count != 4)
+        std::istringstream ls(&line[1]);
+        std::vector<Scalar > vertex{ std::istream_iterator<Scalar >(ls), std::istream_iterator<Scalar >() };
+
+        if (vertex.size() < 3)
```
alecjacobson: I guess if we're going to allow n>3 why not also allow n==2 (or I suppose n==1?)?  Do people use .obj to store 2D meshes? 

_I'm guessing most people are not really storing 4+d meshes but rather sneaking per-vertex data into the format, so maybe this is all in the name of hackery_.

PR #485, opened 2017-03-11
Bugfix in get_smallest_pos_quad_zero

jiangzhongshi:
I've identified two bugs in `get_smallest_pos_quad_zero` from `flip_avoiding_line_search.cpp`

When entering the `else` branch as `a==0`
1. [ line 80](https://github.com/libigl/libigl/blob/910e69482d47a127899d4c6d8cf6f86edd788807/include/igl/flip_avoiding_line_search.cpp#L80): `t1 = t2` assigned wrongly to `-b/c` instead of `-c/b`. 
Note this function solves for `0*x^2 + b*x + c = 0`.

2. Later [line 87](https://github.com/libigl/libigl/blob/910e69482d47a127899d4c6d8cf6f86edd788807/include/igl/flip_avoiding_line_search.cpp#L87) will always return to `INFINITY`  as `t1==t2` always holds after this branch. Because this condition only makes sense when there are actually duplicated roots in the case of `a!=0` and `delta == 0`. I actually fix this by absorbing this case to `delta_in <= 0` at [line 70](https://github.com/libigl/libigl/blob/910e69482d47a127899d4c6d8cf6f86edd788807/include/igl/flip_avoiding_line_search.cpp#L70)

I guess no one has encountered the case of `a==0` so it's not addressed until I make artificial test with it. So in this PR, I resolve the aforementioned bugs and also optimize some coding


danielepanozzo:
This looks great, thank you Zhongshi. It is ready to be merged. Could you please update the PR with a new version that does not have conflicts with the main branch?

Issue #484, opened 2017-03-10
Eigen python bindings: coefficient-wise unary operator call fails with an exception


BenjaminAllain:
The python bindings of Eigen's coefficient-wise unary operators (`cwiseAbs`, `cwiseAbs2`, `cwiseSqrt`, `cwiseInverse`) raise a `TypeError` exception when they are called.

Example:
```
import pyigl as igl
A = igl.eigen.MatrixXd([[1.0, 2.0], [-3.0, -4.0]])
B = A.cwiseAbs()
```
Instead of returning the coefficient-wise absolute value, the last line of the previous sample fails to execute, with the following error message:
```
TypeError: cwiseAbs(): incompatible function arguments. The following argument types are supported:
    1. (self: Eigen::MatrixBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> >) -> Eigen::CwiseUnaryOp<Eigen::internal::scalar_abs_op<double>, Eigen::Matrix<double, -1, -1, 0, -1, -1> const>

Invoked with:  1  2
-3 -4
```

I suggest the following fix #486 , which works on my configuration: in file `libigl/python/modules/py_vector.cpp`, replace
```
.def("cwiseAbs", &Type::cwiseAbs)
.def("cwiseAbs2", &Type::cwiseAbs2)
.def("cwiseSqrt", &Type::cwiseSqrt)
.def("cwiseInverse", &Type::cwiseInverse)
```
with
```
.def("cwiseAbs", [](const Type &m) -> Type { return m.cwiseAbs(); })
.def("cwiseAbs2", [](const Type &m) -> Type { return m.cwiseAbs2(); })
.def("cwiseSqrt", [](const Type &m) -> Type { return m.cwiseSqrt(); })
.def("cwiseInverse", [](const Type &m) -> Type { return m.cwiseInverse(); })
```
 . My configuration is Ubuntu 16.04, with gcc 5.4.0

teseoch:
We now use numpy, this operation is obsolete

PR #486, opened 2017-03-13
bugfix in bindings of Eigen's coeff-wise unary operators

BenjaminAllain:
Fix the bug #484 

- Instead of binding directly the operator function, which returns an intermediate type, we bind a lambda which encapsulates an explicit call and whose return type is explicit. This forces the evaluation of the expression template.


fwilliams:

PR #488, opened 2017-03-14
issue 482: add discrete exterior calculus operator

milkpku:
- add `halfedge_flip.h/cpp` to compute half-edge topology, maybe duplicate with `edge_flaps` or `edge_topology`
- add `discrete_exterior_calculus.h/cpp`, which implement DEC.

There are several difference between standard algorithms and this implementation:
- the 1-form is represented on half-edges rather than edges, since half-edges are much easier to address than edges in the (V, F) representation
- the voronoi area of vertex is approximated by 1/3 of sum of its neighbor triangle faces.

### DIFF
--- a/include/igl/discrete_exterior_calculus.cpp
+++ b/include/igl/discrete_exterior_calculus.cpp
@@ -1,0 +1,257 @@
+
+#include "discrete_exterior_calculus.h"
+#include "halfedge_flip.h"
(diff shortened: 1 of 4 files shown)


alecjacobson:
This looks like a neat project, but doesn't conform to libigl style guidelines and contains many duplicates with existing code.

PR #489, opened 2017-03-14
Incremental eigen decomposition

josefgraus:
Tweaked eigs so that k no longer needs to be defined and max_iter can be changed. If k = 0, eigs will run until max iteration failure and return found pairs. Otherwise, eigs will run until k pairs have been found, or max_iter has been hit.

Updated python bindings, comments. and the code example to reflect the signature change for eigs. The web reference will need to be updated if this change is accepted.


alecjacobson:
Adding a `max_iter` parameter makes sense, but this should be done through an overload (with all const/input parameters coming first), rather than by changing the API.

Sparse eigs really isn't meant for finding all eigen-pairs since this will just be a slow way to solve the dense problem. Or maybe I'm misunderstanding the `k=0` scenario.

josefgraus:
You're right, an overload would be better, and avoid breaking other's code.

The scenario for k=0 is to find as many pairs as possible in a limited number of iterations for arbitrary input, e.g., I'm not sure what exactly I'll be passing over so estimating k might be problematic, but I have time for a certain number of iterations per pair until failure, and the determined pairs are "good enough" approximations for the time I could spare (it has been useful in the project I'm using libigl for).

If you think it might be worth adding an overload for this, I can go ahead and make the changes in short order.

include/igl/eigs.cpp
```diff
@@ -33,8 +34,8 @@ IGL_INLINE bool igl::eigs(
   assert(iB.rows() == n && "B should be match A's dims.");
   assert(iB.cols() == n && "B should be square.");
   assert(type == EIGS_TYPE_SM && "Only low frequencies are supported");
-  DerivedU U(n,k);
-  DerivedS S(k,1);
+  DerivedU U(n, 1);
```
alecjacobson: I don't understand this change. Want this lead to errors below when calling `U.col(i)`?

josefgraus: Line 146 resizes U and S on each pair, so it doesn't go out of bounds.

alecjacobson: But why not just have the right size from the get go?

josefgraus: It would save the resize cost if `U` was just set to an nxn, `U(n,n)`, and then a block copy was returned after max_iter iterations depending on the number of pairs. I suppose the question is which is more expensive in common situations, resizing U (conservativeResize, if I recall, is not an inexpensive operation, for obvious reasons) or preallocating a large U and copying a block at the end (is `n `>> `k` always for some reasonable number of max_iter?) 

I don't mind making this change if it turns out the latter is the better approach.

alecjacobson:
We didn't get to this before our big change. Let's revisit after our big merge if necessary.

PR #494, opened 2017-03-20
Unsigned char in texture_from_png

wucm1994:
Should this be unsigned char instead of char, right? :-)


alecjacobson:
Yes, it should be. We need to synchronize this with the viewer and other places that may have made the same mistake.

jdumas:
With the new viewer, the only `char` left is in the `MeshGL` class. Should we merge this PR then?

PR #505, opened 2017-03-30
Add more sudo apt-get install commands

alecjacobson:
I needed these two on my fresh ubuntu box. Are they unusual? If not, should we add them to the general instructions?


danielepanozzo:
Let's add them if they are needed. They are both pretty old packages, it is puzzling that they are needed only now. Which ubuntu version are you running?

alecjacobson:
I'm using Ubuntu 16.04.2 LTS

PR #530, opened 2017-05-15
add while loop for comb_cross_field.cpp, making comb support separated manifold

milkpku:
I run into an assertion error when running Example 505 (505_MIQ) on my custom mesh (made of an inner surface and a outer surface), and found that comb() function does not support manifold with separated parts.

I did a minor modification: add while loop outside of the flood fill loop, and then it works well. So is it proper to merge into the main repo?


alecjacobson:
Need unit test before to check that existing comb_cross_field doesn't break.

PR #543, opened 2017-06-13
Add python bindings for igl::adjacency_list

fwilliams:
Add python bindings for` igl::adjacency_list (const Eigen::MatrixXd& F, std::vector<vector<int>>& out_list)`.

The python signature is `adjacency_list(F: eigen.MatrixXd)` and returns the adjacency list as a Python list. This return requires copying the vector, out_list, to the returned value. This is unavoidable without either (a) implementing a modified `igl::adjacency_list` to handle Python lists (hacky) or (b) exposing STL vectors as a custom type to python (ugly). Therefore, I believe the extra copy is acceptable.

### DIFF
--- a/python/modules/py_typedefs.cpp
+++ b/python/modules/py_typedefs.cpp
@@ -11,3 +11,9 @@ py::class_<RotationList>(m, "RotationList")
     .def("__iter__", [](RotationList &v) {
        return py::make_iterator(v.begin(), v.end());
 }, py::keep_alive<0, 1>());
+
+
+py::bind_vector<std::vector<int>>(m, "VectorInt");
+py::bind_vector<std::vector<std::vector<int>>>(m, "VectorVectorInt");
+
+
--- a/python/modules/py_typedefs.h
+++ b/python/modules/py_typedefs.h
@@ -1,5 +1,12 @@
+#include <pybind11/pybind11.h>
+#include <pybind11/stl_bind.h>
+
 typedef std::vector<Eigen::Quaterniond,Eigen::aligned_allocator<Eigen::Quaterniond> > RotationList;
-PYBIND11_MAKE_OPAQUE(RotationList);
+PYBIND11_MAKE_OPAQUE(RotationList)
 
 //typedef std::vector<Eigen::Vector3d> TranslationList;
 //PYBIND11_MAKE_OPAQUE(TranslationList);
+
+PYBIND11_MAKE_OPAQUE(std::vector<int>)
+PYBIND11_MAKE_OPAQUE(std::vector<std::vector<int>>)
+
--- a/python/py_doc.cpp
+++ b/python/py_doc.cpp
@@ -31,6 +31,21 @@ const char *__doc_igl_active_set = R"igl_Qu8mg5v7(// Known Bugs: rows of [Aeq;Ai
   // Benchmark: For a harmonic solve on a mesh with 325K facets, matlab 2.2
   // secs, igl/min_quad_with_fixed.h 7.1 secs
   //)igl_Qu8mg5v7";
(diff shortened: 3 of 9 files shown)


fwilliams:
@skoch9 I was told by @danielepanozzo that you were the person to contact about the libigl python bindings. It seems there is a clang script which automatically inserts documentation into the binding code. Is this correct? Should I remove my doctstring in this PR? If not, can you approve it?

skoch9:
Hi Francis,
yes, there is a script that parses the documentation and adds it to the bindings. The general steps are the following: 

1. Add the binding file to the py_igl folder with `__doc_igl_<function_name>` as the name for the documentation string.
2. Run the script `generate_docstrings.py <path_to_c++_header_files> <path_to_py_igl>`. This automatically adds everything you need to include the binding.
So you can remove/change your docstring and the includes and have everything done by the script.

With regards to the binding function itself, I would rather go for an opaque binding and add an AdjacencyList to modules/py_typedefs.* which is passed in to the function as an argument. This way we avoid the copy operation and stay consistent with the C++ function signature. This is not really pretty, but I have planned to wrap/generate bindings that are more pythonic in the (near) future anyways so it is better to have it consistent for now.

fwilliams:
So I found a few bugs in `generate_docstrings.py`. I'll include fixes in this commit. Also, @skoch9, is there a reason we're calling `clang++-3.7` specifcally to generate the documentation? It seems like we could simply use `clang++` and not couple ourselves to a specific version here. I tried it with `clang-4.` and it seems to work. 

It seems that this script hasn't been run in a long time and some functions have been moved into namespaces. For example, `ig::unproject_onto_mesh` is now `igl::embree::unproject_onto_mesh`. This means I'll have to go and update each one to get this to work. I propose that we make this script a githook and compile the python bindings with travis-CI. @danielepanozzo what do you think? Shall I open an issue?

danielepanozzo:
I would prefer to keep it simple and not set up an automatic githook for this. However, one thing that would be very useful would be to change the TRAVIS script to make sure that not only the python bindings are compiled, but also that they can be successfully loaded by python to catch some of the more nasty runtime errors. I will wait until @skoch9 approves this PR before merging it. Thanks!

skoch9:
Thanks @fwilliams for the changes. There is no reason for `clang++-3.7`, I might have made this specific due to some conflicting versions in Ubuntu but I don't remember anymore. Can you also commit your fixes to the script so I can take a look at them?

Concerning the unproject_onto_mesh function, I don't think it has been moved but rather duplicated. So we could add another one in the embree namespace, but the old one should remain as it is. So far, I mostly added functions which are called in the tutorials. For additional functions, I would like to add minimal working examples or tests whenever we create a new binding to make sure the function works. I will try to set up the structure for this in the coming weeks. 

With regards to running the docstring generation, I would also favor the non-travis version and rather add more documentation on how to contribute to the python bindings (with all steps clearly explained).
@danielepanozzo Adding an import test for the bindings would be really useful, can you do this?

danielepanozzo:
I added the import test for the bindings to travis on MacOSX and Linux. Francis, could you please run the script to generate the documentation, and update this PR (with also the changes to pydoc.h/cpp)?

fwilliams:
Sorry for ,the delay on this. I just added my fixes to `generate_docstring.py`. I updated  my commit message describing the problems I fixed. Let me know if there are any issues with my fix.

danielepanozzo:
Looks good to me. I will wait for @skoch9 confirmation before merging it.

skoch9:
Looks good to me too, though there is one minor problem: adjacency_list is imported twice in py_igl.cpp. This should not happen when running the script, otherwise we should fix this in the script as well.
Thanks!

fwilliams:
Whoops. This happened while debugging. I re-ran the script and it seems to work.

skoch9:
Perfect, thanks. We will add tests for the function in another pull request.

PR #551, opened 2017-06-27
Simplify vector_area_matrix.cpp

jmespadero:
A small change to optimize/simplify the code of vector_area_matrix.

Instead to build two sparse matrix that have to be added and scaled,
build directly the triplets for the sparse result matrix.

Also, do a reserve() before of filling the triplets list.


alecjacobson:
Thanks!

Issue #467, opened 2017-02-16
Using pyigl with other pybind11 exports


MichaelRabinovich:
Hey all,

I might be doing something wrong, but I'm having problem importing modules that I've built with pybind11 and pyigl together. After importing pyigl, calling otherwise functioning methods through another pybind11 makes my python program crush with a segmentation fault or a bus error. 

The problem seems to come from compiling my other pybind11 module with another version of pybind11.h than the one you use for libigl. If I built it with your header files some routines seems to work. However I can't use your version since it doesn't seem to include the file Eigen.h necessary for me to seamlessly call my c++ Eigen functions with a numpy array.

Any ideas what can I do?

Thanks,
Michael

danielepanozzo:
Do you use a special version of pybind11 or is it simply a newer version of the one with use in libigl? 

If it is just newer, I would suggest to send a PR to libigl updating the bundled pybind11 to the latest version (and fixing whatever problems this might introduce). You would then be able to use the same version for both libigl and your code. I am adding @s-koch  to this discussion since he is using daily the python bindings and he might want to comment on this before you go ahead and prepare the PR.

s-koch:
Hi, 
I just recently updated the version of pybind11 which is used by libigl to release v2.0 (from January 2017). However, the python bindings of libigl can also be compiled with the latest version in their repository. You can just replace it with the ext/nanogui folder.
Given your question, I am not so sure if this is really just a version problem or if it's related to libigl having its own binding of Eigen types. Can you maybe elaborate on your build process and what calls you want to make?

MichaelRabinovich:
Oh I'm sorry, for some reason I was convinced you added stuff and changed pybind11 in libigl. Updating pybind11 for libigl like you said fixed it.
Thanks!

s-koch:
Ok great! Out of curiosity, are you using the iglhelpers functions to convert between the pybind11 eigen/numpy types and libigl's eigen types?

PR #561, opened 2017-07-03
allow specification of alternate pybind11 directory

decrispell:
This PR allows the user to specify their own pybind11 directory at CMake time via the PYBIND11_INCLUDE_DIR cmake variable.   This allows users to avoid problems when importing multiple libraries that use different pybind11 versions, as discussed in #467 

The default value is the included version, so users shouldn't notice any change if they don't manually specify a new pybind11 location.


decrispell:
Thanks for the quick PR merges and excellent library! 👍

danielepanozzo:
Thank you for the PRs :)

PR #564, opened 2017-07-08
Compatibility with Eigen 3.3.4

jdumas:
Somehow this line was not compiling properly with the latest version of Eigen ...


alecjacobson:
Thanks... but yet another reason I'm not excited about 3.3.*

PR #572, opened 2017-07-19
Fix #571.

### LINKED ISSUE #571: polyhedron_to_mesh.cpp:67:139: error: ‘__1’ is not a member of ‘std’
Hi,

I get this compilation error:

> `libigl/include/igl/copyleft/cgal/polyhedron_to_mesh.cpp:67:139: error: ‘__1’ is not a member of ‘std’`

Seems like there is troubles with `CGAL`. Can you provide a solution for this?

My setup:
- `gcc 6.3`
- `CGAL 4.9`
- `Ubuntu 17.10`

skanti:
Removed the `__1` namespace that kept being unresolved.


danielepanozzo:
Thanks!

PR #575, opened 2017-07-28
Marching cubes over sparse indexed grids

fwilliams:
I also changed the edge lookup table from an std::map to an std::unordered map. I measured a 10-20% performance boost after doing this.


include/igl/copyleft/marching_cubes.h
```diff
@@ -47,7 +47,36 @@ namespace igl
...
+      //    the index of a vertex in points and a scalar in values.
+      //    i.e. points[indices[i, j]] = the position of the j'th vertex of the i'th cube
+      // Output:
+      //   vertices  #V by 3 list of mesh vertex positions
+      //   faces  #F by 3 list of mesh triangle indices
+      // Note: The winding direction of the cube indices will affect the output winding of the faces
+      //
+      template <typename DerivedV1, typename DerivedV2, typename DerivedI, typename DerivedF>
+      IGL_INLINE void marching_cubes(
+        const Eigen::PlainObjectBase<DerivedV1> &values,
```
alecjacobson: Input types should be templated from Eigen::MatrixBase (this is a new guideline and we're slowly changing old functions).

include/igl/copyleft/marching_cubes.cpp
```diff
@@ -25,38 +25,40 @@
 
 #include "marching_cubes.h"
 #include "marching_cubes_tables.h"
-#include <map>
+
+#include <unordered_map>
```
alecjacobson: I think the map --> unordered_map change will be easier to merge. Is it easy for you to put in a separate PR?

fwilliams: See #589.

include/igl/copyleft/marching_cubes.h
fwilliams: What's the reasoning behind this? Is it just that `Eigen::MatrixBase` is more general than `Eigen::PlainObjectBase`? 

In this particular case, I call `conservativeResize()` on `vertices` and `faces`, both of which aren't available to `MatrixBase`. I can do a regular `resize` but I'm allocating memory geometrically so in the worst case I'm using 2x what I need. 

I could use `PlainObjectBase` for these two parameters as a solution.

danielepanozzo:
Looks good to me. Could you please fix the merge conflict? I will merge it after the builds will pass

jdumas:
Could you also make the template instantiations one-liners?

alecjacobson:
it'd be cool to have root-finding as a function handle and linear interpolation as the default.

PR #576, opened 2017-07-31
Add a utility function which computes a sparse set of epsilon-sized g…

fwilliams:
…rid cells surrounding an input surface. The function takes a point on the surface as input and finds all those epsilon grid cells that intersect the surface using a breadth first search over the grid space.

### DIFF
--- a/include/igl/generate_sparse_marching_cubes_grid.cpp
+++ b/include/igl/generate_sparse_marching_cubes_grid.cpp
@@ -1,0 +1,152 @@
+#include "generate_sparse_marching_cubes_grid.h"
+
+#include <unordered_map>
+#include <array>
+#include <vector>
+
+
+struct RowVector3iHash {
+    std::size_t operator()(const Eigen::RowVector3i& key) const {
+        std::size_t seed = 0;
+        for (int i = 0; i < 3; i++) {
+            seed ^= key[i] + 0x9e3779b9 + (seed<<6) + (seed>>2); // Copied from boost::hash_combine
+        }
+        return std::hash<std::size_t>()(seed);
+    }
+};
+
+template <typename T>
+int sgn(T val) {
+    return (T(0) < val) - (val < T(0));
+}
+
+
+template<class DerivedM, class DerivedR>
+int matrix_push_back(Eigen::PlainObjectBase<DerivedM>& mat, const Eigen::PlainObjectBase<DerivedR>& row, int n) {
+    if (mat.rows() == n) {
+        mat.conservativeResize(2*n+1, mat.cols());
+    }
+
+    mat.row(n) = row;
+    return n + 1;
+}
+
+
+template<class DerivedM>
+int vector_push_back(Eigen::PlainObjectBase<DerivedM>& mat, typename Eigen::PlainObjectBase<DerivedM>::Scalar row, int n) {
+    if (mat.rows() == n) {
+        mat.conservativeResize(2*n+1);
+    }
+
+    mat[n] = row;
+    return n + 1;
+}
+
+
+
+template <typename DerivedS, typename DerivedV, typename DerivedI>
+void igl::generate_sparse_marching_cubes_grid(const Eigen::RowVector3d& p0, std::function<double(const Eigen::RowVector3d&)> scalarFunc, double eps,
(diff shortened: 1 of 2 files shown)


include/igl/generate_sparse_marching_cubes_grid.cpp
```diff
@@ -0,0 +1,152 @@
+#include "generate_sparse_marching_cubes_grid.h"
+
+#include <unordered_map>
+#include <array>
+#include <vector>
+
+
+struct RowVector3iHash {
```
alecjacobson: See the style-guidelines. Every .cpp should have a single function and all new functions should be in the igl namespace

alecjacobson: If these are only used once in the function below then you could consider using lambdas. If you think they'll be useful else where then add a .h/.cpp for each function.

alecjacobson:
I added a few comments to the proposed file.

include/igl/generate_sparse_marching_cubes_grid.cpp
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    mat[n] = row;
+    return n + 1;
+}
+
+
+
+template <typename DerivedS, typename DerivedV, typename DerivedI>
+void igl::generate_sparse_marching_cubes_grid(const Eigen::RowVector3d& p0, std::function<double(const Eigen::RowVector3d&)> scalarFunc, double eps,
```
jdumas: For a fixed-sized array, I doubt there will be much differences, no? You could even pass it directly as a `Eigen::RowVector3d` to avoir the reference.

alecjacobson: It's _mostly_ a style thing. This bug crops up all over the place and the best way to fit it is consistency.  For this particular case its unlikely that the cost of a cast/copy will become a bottleneck.

include/igl/generate_sparse_marching_cubes_grid.h
```diff
@@ -0,0 +1,31 @@
...
+    //      and points which are positive lie outside the surface
+    //  eps  The edge length of the cubes surrounding the surface
+    //
+    // Output:
+    //   CS  #cube-vertices by 3 list of scalar values at the cube vertices
+    //   CV  #cube-vertices by 3 list of cube vertex positions
+    //   CI  #number of cubes by 8 list of indexes into CS and CV. Each row represents a cube
+    //
+    template <typename DerivedS, typename DerivedV, typename DerivedI>
+    void generate_sparse_marching_cubes_grid(const Eigen::RowVector3d& p0, std::function<double(const Eigen::RowVector3d&)> scalarFunc, double eps,
```
alecjacobson: Can we come up with a more revealing name for this function? `generate_` can surely be dropped since all functions generate some output from some input. 

Do I understand correctly that this constructing a sparse voxelization along a component of the zero-iso-level of an implicit function?

include/igl/generate_sparse_marching_cubes_grid.cpp
jdumas: Sure, that's fine then. Btw, interestingly enough, regarding your blog post, it is something that `g++ -O2` optimizes out, while `clang++ -O3` doesn't.

alecjacobson: wow. Seriously?! I really wish that clang would optimize this out. I had convinced myself that this was "beyond the compiler's reach".

jdumas: Ikr. Both `X.size()` and `templated_size(X)` take the same time with `gcc`, while the second call to `simple_size(X)` is apparently optimized by `gcc` (about 10x faster), as `gcc` seem to reuse the result from the first call to `X.size()` (which `clang` doesn't do). But at this points it's just trivia.

include/igl/generate_sparse_marching_cubes_grid.h
fwilliams: Yes that's correct. Given an input point near the zero level set, it constructs a shell of epsilon sized cubes along the boundary of the component which the input point lies on. Naming things is hard, do you have any ideas?

include/igl/generate_sparse_marching_cubes_grid.cpp
fwilliams: Daniele had a memory corruption the other day because of this. He was assigning the const reference to another const reference in a copy constructor. It looks like the compiler made a copy of the memory and the reference pointed to garbage when the constructor finished.

PR #579, opened 2017-08-08
Added template instantiation

deuscovrigus:


deuscovrigus:
New template instantiation

PR #583, opened 2017-08-15
Bug fix on incorrect malloc(0) of pointmarkerlist struct when VM.size…

deuscovrigus:
Fix that causes Delunay triangualtion to crash by correctly allocating the pointmarker list matrix


alecjacobson:
Good catch! Thanks.

PR #586, opened 2017-08-16
Hessian energy

odedstein:
Added Hessian energy (ref. https://arxiv.org/abs/1707.04348) and fixed a wrong path in Readme.md


include/igl/hessian.cpp
```diff
@@ -0,0 +1,60 @@
...
+//  and Oded Stein <oded.stein@columbia.edu>
+//
+// This Source Code Form is subject to the terms of the Mozilla Public License
+// v. 2.0. If a copy of the MPL was not distributed with this file, You can
+// obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "hessian.h"
+#include <vector>
+
+#include <igl/grad.h>
```
alecjacobson: `<igl/grad.h>` --> `"grad.h" and so on.

tutorial/712_DataSmoothing/isolines.h
```diff
@@ -0,0 +1,90 @@
+
+#include <vector>
+#include <limits>
+#include <stdlib.h>
+
+#include <igl/remove_unreferenced.h>
+
+
+static void isolines(const Eigen::MatrixXd& V,
```
alecjacobson: This could be useful in general. It'd be cool to (eventually) pull this out and template this into a .h/.cpp file.

include/igl/hessian.cpp
odedstein: Addressed

alecjacobson:
Thanks, Oded. This looks great.

PR #587, opened 2017-08-16
Fix a wrong link in matlab-to-eigen.html

lzhbrian:
On the page of matlab-to-eigen.html, the link to tutorial.html seems to be wrong


alecjacobson:
Thanks.

PR #592, opened 2017-08-23
Matplotlib colormaps

josefgraus:
Added new perceptually uniform colormaps created for matplotlib: https://bids.github.io/colormap/

Magma, Inferno, Plasma, and Viridis are included in this PR.


danielepanozzo:
Thank you for the contribution! It seems that all files are similar, with the exception of that static array at the top of the .h. If this is the case, I suggest to merge them into a single function (that could also contain the current jet.h), that takes the type of colormap as an additional parameter. @alecjacobson : what do you think?

alecjacobson:
We should definitely factor out the duplicate code across all these files and `jet`,`parula`. 

I'd also recommend taking a look at http://colorbrewer2.org/ Unlike the historical colormaps in matlab/matplotlib these are the result of studying how people perceive colors and data.

josefgraus:
That was the original plan, but I was concerned about breaking jet in older code. I can migrate everything over to some "colormap" header, and pull in jet/parula. To keep it one function per include, there would be a simple enum to switch between colormaps. For compatibility sake, we can flag jet and parula as deprecated code via attributes ( http://en.cppreference.com/w/cpp/language/attributes ) until such time as it's appropriate to remove the older implementations.

alecjacobson:
Sounds good to me. Just have `jet` and `parula` internally call your new
function with the relevant enum value.

josefgraus:
That took an unnecessarily large number of passes to get that through the checks.. I suppose I should set up a static build environment for future work to avoid the headache, ha.

josefgraus:
Inferno
inferno

Jet
jet

Magma
magma

Parula
parula

Plasma
plasma

Viridis
viridis

danielepanozzo:
Looks great, than you for the contribution.

alecjacobson:
I am cleaning up these files to meet the libigl style guidelines. Ideally we should do this _before_ merging PRs.

PR #594, opened 2017-08-28
Making the code for AABB traversal consistent

n-taka:
Hi, I just pull the latest version of libigl and found that AABB.cpp has a compilation error below. 

```
c:\lib\n\n\extern\libigl\include\igl\AABB.cpp(404): error C2248: 'Eigen::PlainObjectBase<Eigen::Matrix<double,1,3,1,1,3>>::PlainObjectBase': cannot access protected member declared in class 'Eigen::PlainObjectBase<Eigen::Matrix<double,1,3,1,1,3>>'
```
(windows10, Visual Studio 2015, x64, WindowsSDK 8.1, libigl as header only library)

So, I edit a bit and simply remove `Eigen::PlainObjectBase<...>` in `look_right` just following `look_left`.

### DIFF
--- a/include/igl/AABB.cpp
+++ b/include/igl/AABB.cpp
@@ -401,7 +401,7 @@ igl::AABB<DerivedV,DIM>::squared_distance(
     const auto & look_right = [&]()
     {
       int i_right;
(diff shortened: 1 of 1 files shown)


jdumas:
This has been merged in #659.

Issue #595, opened 2017-08-28
error: calling protected constructor [hessian.cpp, hessian_energy.cpp]]


Luk0r:
/git/libigl/include/igl/hessian.cpp:37: error: ‘Eigen::PlainObjectBase<Derived>::PlainObjectBase(const Eigen::DenseBase<OtherDerived>&) [with OtherDerived = Eigen::Matrix<double, -1, -1>; Derived = Eigen::Matrix<double, -1, -1>]’ is protected within this context
     igl::grad(Eigen::PlainObjectBase<DerivedV>(V),
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~

same for line 37
and line 39 in hessian_energy.cpp

xinyazhang:
Also found at AABB.cpp:404
``include/igl/AABB.cpp:404:45: error: ‘Eigen::PlainObjectBase<Derived>::PlainObjectBase(const Eigen::PlainObjectBase<Derived>&) [with Derived = Eigen::Matrix<float, 1, 3>]’ is protected within this context
       Eigen::PlainObjectBase<RowVectorDIMS> c_right = c;``

Luk0r:
AABB.cpp can be fixed by changing line 404 to:
`RowVectorDIMS c_right = c; // compare line 395`

For the hessian stuff I just dropped the template instantiation at the end of the corresponding cpp files.

danielepanozzo:
Could you please file a PR with the change? It is the easiest way for us to review and integrate it in the main branch. Thanks!

alecjacobson:
I think we went through and fixed all of these.

PR #604, opened 2017-09-11
fix use of protected constructor

Luk0r:
use of protected constructor changed

### DIFF
--- a/include/igl/AABB.cpp
+++ b/include/igl/AABB.cpp
@@ -401,7 +401,7 @@ igl::AABB<DerivedV,DIM>::squared_distance(
     const auto & look_right = [&]()
     {
       int i_right;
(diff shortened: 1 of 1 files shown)


Luk0r:
partly fixing:
https://github.com/libigl/libigl/issues/595

Luk0r:
should be the same fix as in:
https://github.com/libigl/libigl/pull/594

luca-penasa:
I am hitting this same bug (for AABB.cpp)... Is there some reason why this pull cannot be accepted?

btw travis-ci seems to exit correctly from the build although here is reported that the build failed (?)

danielepanozzo:
@alecjacobson I think this could be merged, the reported travis failure seems to be just a bug in travis, the build runs successfully.

jdumas:
This has been merged in #659.

Issue #130, opened 2015-10-08
Any functions can compute geodesic distance


KevinChenGit:
Hi Daniele,

I started using Libigl two days ago. This library is just fantastic with amazingly rich features and tutorials! The viewer interface is just perfect. However, I didn't find an explicit function that computes geodesic distance. I think it should be somewhere inside the lib as it is a very fundamental function. Could you give me some instructions? Thank you very much!

Best regards,
Kevin

alecjacobson:
While we don't have explicit functions for geodesic distance it should be very easy to compute approximate distances via:

"Biharmonic Distance" [Lipman et al.] using the `ig::cotmatrix`, `igl::massmatrix` and `igl::eigs`

or 

"Geodesics in Heat" [Crane et al.] using `igl::cotmatrix`, `igl::massmatrix` and `igl::min_quad_with_fixed`.

If you're familiar with Matlab you could use the implementations of either of these in https://github.com/alecjacobson/gptoolbox

`mesh/biharmonic_distance.m` or `mesh/heat_geodesic.m`

For now, we'll consider this a feature request.

KevinChenGit:
Thank you very much, Alec!

PR #627, opened 2017-10-28
Geodesic and others

jiangzhongshi:
I added the code of computing exact geodesic as a submodule into libigl.
https://code.google.com/archive/p/geodesic/

Besides, I wrapped a python interface and did a few other minor python improvements.


alecjacobson:
This is cool. A new dependency and module is a big change though. I’d like to see more functionality before including this. Otherwise this might be better suited as a separate project that uses libigl to create a nice, easy to use wrapper.

danielepanozzo:
Hi Alec,

I don't think there is anything else from exact_geodesic that should be exposed. Zhongshi's wrapper exposes all the functionalities. I think this could be merged as is (after the build test will pass) and doing it as a submodule (similar to triangle/tetgen) I think is the way to go.

The other way to do this would be to cut&paste the code of exact_geodesic inside the cpp file of the wrapper, but it will not be a good way since the license is slightly different (MIT) and it will become very hard to maintain.

This is a very useful feature that has been requested a few months ago and I would like to merge it soon if you agree: https://github.com/libigl/libigl/issues/130

jdumas:
Speaking of which, have you also considered looking at [this SIG16 paper](http://i.cs.hku.hk/~xghan/Projects/VTP.htm) as well? (Source code available [here](https://github.com/YipengQin/VTP_source_code), albeit under which licence it's not clear).

jiangzhongshi:
@jdumas Interesting. Actually, the code seems like a modification and improvement of the version I am integrating and share the exact same interface. Ideally, that should work by exactly substituting corresponding files. 
But, yeah they didn't have a license and don't know if extensive tests are done. It's a possible thing to look at.

alecjacobson:
If it’s nothing more than a wrapper then I think it’s better as a separate
repo that depends on libigl or just matches it’s style. Ideally new
dependencies open up new functionality beyond just a wrapper.

danielepanozzo:
I disagree. Computing geodesic distances is a basic and useful feature that is convenient to have  directly integrated in the library and usable whenever is necessary without having to worry about it. Similarly to triangle, tetgen, embree, and lim that are also wrappers over other libraries. I think many people will be interested in having this directly exposed and usable from python too.  Having it as a separate repo would require some effort to use it, it will make it messy to integrate in the python wrappers. I don't really see any downside in adding one extra useful folder to external, this is a complex algorithm that is hard to reimplement. The code is pure c++, it link statically so there is no issue on compiling/linking this on multiple platforms. 

If you are worried about the few extra lines in the cmake, I think it would be possible to integrate it as a header-only, but this will make it different from all the other dependencies and I would prefer to keep things uniform.

danielepanozzo:
And I think this would be a good feature to add in the tutorial too

alecjacobson:
Minimal dependencies is a core principle of libigl's design.

We make exceptions to dependencies beyond stl and Eigen when there's a large gain in functionality.

Tetgen, triangle and embree are great examples. Tetgen and triangle have tedious C APIs and directly calling them with Eigen matrices allows a effective way to create meshes that we can call 100+ libigl core functions on.

Embree is a very sophisticated ray-tracing engine that would be difficult to match. We've used it to implement ray-mesh intersection but also other useful functions like projecting cursor movements inside a mesh or computing ambient occlusion.

LIM and cork are _bad_ examples. These are little more than wrappers to other code and AFAIK there are no other libigl functionalities that depend or could benefit from them. LIM has been superseded by SLIM in the main library. Similarly cork is a bit faster but otherwise inferior to the mesh booleans in copyleft/cgal.

This current PR (as of yet) is _only_ adding a straightforward Eigen wrapper to an unsupported library last edited in 2008 implementing a method from 1998. This does not merit another dependency. (FWIW I have many of these lying around for QSlim, El Topo, SCISM, etc. It does not make sense to add all of these to libigl lest it becomes a dumping ground for wrappers).

Currently, this wrapper is probably better (and more useful to others) as a gist.

If this PR also contained a variety of interesting functionality enabled by exact geodesics then the dependency could have more merit.

Alternatively, if this _algorithm_ implemented in libigl-style without dependency, then it would much easier to welcome it into the core library.

jdumas:
QSlim, El Topo, SCISM, etc. sound like useful wrappers to have around as well.
The problem with providing just a gist is that you still have to do the work of integrating the wrapper + the external library into your own project, so need to adapt your cmake script everytime. If this is a function you are likely to use often (such as geodesic), this might be inconvenient.

If you don't want to *clutter* the folder of external dependencies with git submodules, I can also suggest to try adding optional dependencies as CMake external projects (with [`ExternalProject_Add`](https://cmake.org/cmake/help/v3.1/module/ExternalProject.html)). That way CMake will *download*, configure and compile the external dependencies *if they are needed* (i.e. if the corresponding file is enabled and used in the project). `ExternalProject_Add` seems very similar to a git submodule, in that you can specify a SHA1 commit number to point to, but CMake does the job of fetching the project itself.

alecjacobson:
I agree the wrappers are useful. They don't belong in libigl though. The purpose of libigl is not to be a monolithic cmake build system for a bunch of geometry processing libraries. 

ExternalProject_Add would complicate things even more.

jdumas:
While the command for `ExternalProject_Add` is pretty simple to use, I also realize that it downloads projects at `make` time, which is annoying. It makes it difficult to use as a drop-in replacement for git submodules that you would enable/disabled based on an option given by the user.

PR #633, opened 2017-11-03
cgal (dynamic)+libigl (dynamic)+boost configure issue fixed and is working

gaoxifeng:


danielepanozzo:
Hi Xifeng, thanks for the PR. There a few problems:

1. Could you remove the changes to both of the external library? They should not be there.
2. The change to the tutorial cmake file is not necessary
3. I think there are some changes missing in the shared cmakefile, I don't see the lines that we added tonight

Thanks!

jdumas:
And also indent with space :-p

Btw, about the missing flags for this PR, I suggest going for something like `target_compile_options(tgt PRIVATE "/MD$<$<CONFIG:Debug>:d>")` instead, as it should work for both `Release` and `Debug` config, but also for other config like `RelWithDebInfo` and it is also a bit mot concise.

gaoxifeng:
Hi Daniele,

see my response below.

danielepanozzo:
Could you integrate the suggestion proposed by Jeremie? It would be nice to also support RelWithDebInfo.

Also, there are still unnecessary changes to:
external/embree
external/tinyxml2
tutorial/CMakeLists.txt (let's keep it by default to static)

danielepanozzo:
If you give me push permission on your branch I can also directly take care of this

danielepanozzo:
This has been replaced by #634

PR #637, opened 2017-11-06
ShapeUp projection method

avaxman:
With tutorial and demo updated.


alecjacobson:
This looks really cool. Could you please add "Input" / "Output" comments in the headers above each exposed function? See `cotmatrix.h` for an example.

avaxman:
You mean the projection functions? becuase there are such comments in the
shapeup core functions, no?

alecjacobson:
Yeah, I think those are the only ones missing (I just clicked on the first header I saw).

avaxman:
Right. Do I just do that and create another pull request?

avaxman:
hmmm pull requests update automatically, so there it is :-)

PR #667, opened 2017-12-26
Improved slim

danielepanozzo:
1. Added two generic igl functions to cache the sparsity pattern for:
- sparse matrix assembly
- weighted AtA

2. Used them to Improved the running time of SLIM (~10x on the assembly part, overall it will be a 2-4x)


jdumas:
I'm wondering if it wouldn't be faster (more cache-coherent) to sort a `std::vector<Triplet>` and then overwrite `X.valuePtr()` sequentially (if you cannot produce `V` already sorted).

danielepanozzo:
I want to avoid the sort, since it is slow for matrices with millions of entries. In this way, it is actually not even necessary to build the entire triplets vector, it is enough to save the values and the sparse_cached function will assemble the matrix correctly.

jdumas:
Yes I guess you're right :p

PR #672, opened 2017-12-28
Prevent libigl from overwriting previously existing boost options 

metorm:
A problem happens when the project imports boost before libigl and the main project requires more boost module than libigle.

In this case, libigl would overwrite the previously existing boost options, then the main project won't be able to link to its unique boost module.


jdumas:
We should probably use imported targets instead of `${Boost_LIBRARIES}` anyway. It would be a better approach.

metorm:
Anyway, this is better than unconditionally import boost again.
Does boot::target requires higher boost or cmake version?

jdumas:
It depends. Higher versions of Boost require higher versions of CMake (since the `FindBoost.cmake` is provided by CMake itself). But imported targets for Boost components have been available since CMake 3.1 if not earlier. In general, in your project it is best to use the variables `${Boost_LIBRARIES}` and `${Boost_INCLUDE_DIRS}` immediately after calling the `find_package(Boost)`, for this very reason that the variables can be overwritten by other components (this is why [modern CMake scripts](https://www.youtube.com/watch?v=bsXLMQ6WgIk) shouldn't even use variables at all, but rely on targets instead).

metorm:
Using the variables immediately after calling `find_package(Boost)` is a choice, but the route is not practical when your project is composed by several sub directories and each of the sub directories having a `CMakeLists.txt`.

jdumas:
Which is why you should use targets instead of variables. CMake's approach to dealing with variables is very brittle and doesn't scale well on a large project (this PR is an example of that).

jdumas:
I've thought about this PR again, and I don't think it should be merged.

It is expected that multiple calls to `find_package(Boost)` overrides the variables created by previous calls to the same `FindBoost.cmake`. Thus, any usage of the variables it defines should be localized around the relevant call to `find_package(Boost)`. It is up the user to use those variables before including `libigl`. Relying on those variables still being correctly defined later on in a large project is bad design and should be avoided.

You have several options to keep the CMake scripts of your large project concise: either use the imported targets defined by `find_package(Boost)` (which are not overridden by subsequent calls), or define your own interface target and use it to carry the dependencies you need to the desired subfolders.

E.g. in your root `CMakeLists.txt`:

```cmake
find_package(Boost 1.48 REQUIRED ...)
add_library(my_deps INTERFACE)
target_include_directories(my_deps INTERFACE ${Boost_INCLUDE_DIRS})
target_link_libraries(my_deps INTERFACE ${Boost_LIBRARIES})
```

And later on, in your subfolders, simply use:
```cmake
target_link_libraries(my_project PUBLIC my_deps) # will pull boost with it
```

metorm:
OK, I'll try your method ...

metorm:
@jdumas Imported targets is not working as expected, may it is also overwritten by libigl?
Any way, the  following way is proven to work well:

```
find_package(Boost 1.60 REQUIRED COMPONENTS date_time system filesystem thread chrono program_options)
if (Boost_FOUND)
    include_directories(${Boost_INCLUDE_DIRS})
    set(Boost_LIBS_FOR_PROJECT_EXE ${Boost_LIBRARIES})
else ()
    set(BOOST_ROOT "" CACHE PATH "Path to Boost root directory.")
endif ()

...

TARGET_LINK_LIBRARIES(MyEXE ${Boost_LIBS_FOR_PROJECT_EXE})

```

jdumas:
Imported targets can be problematic if your version of CMake and Boost are out of sync. This is Boost's fault for not supporting CMake, so the guys behind CMake are doing their best to patch it around...

Also, you should really not be using `include_directories()` nor variables in a modern CMake project anymore. Please watch [this video](https://www.youtube.com/watch?v=bsXLMQ6WgIk) that explains why.

Issue #383, opened 2016-10-27
glfw_objects missing when try to make projects with Xcode


ruoqingsun:
I am currently launching all the projects in Mac OS Sierra (version 10.12.1). Building with the "cmake, make" procedures works perfect. 

But when I tried "CMake -G Xcode ../" procedure to import everything into XCode project, it seems that in the building phase, no those C objects in glfw_objects or libglfw3.a is generated. 

![image](https://cloud.githubusercontent.com/assets/4344440/19774795/8d5eecfc-9c6e-11e6-8c4a-f999e00e7278.png)

And I get the following error after building failure in Xcode:
clang: error: no such file or directory: '/Users/ruoqing/Documents/GitHub/libigl/tutorial/build/libigl/glfw/src/Debug/libglfw3.a'

alecjacobson:
I'm not currently using Xcode and don't know how stable the cmake `-G
Xcode` is meant to be (though I thought it was stable). This is potentially
an issue with the way that the glfw cmake is set up. You might try first
seeing whether https://github.com/glfw/glfw compiles correctly using cmake
and Xcode this way.

danielepanozzo:
I can also reproduce this with the latest xcode and cmake. It is likely that they changed something in the new xcode build system and cmake has not been updated yet.

ruoqingsun:
I am currently using xcode 8 and cmake 3.7.0. The problem is I can successfully build the project, but when running the project, the project cannot check dependencies in the listed Makefiles to generate glfw objects. @danielepanozzo Which version of xcode and cmake are you applying in reproducing the project? How to let the xcode follow the dependencies listed in makefile?

danielepanozzo:
Sorry I was not clear. I meant that I can reproduce the *error*, I also cannot get it to follow the dependencies and I get the same problem with glfw that you have.

metorm:
I have posted a Temporary fix for this #673

jdumas:
Looks like this is an issue caused by CMake's *object* targets not playing well with the XCode generator. This is definitely a bug in CMake. In the meantime, @metorm could you try to replace (just to see if it compiles) the folder of glfw in the nanogui submodule by the latest [upstream version](https://github.com/glfw/glfw)?

jdumas:
This problem should be fixed in the latest version of libigl, as we now link against `glfw` from upstream, which is not defined as an `object` target in the CMake.

PR #673, opened 2017-12-28
 Temporary fix for the compiling problem with Xcode 

metorm:
In macOS/ Xcode, libigl failed to build with glfw3 enabled. The root cause of this problem is unknown yet (Maybe something wrong with nano-gui). This patch allows the user to use pre-installed glfw3 in the system to compile, using Xcode 8 or later.

This is mentioned in #383 several months ago and not solved yet.


jdumas:
Have you tried with a more recent version of NanoGUI? There are two recent PR in NanoGUI that seem related to this issue: [here](https://github.com/wjakob/nanogui/pull/304) and [there](https://github.com/wjakob/nanogui/pull/305). If this works out with a more recent version of NanoGUI, maybe we should instead update the NanoGUI submodule instead of patching things in a hacky way.

metorm:
Not yet. Recently I am too busy to test different build configurations ...

svenevs:
FWIW, updating NanoGUI should fix the XCode problem.  It came from NanoGUI **and** the `GLFW` submodule there (they both have the same problem: a pure object library with no source code makes XCode choke).

As a side note, you can vendor NanoGUI [as GLFW itself](http://nanogui.readthedocs.io/en/latest/compilation.html#nanogui-glfw-and-other-projects).  The key being using `$<TARGET_FILE:nanogui>` and making it a dependency.  I'm not familiar with libigl and what the options are (e.g., allow either NanoGUI or GLFW or both).  If you always build NanoGUI, then use it as GLFW.  This will circumvent the XCode problem, both from NanoGUI and GLFW.

jdumas:
Hi,

I'm closing this PR as I think it does not apply anymore. We now link directly against `glfw` from upstream, which is not defined as an object target. You can still override the version of `glfw` that is used in your project if you define the target before including libigl.

PR #675, opened 2018-01-10
added support for per_vertex colors with per_corner normals

danielepanozzo:


alecjacobson:
This seems like a no-brainer merge, but we should make sure it doesn't get lost when we eventually merge with the viewer-refactor branch

PR #679, opened 2018-01-18
Rename a variable to compile with mingw-w64

metorm:
There is a macro called `RGB` somewhere in mingw-w64(gcc7.2.0)
The macro conflicts with the variable `RGB` in writeOFF.cpp
Simple renaming solves the problem.


alecjacobson:
On principle, I don't like the idea of changing our code because of some
bad design in another library.

mingw is probably popular enough that we should do this -- however
begrudgingly

metorm:
Well, it’s just a name of a temporary variable, why so serious?
🤔😂
在 2018年1月19日，02:01，Alec Jacobson <notifications@github.com<mailto:notifications@github.com>> 写道：

On principle, I don't like the idea of changing our code because of some
bad design in another library.

mingw is probably popular enough that we should do this -- however
begrudgingly

PR #681, opened 2018-02-02
added templates and fixed ui for high dpi

teseoch:
added templates and fixed ui for high dpi (mac retina display with multiple screens)


danielepanozzo:
@gaoxifeng and @jdumas  could you please confirm that this patch does not break the viewer on windows and linux before I accept it?

jdumas:
Works fine for me on linux =)

PR #689, opened 2018-02-10
Fixed Python CMakeLists.txt for CMake 3.10.0

Gjacquenot:
`target_link_libraries` creates an error in case one do not specify PUBLIC/PRIVATE for each use of `target_link_libraries` on a target

    -- LTO enabled
    CMake Error at CMakeLists.txt:134 (target_link_libraries):
      The keyword signature for target_link_libraries has already been used with
      the target "pyigl".  All uses of target_link_libraries with a target must
      be either all-keyword or all-plain.


jdumas:
I'm surprised it doesn't produce an error even with earlier versions of CMake. I think it's an easy merge.

PR #690, opened 2018-02-10
Update principal_curvature.cpp

Gjacquenot:
This PR brings a code refactoring for principal_curvature.cpp:

- try to add const whenever possible
- use coherent std declaration
- remove unused parameters: (svd, zeroDetCheck)
- add a test not to exit program when data are invalid
- replace float with double to have a coherent code

@danielepanozzo, @alecjacobson, @jdumas   : We can discuss each bullet.

### DIFF
--- a/include/igl/principal_curvature.cpp
+++ b/include/igl/principal_curvature.cpp
@@ -105,7 +105,7 @@ public:
     }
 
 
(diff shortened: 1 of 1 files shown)


danielepanozzo:
Thank you for the patch!

PR #696, opened 2018-02-15
Viewer refactor

danielepanozzo:
Major viewer patch:
1. Multiple meshes are now supported in the viewer
2. Splits the viewer in multiple folders to respect the dependencies rule
3. Nanogui is replaced by imgui
4. All nested dependencies in nanogui are now in the external folder
5. the menubar is an external plugin
6. The python bindings have been updated to pybind11 and work correctly with the above changes


alecjacobson:
Ack. This is not merging nicely with the recent PR merge.

PR #698, opened 2018-02-16
Exact Geodesic

jiangzhongshi:
Modification of #627 
added the code of computing exact geodesic as a .h/.cpp into libigl.
https://code.google.com/archive/p/geodesic/


jdumas:
I'd suggest replacing the `std::auto_ptr<>` by `std::shared_ptr<>`, and maybe `M_PI` by `igl::pi`?

danielepanozzo:
yep, I did the two changes and added a tutorial for this function. I think it is ready to be merged (if the checks are all green)

PR #699, opened 2018-02-16
Fixes #693

### LINKED ISSUE #693: Cannot open include file: 'gmpxx.h': No such file or directory - Static Library Build (Win 10)
When attempting to build LibIGL as a static library, a get a significant number of errors relating to the igl_cgal project, complaining about not being able to find gmpxx.h in my CGAL installation directory.

I believe the issue is that LibIGL is making the presumption that gmpxx.h is included with all CGAL builds. However, for the windows installer gmpxx.h isn't included.

[http://cgal-discuss.949826.n4.nabble.com/Can-t-locate-gmpxx-h-included-in-CGAL-gmpxx-h-td4655254.html](url)

I have previously (in the past 3 months) built the CGAL component of LibIGL successfully with the same installation of CGAL, so I am presuming some much recent update to LibIGL has caused the lack of gmpxx.h to become an issue.

jdumas:


include/igl/copyleft/cgal/assign_scalar.h
```diff
@@ -51,6 +53,7 @@ namespace igl
         const CGAL::Exact_predicates_exact_constructions_kernel_with_sqrt::FT & cgal,
         float& d);
 
+#ifndef WIN32
```
alecjacobson: Why is this a non-windows feature? In general, we sure try to make everything available to everyone.

jdumas: See discussion in #693. Apparently `<gmpxx.h>` is not distributed with CGAL on Windows (I also tried to compile it on my side and it was complaining about this part).

danielepanozzo: gmp is compiled with mingw on windows. Mingw c binary objects can link directly to VS binaries, but cpp objects cannot (idk why). It is thus not possible to use gmpxx directly, you will have to compile it separately with VS. I suggest to try to not use gmpxxin libigl, and stick to the standard c functions in gmp instead.

jdumas:
Wait so you're saying that when you compile a .cpp file with mex on Windows it doesn't define `WIN32`? I'm actually not sure where the `WIN32` is coming from, because the [documentation](https://msdn.microsoft.com/en-us/library/b0084kay.aspx) of MSVC only mention the macro `_WIN32` and `_WIN64`.

alecjacobson:
Yes, according to @oqilipo (I don't have windows to try)

jdumas:
As for making it work on Windows, there is a [fork](http://www.mpir.org/index.html) of gmp that has a Windows build, and well as a copy of [gmp](https://github.com/BrianGladman/mpfr) with Windows build instructions. There was a [discussion](https://github.com/CGAL/cgal/issues/439) on CGAL about supporting MPIR, so maybe it is compatible now.

Either way, I'm not really keen on spending the time to hook up a build of MPIR/MPFR into our CGAL build for Windows, as it's probably gonna be a time consuming task... But if someone wants to do it, that'd be a nice addition :D

Issue #692, opened 2018-02-12
Don't mix camera motion and model matrix in the viewer


jdumas:
Hi there,

I've been playing around a bit with the multi-mesh viewer and the 3D manipulation gizmo from [ImGuizmo](https://github.com/CedricGuillemet/ImGuizmo). It's very much WIP for now, but I think it already looks pretty good:

![out](https://user-images.githubusercontent.com/578702/36080522-69eabc78-0f5f-11e8-8f85-d0ce81e66c25.gif)

While setting this up, I've been bugged by the fact that the libigl viewer, when rotating the camera, is *updating the model matrix* to reflect the change of the scene viewpoint. This is wrong and it makes it difficult for me to implement this slicing properly on scenes with different sizes.

I know that the viewer is not a modeling tool, but I think this inconsistency could be fixed pretty easily, and it just makes everything very confusing. So basically, the correct approach to implement should be:

- Use the `model` matrix to rescale/translate the scene to the unit bbox (no rotation).
- When zooming/panning/rotating the camera, apply those transformations to the `view` matrix.
- Also remove the `view * light_position_world` in the shader, as the light position should be given directly in camera space.

Those are all pretty small changes, but I just can't figure how to zoom properly with the perspective camera in the current viewer.

Anyway, this would enable me to use the gizmo shown above on scenes with different mesh sizes, so I think it'd be great to have (the current code is in the [slicing](https://github.com/jdumas/libigl-example-project/tree/slicing) branch on my fork of the `libigl-example-project`).

alecjacobson:
This seems to be a duplicate in many ways with https://github.com/libigl/libigl/issues/617 .

We should have a meaningful camera matrix, but no per model matrices. We've exposed the shaders so any more complex application built on top of the viewer can add this feature individual. But the basic viewer needs to be, well, basic.

jdumas:
I suggest to keep the model matrix to set the scale/translation of the whole scene though. We don't need to have one model matrix per model i agree, but it's be nice to be able to assume the camera is always visualizing and rotating around a scene that fits in a unit box centered around its centroid. And rescaling each mesh manually to fit into the unit box would be quite cumbersome.

alecjacobson:
I see no reason to assume the camera is looking at a unit box...

jdumas:
[...]

Issue #617, opened 2017-10-10
"Camera" in viewer is constant


alecjacobson:
There seems to be some stale "camera" member fields in the ViewerCore:

`ViewerCore::camera_eye`,`ViewerCore::camera_center`,`ViewerCore::camera_up`

These never change. Instead the view of the mesh is altered via `ViewerCore::model` by manipulating `ViewerCore::camera_zoom`, `ViewerCore::trackball_angle`.

As we switch to a multi-mesh viewer, we should strip out this "model"-based transformation. We shouldn't be maintaining a separate transformation per-model (the viewer is not a modelling tool).

It would also be nice to have a easy way to grab the camera position relative to the mesh (without having to jump through `ViewerCore::model`, `ViewerCore::view`, etc).

avaxman:
I noticed this as I struggled to retrieve the camera parameters (for instance, to export rendering file for POVRay). It would be nice to get a function that returns them.

alecjacobson:
We're working on a refactor of the viewer right now and this is one of the
issues that will be resolved.

trnglt:
I'm trying to follow where things are now. What's the recommended way to retrieve and set the camera view programmatically now? Thanks!

PR #700, opened 2018-02-17
Remove model matrix from the viewer.

jdumas:
As discussed in #692 and #617, this removes the model matrix from the viewer, which created more confusion than necessary. Also renamed `model_zoom` and `model_translation` to `camera_base_zoom` and `camera_base_translation` respectively (it's best to keep them separate from `camera_zoom` and `camera_translation` to keep the distinction with the user interactions). I also removed `model_zoom_uv` and `model_translation_uv` which were not used anywhere.

I've updated the shaders with the light position accordingly, and the python bindings should be good too.


danielepanozzo:
Looks good to me. @alecjacobson can you merge if you also approve?

jdumas:
I just have a small doubt. Shouldn't the `normal_eye` in the shader be transformed according to the inverse transpose of the `view` matrix?

Also, it looks like the specular coeff is computed using Phong shading (with the reflected ray from the camera). Shouldn't we use Blinn-Phong instead (with the half-vector instead)?

jdumas:
Also, I was wondering if we should make the trackball/two axis rotation centered around the scene (i.e. apply `camera_zoom` and `camera_translate` after the rotation). Right now when you pan the view and rotate the model, the rotation center is displaced and this is a really not intuitive to me.

jdumas:
Alright I've pushed the change for the normal transformations, and also updated the shaders to normalize `normal_eye` also in the fragment shader (was only normalized in the vertex shader so far, maybe this can be removed).

Now the shading should be correct (you can check on tutorial 201 for reference). Let me know if I should rename the `view` and `proj` matrix in the shader to be consistent with `mat4 normal_matrix;` (`normal` was obviously already taken).

There is also a funny bug in tutorial 201 (which was already present before), if you press <kbd>3</kbd> and then <kbd>2</kbd>, weird stuff happens.

danielepanozzo:
@alecjacobson This is also ready to merge. Can we proceed?

jdumas:
Hi there. This PR has been pending for a while. Just sending a reminder as I'd like to merge this one before #748.

PR #701, opened 2018-02-17
Continuous Integration for CGAL on Windows

jdumas:
This PR enables CI of CGAL for Windows/AppVeyor. If you look at the [AppVeyor log messages](https://ci.appveyor.com/project/jdumas/libigl/build/1.0.59/messages), you will see that it is building tutorials 609 and 610.

The main change is that it adds a `CMakeLists.txt` in `external/` that allows you to build CGAL (and Boost) as a CMake external project. This makes it possible to build CGAL within AppVeyor, without adding it as a submodule, which we probably don't want. This CMake script also downloads the GMP and MPFR dll (this is what the CGAL installer does on Windows), and puts them at the right location. Finally, CGAL and Boost are built statically, and only the GMP/MPFR dll needs to be copied around for running a program.

List of changes included in this PR:
- Added a `CMakeLists.txt` in `external/` to build CGAL+Boost automatically (intended for Windows users, but should also work on Linux/macOS).
- Updated the `external/README.md` accordingly.
- Updated AppVeyor bulid to Visual Studio 2017.
- Updated `FindCGAL.cmake` to look for CGAL in the `external/cgal` folder.
- Fixed some missing template instantiations in `igl/compine.cpp` `¯\_(ツ)_/¯`
- Consolidated `libigl.cmake` accordingly. In particular, it will set `BOOST_ROOT` to `external/boost` if the folder exists, prior to calling `find_package(Boost)` and `find_package(CGAL)` (which calls FindBoost I think).
- Updated the CMake helper function `igl_copy_cgal_dll(target)` to copy CGAL's dll + also GMP/MPFR's dll into the directory where the target is supposed to land. This means Windows users don't even have to copy the dlls around by hand =)

A note about the Boost version:
- Boost 1.66 is not yet supported by CMake (as of 3.10.2), due to a [change](https://github.com/boostorg/build/issues/273) in Boost's library naming scheme.
- Boost 1.65 had an issue on my Windows machine, some [weird stuff](https://github.com/boostorg/build/issues/230) with environment variables.
- Boost 1.64 and earlier do not support Visual Studio 2017 properly...

In the end, the problem I had with Boost 1.65 had a pretty simple [fix](https://github.com/boostorg/build/commit/9821622630035c0d7c3f4d7c196f847a335da2b6), and this was the less troublesome version of Boost to patch so that I could build it as an external project with CMake on Windows. However, CMake doesn't have an implementation of Unix's `patch`, so the only way I could patch this file `os.jam` was to have CMake copy the corrected version for me on Window. This is why there is this file in `external/os.jam`, that I didn't really know where to put, but I'm open to suggestions =)


danielepanozzo:
This is awesome! I think we should merge immediately, I don't see any problem or downside. @alecjacobson can you also take a look and merge if you agree?

jdumas:
The last build on AppVeyor seemed to have failed because of a time limit, but I [re-triggered the build](https://ci.appveyor.com/project/jdumas/libigl/build/1.0.70) on my branch (same commit) and it worked fine (compiled in 30min), so I think it can be merged.

danielepanozzo:
I re-triggered the build. Let's merge it if it is successful.

PR #708, opened 2018-02-23
Adding 2D version of Simplicial Complex Augmentation Framework for Bijective Maps into libigl and a tutorial

jiangzhongshi:
- `scaf.h/scaf.cpp` file pair implementing the bijective parameterization.
- `711_SCAF/main.cpp` contain a simple demo
- Texts into `tutorial.md`
- Some modifications and fix in the tutorials

During insertion new tutorial into 711, I added these two lines to the `711_SLIM/CMakeLists.txt`, so the project name goes with the directory name. If you think this is a good practice, I can change all of them.
```
get_filename_component(PROJECT_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME)
project(${PROJECT_NAME})
```


jdumas:
Why did you update the OpenGL version in the doc to 4.1? The minimum version of OpenGL for the viewer should still be 3.3 (see [options](https://github.com/libigl/libigl/blob/0307caa66de1ff1488ae88bcac3ce6b9ebfed721/external/glad/include/glad/glad.h#L16) of the glad loader and [hint](https://github.com/libigl/libigl/blob/0307caa66de1ff1488ae88bcac3ce6b9ebfed721/include/igl/opengl/glfw/Viewer.cpp#L144) given in the viewer).

Since you're suggesting to update the CMake files for the tutorials, we could also get rid of the first line `cmake_minimum_required(VERSION 2.8.12)` that doesn't seem to be necessary.

jiangzhongshi:
That's my bad. I just thought that paragraph would be out-of-date so I changed it to what the console tells me the version I am using.

For CMake in tutorials, I can do that too.

alecjacobson:
This changes way too many files. Could you try again but only commit changes to the relevant files mentioned in the description?

jiangzhongshi:
The reason for such large number of files is that I changed every first two lines of the `CMakeLists.txt` in the tutorial section, so the code inside isn't dependent on the numbering like `102_DrawMesh`.

I can separate them into two PR. One dealing with SCAF (just roll back one commit will do), another one changing all the `CMakeLists.txt`, if you all think it's a good change.

alecjacobson:
Thanks. Yes. In general, it's always easier to separate into multiple PRs.
-Alec

PR #710, opened 2018-02-24
Update multi-mesh tutorial

jdumas:
- Color selected mesh in red.
- Shows how to extend `ViewerData` to store arbitrary attributes (here the default mesh color), without having to do the book-keeping yourself.


alecjacobson:
I like the selection color. But I’d rather wait on this attributes stuff.

jdumas:
If you have a simple way to implement it otherwise, I'm happy to hear it. Imho this is a typical use case where custom attributes are useful.

danielepanozzo:
The attribute stuff is optional, I don't think it harms having it. I vote for keeping it (as in this PR) for power users that wants to use it (it is convenient if you have many meshes and you don't want to manually handle the properties externally), but avoid it in the tutorials since it makes them harder to read. It is simple to have the same effect using global variables for these simple cases.

alecjacobson:
Let's at least keep the attribute class out of the tutorial for now since it's off topic. I'd merge a PR for the color selection separately (but this is really low priority).

jdumas:
Here is a version without attributes in the tutorial.

danielepanozzo:
Looks good to me.

jdumas:
I've updated the PR to remove the Attribute stuff, in favor of a unique id field in the `ViewerData` class. The problem is that right now, there is no "safe" way for the user to identify a mesh, since indices are shifted when a mesh is erased from the viewer. Adding a unique non-decreasing id should help distinguish these cases, and provide the user a safe way to identify a mesh.

If we do this, then we do not need the Attribute class anymore, and the user can safely store attributes in a `std::map<int, Foo>`, `std::unordered_map<int, Foo>`, or `std::vector<std::pair<int, Foo>>` however he chooses.

The only problem is if a user adds/delete more than 2^32 meshes, then the id might collide ... but I think that's a pretty extreme use case.

List of changes:
- Deleted file `Attribute.h` as it is no longer needed.
- Add a `int id;` field in the `ViewerData` data class.
- `append_mesh()` now returns the id of the newly inserted mesh. I can change it back, but I do not really see the use case for this function returning the size of the `data_list` vector.
- Added a `size_t mesh_index(int id);` function in the Viewer, to retrieve the index from a unique id. Since the indices are stored in a `size_t`, I was not sure what to do if such the id is not found, so for now it returns 0.
- Mulit-mesh tutorial with selection color has been updated to use this unique id.

danielepanozzo:
I like this solution, it is simpler and more elegant. The conflict after 2^32 meshes is not an issue, you will probably run out of opengl buffers before reaching a conflict :)

jdumas:
I was thinking about creating and deleting a lot of meshes (while keeping the total number of meshes in the viewer low). But yes it is unlikely to be a real issue anytime soon =)

PR #723, opened 2018-03-15
Change CMakeLists.txt in tutorials and some fix

jiangzhongshi:
This PR includes the following
- Changing the first two lines of the CMakeLists.txt with a uniform way, so switching tutorial orders or adding more, would be easier. 
```
for i in $(find . -name "CMakeLists.txt" -mindepth 2); do sed -i '1d; 2s/.*/get_filename_component(PROJECT_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME)\nproject(${PROJECT_NAME})/' $i; done
```
- Some fixes in tutorial.md and html, replacing nanogui with imgui

### DIFF
--- a/tutorial/101_FileIO/CMakeLists.txt
+++ b/tutorial/101_FileIO/CMakeLists.txt
@@ -1,5 +1,5 @@
-cmake_minimum_required(VERSION 2.8.12)
-project(101_FileIO)
+get_filename_component(PROJECT_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME)
+project(${PROJECT_NAME})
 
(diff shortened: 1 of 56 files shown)


jdumas:
Thanks!

The viewer menu tutorial has been updated to provide lambda-function callbacks that can be used to more easily override the default menu (`callback_draw_viewer_(window|menu)` and `callback_draw_custom_window`). There are still virtual functions that can be overridden in a derived class for a more advanced usage, but it doesn't need to be in the tutorial.

Would you mind updating the tutorial.md accordingly?

jiangzhongshi:
Yes, it's much better this way.

PR #725, opened 2018-03-15
Dirac Operator

jiangzhongshi:
I have implemented the Dirac operator, along with the corresponding binding, do you think it will be helpful here?


include/igl/dirac.h
```diff
@@ -0,0 +1,47 @@
+// This file is part of libigl, a simple c++ geometry processing library.
+// 
+// Copyright (C) 2018 Zhongshi Jiang <jiangzs@nyu.edu>
+// 
+// This Source Code Form is subject to the terms of the Mozilla Public License 
+// v. 2.0. If a copy of the MPL was not distributed with this file, You can 
+// obtain one at http://mozilla.org/MPL/2.0/.
+#ifndef IGL_DIRAC
```
stewartdent: Should be IGL_DIRAC_H (see style guidelines)

alecjacobson:
I think this a great addition. Is D = DA' ? If yes, then I'd leave this out or provide a version that only computes D as that's a common case. It'd be great if the quaternion convention matched that of other libigl functions (I've not checked this) and (if possible) the D matrix matched the one from gptoolbox's dirac_operator.m

In gptoolbox, I used dirac_operator.m instead of dirac to avoid confusion with dirac delta. We could consider the same here.

jiangzhongshi:
No. It's not.
DA=M_V^{-1} D^H M_F

1. There is a change in vertex-area weighting and face-area weighting.
2. Also, D^H != D^T as here I use the 4x4 matrix in place of quaternions, each 4x4 block is transposed due to quaternion conjugation q^H.

alecjacobson:
Ah, right. Now I remember.

So, I'd recommend splitting this up. For example,
 - dirac_operator(V,F,D)
 - (dirac_)adjoint(D,M,DA)
 - conjugate_transpose (D,DH)

The definition of DA depends on the choice of per-vertex mass matrix. Your current implementation assumes the "barycentric" mass matrix.

jiangzhongshi:
Sure, I am only worried it might be a performance hit and ugly, during the conjugation.

What about
* dirac_operator(V,F,D)
* dirac_operator(V,F, D, MassMatrixType, DA)

alecjacobson:
I'd rather like to expose more reusable functions. conjugate_transpose is useful in general.

Building the conjugate transpose of a specific operator during its construction is a specific type of possibly premature optimization. It's not even clear that it will be faster.

For callers that only want D, then the proposed api either 1. has duplicate code or 2. has unnecessary computation (needlessly slow). In these cases, err on the side of less duplicate code and replacing unnecessary code with modular, smaller, more reusable functions.

jiangzhongshi:
Actually, I am not really sure about how to do a proper quanternion_conjugate_transpose in this case.

For the matrix D: 4F by 4V, I will have to identify the 4x4 blocks and transpose them individually, followed by a complete transpose. If there is an easier way of achieving this, then splitting would be done.

jiangzhongshi:
And for the api I proposed, specifying some default parameter would avoid both problems.

jiangzhongshi:
This might not be really useful here since Eigen already have quaternions and this was just a workaround for the use of Python.

PR #748, opened 2018-04-13
Multiple viewport

Roipo:
I made a very minimal support for multiple viewports by making several ViewerCore's similar to how multiple data's are implemented now. However it requires changing viewer.core to a pointer. Is that ok, or maybe there is a better workaround?

### DIFF
--- a/include/igl/opengl/ViewerCore.h
+++ b/include/igl/opengl/ViewerCore.h
@@ -93,6 +93,9 @@ public:
 
   // ------------------- Properties
 
+  // Unique identifier
+  int id;
+
   // Colors
   Eigen::Vector4f background_color;
 
--- a/include/igl/opengl/ViewerData.cpp
+++ b/include/igl/opengl/ViewerData.cpp
@@ -32,6 +32,8 @@ IGL_INLINE igl::opengl::ViewerData::ViewerData()
   shininess(35.0f),
   id(-1)
 {
+  tex_col1 << 0.368f, 0.477f, 0.933f, 1.0f;
+  tex_col2 << 1.0f, 0.903f, 0.649f, 1.0f;
   clear();
 };
 
@@ -437,19 +439,34 @@ IGL_INLINE void igl::opengl::ViewerData::grid_texture()
 
   unsigned size = 128;
   unsigned size2 = size/2;
-  texture_R.resize(size, size);
-  for (unsigned i=0; i<size; ++i)
+
+  texture_R.resize(size, size); texture_R.setZero();
+  texture_G.resize(size, size); texture_G.setZero();
+  texture_B.resize(size, size); texture_B.setZero();
+
+  Eigen::Matrix<unsigned char, 3, 1> t1 = (255.0f * tex_col1.block<3, 1>(0, 0)).cast<unsigned char>();
+  Eigen::Matrix<unsigned char, 3, 1> t2 = (255.0f * tex_col2.block<3, 1>(0, 0)).cast<unsigned char>();
+  int b = 2;
+  for (unsigned i = b; i<size - b; ++i)
   {
-    for (unsigned j=0; j<size; ++j)
-    {
-      texture_R(i,j) = 0;
-      if ((i<size2 && j<size2) || (i>=size2 && j>=size2))
-        texture_R(i,j) = 255;
-    }
+	  for (unsigned j = b; j<size - b; ++j)
+	  {
+		  if ((i > size2 + b && j < size2 - b) || (i < size2 - b && j > size2 + b))
+		  {
+			  texture_R(i, j) = t1(0);
+			  texture_G(i, j) = t1(1);
+			  texture_B(i, j) = t1(2);
+		  }
+		  if ((i < size2 - b && j < size2 - b) || (i > size2 + b && j > size2 + b))
+		  {
+			  texture_R(i, j) = t2(0);
+			  texture_G(i, j) = t2(1);
+			  texture_B(i, j) = t2(2);
+		  }
+	  }
   }
 
-  texture_G = texture_R;
-  texture_B = texture_R;
+
   texture_A = Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic>::Constant(texture_R.rows(),texture_R.cols(),255);
   dirty |= MeshGL::DIRTY_TEXTURE;
 }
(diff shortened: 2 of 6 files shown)


danielepanozzo:
Thanks for the PR! This seems a potentially useful feature, a few comments:
1. For consistency, it should be done in the same way as data. The cores should be stored in a std::vector, and the active one accesses with a method core()
2. Can you provide an example of how to use this feature in a tutorial? 
3. It is important that the behavior for standard users does not change (or changes minimally, like using .core() instead of .core). Is this the case?

Roipo:
I made the changes you suggested and added a tutorial (108_MultipleViews)

danielepanozzo:
Nice, I think the API looks nice and clean:
https://github.com/libigl/libigl/pull/748/files#diff-54b9872d360425fcddb1e92ce118a529

 @alecjacobson @jdumas what do you think? There is some work still required to make it pass all the build tests but before doing we should agree on the API.

jdumas:
I think this is a cool feature. I just have a couple of questions. Also, I would like to merge #700 before doing further changes to the viewer (#700 has been sitting around for a while, waiting a blessing from @alecjacobson, and might conflict with this PR).

My main concern is that the current `std::set<std::pair<int, int>> coreDataPairs` should be either:
- A `std::vector<std::pair<int, int>>` instead (no need to use a `std::set`)
- Or, maybe better, a `std::multimap<int, int> core_data`, or maybe a `std::unordered_multimap<int, int> core_data` (we could also name it `visible_data` or `visible_core_data` to make it more explicit?).

And more importantly, if `coreDataPairs` stores the `id` of the data objects, then the draw loop should also use those ids to draw the corresponding data (via `mesh_index(id)`), otherwise this will break as soon as we remove a mesh. However, this means that the draw loop will have a quadratic complexity in the number of `data` objects used in the viewer.

Other comments:
- Why the introduction of `tex_col1, tex_col2` in `ViewerData`?
- For the tutorial example, you could make one the view 2D (by using `ROTATION_TYPE_NO_ROTATION` + ortho camera). We could even have a split view with 4 viewport and 2D views for each axis, but that might be too much for a tutorial example.

PS: A couple of remarks on the style:
- Indentation should use two spaces instead of tabs.
- Variable names should user lower_case instead of cameCase (e.g. `coreDataPairs`, `windowWidth`).

danielepanozzo:
@jdumas yes I like this solution. Limiting the number of viewport to 32 is not a real limitation, I doubt that anyone will ever need more than 4. The important criteria here is that no matter what advanced features we have, it should be possible to use the viewer in the standard mode (1 mesh in 1 viewport) without having to write any additional code.

jdumas:
`mesh.set_visible(bool value, int core_id = 1);` ? (So you can use it without a `core_id` for single viewport).

Roipo:
👍

Roipo:
Something like this?
btw, why not use map for data_list, to avoid long lines like:
`viewer.data_list[viewer.mesh_index(meshID)]` ?

Roipo:
I thought calling `data_list[mesh_index(mesh_id)]` can be replaced by a map `data_list[mesh_id]`, but anyway, now it's implemented like you suggested and is good enough imo.

jdumas:
Thanks, this is looking much better now!

There are still a few issues, some of which I mentioned above:
- Indentation should use 2 spaces rather than tabs
- Variable names should use `lower_case` style.
- I still do not understand the purpose of the `Eigen::Vector4f tex_col1, tex_col2;` that was added to `ViewerData`?
- `selected_core_id` should be initialized to 1 rather than 0.
- I am wondering if the draw loop should iterate over meshes and then viewports (which is done now), or over viewports and then meshes (which would be more logical?). I don't think it really matters as performance is not the primary concern though.
- `append_mesh()` and `append_core()` should use `set_visible(true, id)` instead of the low-level bitwise operations.
- And of course, it needs to pass CI & resolve conflicts with the main branch.

Roipo:
Should be good now.
The purpose of `Eigen::Vector4f tex_col1, tex_col2` is not related to multiple views and I can reverse that commit. I made it because right now, for the texture grid, there is only one color, and the other color is black. `tex_col1, tex_col2` can be used to get two different colors.

jdumas:
Yes I think this should be a different PR then, best not get confused =)

jdumas:
Sorry there are still some tabs left :p (in tutorial 108 and Viewer.h at least).
And could you also update the tutorials which are not compiled by default (e.g. ViewverMenu, Boolean, AmbientOcclusion, etc.)?

Roipo:
Sorry for this, and thanks for being so patient! My editor automatically tabified everything. I believe I did update all the tutorials
Also, there is no 'selected_core_id', only 'selected_core_index', and it does need to be initialized to 0.

jdumas:
No worries, that's why we have code reviews =)
@alecjacobson and @danielepanozzo will probably have other remarks, but hopefully this has been whittled down to a more reasonable status.

And you're right for `selected_core_index`, idk why I thought it was the `id` and not the index that was stored.

danielepanozzo:
Looks great! Only two more comments before we merge:
1. On MacOSX, it does not work when I make the window full screen. If the window is smaller than the entire screen it works perfectly. Is this happening also on windows/linux or is it a mac specific bug?
2. Why is the file include/igl/diag.cpp modified? I would prefer to have this in a separate PR if it is not strictly necessary

jdumas:
For pb 1), it's because the viewports are not resized automatically to the window size when there are more than one viewports. But maybe the tutorial could be modified to account for that? Maybe by specifying that viewport A should take 50% of the screen in X, and viewport B should take the 50% on the right? Instead of using absolute number of pixels as is currently done.

Roipo:
Hmm, there is no `callback_post_resize`. How should it be implemented?

Roipo:
How about keeping only one `Eigen::Vector4f viewport`, but adding and enum
`enum ViewportPosition
{
  VIEWPORT_POS_ABSOLUTE = 0,
  VIEWPORT_POS_RELATIVE = 1,
};`

jdumas:
Not sure I like this better, as this would imply a need to check everytime you use `viewport` whether it is relative or absolute. On the other we probably wouldn't even need the enum : if the component is between 0 and 1 (float) it is relative, otherwise it is absolute. However, I find the approach with two member variables more explicit and less error-prone (and the absolute viewport really only needs to be updated when the window size changes) . But that's only my opinion, maybe Alec or Daniele will have a different one.

danielepanozzo:
I think adding the callback is a cleaner solution. It should however go in a separate PR, no reason to hold this up for this problem if it is only due to a fixed resolution in the code (which is totally fine for a tutorial). I think it is good to go, @alecjacobson can you take a look and merge it if you also approve?

Issue #758, opened 2018-05-03
Runtime error in tutorial Example 607


majia67:
I use the latest version of libigl and have updated its submodules. The build is successful, but an error is raised when the example starts to execute:

> The code execution cannot proceed because igl_stb_image.dll was not found. Reinstalling the program may fix this problem.

I guess there is a bug in the CMakeLists.txt file associated with stb_image. Could someone take a look at this as I need to use PNG handling?

Testing environment:
OS: Windows 10 x64, 1709
IDE: Visual Studio Community 2017, 15.6.7
Build: x64 Release

jdumas:
Can you try removing the `SHARED` from [this line](https://github.com/libigl/libigl/blob/a16f338ca2b0781a34bdd5dcc8ef7447ffa093b6/external/stb_image/CMakeLists.txt#L4) and see if it helps?

majia67:
Thanks, @jdumas ! Removing the `SHARED` works for me!

majia67:
@jdumas Do you think I should submit a pull request to fix this bug?

PR #765, opened 2018-05-05
fix stb_image module loading

majia67:
This PR fixes the issue mentioned in https://github.com/libigl/libigl/issues/758

@jdumas


jdumas:
Thanks!

Btw I remember why I put the `SHARED` here in the first place: it was because the symbols visibility has been set to `hidden` by default. The goal was to prevent a conflict with the `stb_image` used in NanoGUI. I think this can be reverted now that NanoGUI is no longer a dependency. So I will need to look a bit more closely before merging it.

jdumas:
Fixed by #830

PR #768, opened 2018-05-07
fix a bug in remesh_intersections

YijingLi-Gene:
Hi!

I'm a student of Professor Jernej Barbic.
My new project is dependent on the mesh boolean routine in libigl and I have found a bug in remesh_intersections.cpp. I have fixed it and created a pull request here.

Previously in the function, if a new vtx is visited when stitch_all==false, it does not handle the case when the vtx is in the interior of a face. This commit fixes that.

### DIFF
--- a/include/igl/copyleft/cgal/remesh_intersections.cpp
+++ b/include/igl/copyleft/cgal/remesh_intersections.cpp
@@ -148,6 +148,8 @@ IGL_INLINE void igl::copyleft::cgal::remesh_intersections(
     std::vector<Index> source_faces;
     std::vector<Point_3> new_vertices;
     EdgeMap edge_vertices;
+    // face_vertices: Given a face Index, find vertices inside the face
+    std::unordered_map<Index, std::vector<Index>> face_vertices;
 
     // Run constraint Delaunay triangulation on the plane.
     // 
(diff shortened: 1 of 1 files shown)


alecjacobson:
This is surprising. Can you post the example that causes remesh_intersections.cpp to fail? 

To change this function we will need to retest on a large benchmark to make sure new bugs are not introduced. @qnzhou

YijingLi-Gene:
Let me prepare the example.
I have also found that all the usages of remesh_intersections() in libigl pass the parameter with stitch_all==true, including the mesh_boolean code. This bug only happens when stitch_all==false. This is probably why this bug has not been found before.

YijingLi-Gene:
Here is the example. 
It includes the code and the model. The model is a self-intersecting obj mesh. 
Without the patch, remesh_intersections creates a mesh which is not vertex-manifold when stitch_all == false.
With the patch, the result is manifold.

[test_remesh_intersection.zip](https://github.com/libigl/libigl/files/1981694/test_remesh_intersection.zip)

qnzhou:
OK, I am able to reproduce the problem.  The problem can be reproduced with just 2 intersecting triangles.  @alecjacobson We did not merge new vertices created due to intersection when they inside the triangle.  The proposed fix looks correct to me.

alecjacobson:
@qnzhou will this affect performance much?

qnzhou:
It won't affect performance for boolean since it is done with `stitch_all` set to true.  The change only affects the performance when user does not wish to stitch coinciding vertices.

YijingLi-Gene:
Thank you!

PR #799, opened 2018-06-05
preliminary volume rendering and minor viewer refactoring

fwilliams:
This is an early stage PR of an implementation of volume rendering in the viewer. 

The architecture of the viewer is built around a set of `ViewerData` (renamed `MeshData`) which represent meshes. The programmer selects which of a list of meshes is current and modifies its state. I also renamed `Viewer::data()` to `Viewer::mesh()`

Similarly, the new and improved viewer has a list of `VolumeData` objects, each of which represents a volume to be rendered. The API for `VolumeData` is basically the same as the one for `MeshData`, you access the current volume by calling `VolumeData& ViewerData::volume();`, you can append and remove volumes as well. 

The OpenGL implementation of volume rendering lives in `VolumeGL` which is called by the high level `VolumeData`.

### DIFF
--- a/include/igl/opengl/MeshData.cpp
+++ b/include/igl/opengl/MeshData.cpp
@@ -6,7 +6,7 @@
 // v. 2.0. If a copy of the MPL was not distributed with this file, You can
 // obtain one at http://mozilla.org/MPL/2.0/.
 
-#include "ViewerData.h"
+#include "MeshData.h"
 
 #include "../per_face_normals.h"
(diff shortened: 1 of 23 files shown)


fwilliams:
Still `TODO`:
* Error Checking
* Transfer Functions
* Expose Rendering Parameters
* Hook up lighting (it's implemented)
* Test scenes with multiple volumes and make sure they look good
* Non janky tutorial
* Fix resizing (this doesn't quite work)

PR #801, opened 2018-06-05
Fix eigen issue

hardikk13:
DynamicSparseMatrix has been marked deprecated by Eigen. Fixing that.

### DIFF
--- a/include/igl/slice_into.cpp
+++ b/include/igl/slice_into.cpp
@@ -33,8 +33,7 @@ IGL_INLINE void igl::slice_into(
   assert(C.maxCoeff() < yn);
 #endif
 
-  // create temporary dynamic sparse matrix
-  //  Eigen::DynamicSparseMatrix<T, Eigen::RowMajor>  dyn_Y(Y);
+  // create temporary matrix.
   Eigen::SparseMatrix<T> tempY(Y);
   std::vector<Eigen::Triplet<double>> triplets;
   // Iterate over outside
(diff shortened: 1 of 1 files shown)


danielepanozzo:
Thanks for the PR. Since this is touching a very important function in libigl I want to do some extra testing before merging it. @alecjacobson we run into troubles with this function during the latest eigen update. Do you still have the checklist that you used for that update? Or will it be enough to do a pass on the tutorials and make sure that they still work?

alecjacobson:
Dynamic sparse matrix has been deprecated for many years now. (See https://github.com/libigl/libigl/pull/154 from 2015)

As long as it's still the fastest way to implement this function, I would like to keep it (despite the warnings). The moment we switch to a version of Eigen that doesn't support it or when we find something faster, I'm happy to get rid of it. (Feel free to reopen with a performance comparison to the contrary).

jdumas:
@alecjacobson do you have a benchmark for testing this? Or just a random matrix with random slices would be enough? Eigen's `setFromTriplets()` is slower because it is performing two passes on the data and a O(n) count-sort (in-place transpose). When inserting triplets in ascending order (without duplicates), it is in fact faster to use `A.reserve(nnz);` where `nnz` is of type `VectorXi`, and for each outer index do `A.insert(i, j) = val;` with `j` in ascending order. I reported the issue [a while ago](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=974) so it should be fixed in 3.3 and later, but I think the fix hasn't been backported to the 3.2 branch.

```cpp
VectorXi nnz;
...
A.reserve(nnz);
for (j=0; j < outer_size; ++j)
	for (i = ...)
		A.insert(i, j) = value;
```

alecjacobson:
No, but feel to create one.

jdumas:
Ok. It's pretty low on my todo list tbh, so not sure when this will happen =)

hardikk13:
We are using igl w/ a version of Eigen that make this an issue for us. However, I've created a temporary fix to the solution by submitting by the change to a fork from here and using that as a submodule. I'm happy to take a stab at the solution suggested by @jdumas if that is feasible to be included in igl. As we'd like to keep using the latest version of the igl.

jdumas:
If you can do some benchmark and try out what I said, it'd be great yeah =)

As a workaround to keep a code that works with both Eigen 3.2, and version 3.3+, we can use the preprocessor macro:

```cpp
#if EIGEN_VERSION_AT_LEAST(3,3,0)
...
#else
...
#endif
```

PR #802, opened 2018-06-06
ply.h function conflicting on windows

teseoch:
moved code of ply.h inside namespace igl::external::ply so it doesnt conflict the name with other libraries like geogram


alecjacobson:
This is probably a good idea. For consistency with our [style](http://libigl.github.io/libigl/style-guidelines/), the namespace should match the directory layout. In this case it should just be `igl::ply`

Issue #711, opened 2018-02-24
Improve CMake modularity


jdumas:
Previous discussion was started [here](https://github.com/libigl/libigl/commit/321343d402a590c855040ed89469025155484981#commitcomment-27514114), I just want to continue it in a dedicated thread.

The idea is to improve the flexibility of the CMake build system, with respect to optional libigl modules. Right now we have a bunch of `option(LIBIGL_WITH_MODULE ...)` that may depend on external libraries that are not shipped with libigl (e.g. CGAL, MOSEK, Matlab, etc.). Then it's not clear if what should happen if we want to enable the module, but one dependency is not satisfied. There are basically two philosophies here:

1. The CMake build should produce an error if an option is turned `ON`, but a dependency is not available. I.e. the option indicates a strong requirement, and it is up to the user to set the option correctly in the first place.
2. The CMake should turn the option `OFF` if a dependency is not satisfied, and the user should test if the target `igl::module` exists before trying to link to it. This indicates an *optional* dependency, and the CMake script will fail anyway if we link to a `igl::module` target that doesn't exist (this is why using an alias target `foo::bar` is better than using a CMake variable`FOO_BAR`).

What I was proposing in the discussion linked above, is to have the behavior be a part of the call to `find_package()`. For example you could imagine writing something like this:

```cmake
find_package(LIBIGL REQUIRED COMPONENTS core opengl glfw OPTIONAL_COMPONENTS mosek)
```

The line above will try to load `igl::core`, `igl::opengl` and `igl::glfw`, producing an error if any of them is not found. It will also try to load `igl::mosek`, but will gracefully ignore the module if, e.g. MOSEK is not available.

By default we could treat all components but `igl::core` as `OPTIONAL_COMPONENTS`. So a call to:
```cmake
find_package(LIBIGL REQUIRED)
```

Would be equivalent to:

```cmake
find_package(LIBIGL REQUIRED COMPONENTS core OPTIONAL_COMPONENTS opengl glfw mosek ...)
```

Before I spend any time looking into this (this shouldn't be a big change to the CMake), I'd like to know if this is something you'd be willing to integrate or not.

jdumas:
Eventually I would also like to have a single `CMakeLists.txt` at the root of the libigl repository that would be used to build both the python bindings and the tutorials in a single shot, thus dividing by 2 the time spent in Travis for the CI.

Then wouldn't be much there, except for the following lines:

```cmake
cmake_minimum_required(VERSION 3.1)
[...]

PR #807, opened 2018-06-11
fixed problem with findmatlab on windows

teseoch:
this line:
find_package(Matlab COMPONENTS MEX_COMPILER MX_LIBRARY ENG_LIBRARY)
fails on windows without any way to disable it.

Now, LIBIGL_WITH_MATLAB is set to on by default.

If LIBIGL_WITH_MATLAB = ON, it searches for matlab and set LIBIGL_WITH_MATLAB to off in case of failure (it also prints a warning).

To avoid the problem, simply set LIBIGL_WITH_MATLAB to OFF

### DIFF
--- a/shared/cmake/libigl.cmake
+++ b/shared/cmake/libigl.cmake
@@ -8,7 +8,7 @@ find_package(Boost 1.48 COMPONENTS thread system) # --> BOOST_FOUND
 if(CGAL_FOUND AND BOOST_FOUND)
   set(CGAL_AND_BOOST_FOUND TRUE)
 endif()
-find_package(Matlab COMPONENTS MEX_COMPILER MX_LIBRARY ENG_LIBRARY) # --> Matlab_FOUND
+
 find_package(MOSEK) # --> MOSEK_FOUND
 find_package(OpenGL) # --> OPENGL_FOUND
 
@@ -20,7 +20,7 @@ option(LIBIGL_WITH_COMISO            "Use CoMiso"         ON)
 option(LIBIGL_WITH_CORK              "Use Cork"           OFF)
 option(LIBIGL_WITH_EMBREE            "Use Embree"         OFF)
 option(LIBIGL_WITH_LIM               "Use LIM"            ON)
-option(LIBIGL_WITH_MATLAB            "Use Matlab"         "${Matlab_FOUND}")
+option(LIBIGL_WITH_MATLAB            "Use Matlab"         ON)
 option(LIBIGL_WITH_MOSEK             "Use MOSEK"          "${MOSEK_FOUND}")
 option(LIBIGL_WITH_OPENGL            "Use OpenGL"         "${OPENGL_FOUND}")
 option(LIBIGL_WITH_OPENGL_GLFW       "Use GLFW"           "${OPENGL_FOUND}")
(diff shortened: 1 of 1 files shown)


jdumas:
How is that an issue specific to Matlab and Windows, and not shared also with the `find_package(MOSEK)` and `find_package(OpenGL)` in the beginning? Setting the libigl options to `OFF` in case a dependency is not found was the behaviour I originally intended when I rewrote the CMake files, but this was changed by @alecjacobson in [321343d402a](https://github.com/libigl/libigl/commit/321343d402a590c855040ed89469025155484981). I tried to suggest a better alternative in #711 but this was never really discussed.

alecjacobson:
`find_package` is not supposed to fail, it's just supposed to set Matlab_FOUND to false. 

The ideal behaviour is:

- by default everything module that can be compiled is compiled,
- meanwhile, any module can be safely turned OFF regardless if it's found or not

The original CMAKE rewrite failed the first criteria, but it sounds like the change I made is failing for the matlab module (at least on windows). We should try to fix this (maybe #711 already does).

The proposed PR is not a sustainable solution though, so closing makes sense now.

jdumas:
Apparently the `find_package()` from CMake produces a fatal error on Windows, although I don't know exactly which one since @teseoch didn't want to tell me.

#711 is not fixing much at the moment since it's mostly me rambling on my own.

jdumas:
Ok I manage to obtain more information on this issue. I'll add it to the FAQ at some point (once we move the website), but here the story in the meantime:

On Windows (at least) and with a recent version of Matlab (2017a), the line [find_package(Matlab)](https://github.com/libigl/libigl/blob/508cb9940f4d1e8e54137d5afe2fd2eb9c4dc672/shared/cmake/libigl.cmake#L11) fails with the following error message:

> CMake Error at C:/Program Files/CMake/share/cmake-3.9/Modules/FindMatlab.cmake:576 (string):
 string sub-command STRIP requires two arguments.

Now, I believe this issue has been fixed by [this commit](https://github.com/Kitware/CMake/commit/a245f850bc4f4f209e759b6eab5b48c0693e38af#diff-b0f7025fb70c823cb0c39809f15668d1) in the CMake repository, so it should work again with a recent version of CMake (at least 3.10).

The recommended solution for now if you run into this issue (i.e. have Matlab > 2017a installed and CMake < 3.10) are:
- Update your version of CMake if possible.
- Otherwise, copy the `FindMatlab.cmake` from the latest CMake release and add it to the `CMAKE_MODULE_PATH` of your project.

Obviously this is not ideal since users that have matlab installed but don't use it with libigl can run into this issue. Looks like it could also happen under Linux/Mac under the right condition, but I haven't confirmed that yet.

ZouaghiHoussem:
Hello,
I'm having the same issue in mac, I have cmake (3.15.0) this is the error showen by the cmake gui console

> CMake Error at /Applications/CMake.app/Contents/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Matlab (missing: Matlab_INCLUDE_DIRS Matlab_MEX_LIBRARY
  Matlab_MEX_EXTENSION Matlab_ROOT_DIR MEX_COMPILER MX_LIBRARY ENG_LIBRARY
  MAT_LIBRARY) (found version "NOTFOUND")
Call Stack (most recent call first):
  /Applications/CMake.app/Contents/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /Users/mac/Documents/CPPWorkSpace/libigl/cmake/FindMatlab.cmake:1513 (find_package_handle_standard_args)
  /Users/mac/Documents/CPPWorkSpace/libigl/cmake/libigl.cmake:300 (find_package)
  cmake/FindLIBIGL.cmake:36 (include)
  CMakeLists.txt:13 (find_package)

PR #810, opened 2018-06-14
Fast Winding Numbers for points

GavinBarill:
Hello Alec,

Here's the pull request. Let me know if you'd like me to make any changes.

Thank you,
Gavin


include/igl/copyleft/cgal/point_areas_and_normals.h
```diff
@@ -0,0 +1,59 @@
...
+    // Inputs:
+    //   P  #P by 3 list of point locations
+    //   I  #P by k list of k-nearest-neighbor indices into P
+    //   O  #P by 3 list of point orientation vectors (optional)
+    // Outputs:
+    //   A  #P list of estimated areas
+    //   N  #P by 3 list of estimated normals
+    template <typename DerivedP, typename DerivedI, typename DerivedO,
+      typename DerivedA, typename DerivedN>
+    IGL_INLINE void point_areas_and_normals(
```
alecjacobson: Let's just call this point_areas and have as an optional output the tangent plane (normals, currently `N`). Also, add a note to the header that this _should_ be implemented by pre-filtering `I`, (and in that case the api would be `point_areas(P,I,A)` and `point_areas(P,I,N,A)`)

include/igl/build_octree.h
```diff
@@ -0,0 +1,64 @@
...
+  //   children       a vector of vectors, where the ith entry is a vector of
+  //                  the ith octree cell's of octree children
+  //   centers        a vector where the ith entry is a 3d row vector
+  //                  representing the position of the ith cell's center
+  //   widths          a vector where the ith entry is the width of the ith
+  //                  octree cell
+  //
+  template <typename DerivedP, typename IndexType, typename CentersType,
+    typename WidthsType>
+  IGL_INLINE void build_octree(const Eigen::MatrixBase<DerivedP>& P,
```
alecjacobson: Change this file/function name to `octree`

If it's quick children, centres, widths should be arrays rather than vectors.

include/igl/knn_octree.h
```diff
@@ -0,0 +1,51 @@
...
+  //                  the ith octree cell's of octree children
+  //   centers        a vector where the ith entry is a 3d row vector
+  //                  representing the position of the ith cell's center
+  //   widths         a vector where the ith entry is the width of the ith
+  //                  octree cell
+  // Outputs:
+  //   I  #P by k list of k-nearest-neighbor indices into P
+  template <typename DerivedP, typename KType, typename IndexType,
+    typename CentersType, typename WidthsType, typename DerivedI>
+  IGL_INLINE void knn_octree(const Eigen::MatrixBase<DerivedP>& P,
```
alecjacobson: Change this to `knn` (and match api to new `octree` api.

include/igl/fast_winding_number.cpp
```diff
@@ -0,0 +1,371 @@
+#include "fast_winding_number.h"
+#include "build_octree.h"
+#include "knn_octree.h"
+#include "parallel_for.h"
+#include "copyleft/cgal/point_areas_and_normals.h"
```
alecjacobson: This is not legal. You can only include things at or above. Separate into two files: one in igl/ that takes area as a input and then one in igl/copyleft/cgal that computes areas inside.

PR #819, opened 2018-06-21
fix: principal_curvature returns vector of bad vertex indices

vskhitkov:
Made igl::principal_curvature return false when it's internal CurvatureCalculator fails


tutorial/CMakeLists.txt
```diff
@@ -32,13 +32,13 @@ else()
 endif()
 
 ### Choose which chapters to compile
-option(TUTORIALS_CHAPTER1 "Compile chapter 1" ON)
```
danielepanozzo: These changes should be removed

tutorial/203_CurvatureDirections/main.cpp
```diff
@@ -39,7 +39,8 @@ int main(int argc, char *argv[])
   // Compute curvature directions via quadric fitting
   MatrixXd PD1,PD2;
   VectorXd PV1,PV2;
-  igl::principal_curvature(V,F,PD1,PD2,PV1,PV2);
```
danielepanozzo: z is unused, can you call the version without z?

PR #827, opened 2018-07-06
Add python wrapper for igl::colormap

maxjiang93:
Adding a python wrapper for igl::colormap for additional colormaps.


fwilliams:

PR #830, opened 2018-07-08
Update submodules

jdumas:
- Update embree to 2.17.4
- Compile stb_image as a static library instead of a shared one


danielepanozzo:
Works for me, @alecjacobson can you merge if you approve?

PR #834, opened 2018-07-12
Update to Embree 3.2.1

ikoruk:
Generated by running `./scripts/cpp-patch.py --patch embree2_to_embree3.patch` as suggested in the upgrade section of https://embree.github.io/api.html and then fixing up the remaining issues.

Seems to work for me but I haven't tested any advanced usage.

### DIFF
--- a/include/igl/embree/EmbreeIntersector.h
+++ b/include/igl/embree/EmbreeIntersector.h
@@ -21,8 +21,8 @@
 #include <Eigen/Core>
 #include <Eigen/Geometry>
 
-#include <embree2/rtcore.h>
-#include <embree2/rtcore_ray.h>
+#include <embree3/rtcore.h>
+#include <embree3/rtcore_ray.h>
 #include <iostream>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


jdumas:
It looks like both errors are related to embree: [#197](https://github.com/embree/embree/pull/197) and [#159](https://github.com/embree/embree/issues/159). Probably will need to wait for the next release to see the fix.

ikoruk:
Embree has updated to the new version but it doesn't appear like they fixed cmake

ikoruk:
Not sure what exactly happened with the gcc build though...

ikoruk:
According to [this](https://github.com/JuliaLang/julia/pull/28857#issuecomment-415860776) the reason the CI is failing is maybe an old version of binutils. It would be quite surprising to me if embree didn't compile on gcc.

jdumas:
Hi. Closing this for now as we have some big changes coming up. Let's revisit that once the aforementioned issues are fixed in embree.

ikoruk:
I'm not sure these are issues with embree but rather issues with the CI. And given that it doesn't make much sense to have igl be the one to compile embree, I'm not sure this should be the limiting factor. I've been able to run with embree 3 locally without any issues.

jdumas:
Thanks a lot! Feel free to update the `.travis.yml` as needed also to compile on Travis! We can probably update the minimum version of gcc from 4.8 to 5 if it helps.

PR #839, opened 2018-07-19
Make libigl installable using CMake

ithron:
Libigl does currently not create a xxxConfig.cmake file that can be used by  `find_package(libigl)`.

Added cmake exports and an install target to install the core library into `${CMAKE_INSTALL_PREFIX}`.
This way other projects can include libigl using `find_package`.

Installation it currently not available for optional modules, but the commands to enable installations are given as comments.


jdumas:
Hi. That's a good idea and it's something I've wanted to do for a while. I have several several concerns regarding the proposed implementation:
- It seems that it only exports header files for now. If we want to install a header-only version of libigl, this needs to copy also the .cpp
- Targets should be exported without the `igl_` prefix (I.e. I want to be able to use `igl::core` after the `find_package()`). Is that the case with this implementation? From what I understand it will import `igl::igl_core`, but I may be mistaken (don't have much experience yet with exported targets).
- Generally I'd shy away from using global variables and glob, as it would be preferable to have CMake retrieve the list of files associated to a given target for export purposes (i.e. avoid the `foreach(HEADERS ${INSTALL_HEADERS})`).
- Also need to make sure it works when libigl is included as a subdirectory, e.g. by using `CMAKE_CURRENT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR.

Ideally, there's a lot more that I'd like to do. E.g. supporting `COMPONENTS` and `OPTIONAL_COMPONENTS` in the `find_package()` (see #711), packaging with CPack, etc. Then of course there's the question of properly exporting targets for third-party dependencies, properly handling version numbers, etc.

Other style issues:
- CMake commands should be lowercase only `LIST` -> `list`
- CMake variables should be prefixed by `LIBIGL_*` to not pollute the global space
- The export name should be `igl-export` or `igl_export`.

jdumas:
Also it's probably not libigl's job to install Eigen's headers (nor any other third-party headers for that matters), but that raises the question on how we should install third-party libraries.

ithron:
> It seems that it only exports header files for now. If we want to install a header-only version of libigl, this needs to copy also the .cpp

If used as a header only library, for what do we need the .cpp files for?
If I understand correctly, the .cpp files are for the static build. So if building with `--DLIBIGL_USE_STATIC_LIBRARY=ON` the static library is build and installed into `lib` while the headers reside in `include`.

> Targets should be exported without the igl_ prefix (I.e. I want to be able to use igl::core after the find_package()). Is that the case with this implementation? From what I understand it will import igl::igl_core, but I may be mistaken (don't have much experience yet with exported targets).

That's right. I'll have a look on how to change the export name.

> Generally I'd shy away from using global variables and glob, as it would be preferable to have CMake retrieve the list of files associated to a given target for export purposes (i.e. avoid the foreach(HEADERS ${INSTALL_HEADERS})).

That'll be nice. Maybe one can the the `PUBLIC_HEADER` property for this purpose. Of course each target then has to explicitly specify its public headers.

> Also need to make sure it works when libigl is included as a subdirectory, e.g. by using CMAKE_CURRENT_SOURCE_DIR instead of `CMAKE_SOURCE_DIR.

Fixed that, also the style issues.

> Then of course there's the question of properly exporting targets for third-party dependencies, properly handling version numbers, etc.

Exporting the third-party dependencies it problematic in the current setup.
For the dependencies included with `find_package` everything is fine, but `add_subdirectory` is a problem. 
One solution would be to use a package manager like [Hunter](https://github.com/ruslo/hunter).
With hunter the `add_subdirectory(external/PKG)` call would be replaced by 
```cmake
hunter_add_package(PKG)
find_package(PKG CONFIG REQUIRED)
```
and everything is fine.

I actually issued this PR in order to be able to build a hunter package for libigl.

jdumas:
The .cpp are needed in header-only mode because they contain the implementation of the templated functions. In header-only mode the .h do a `#include` of the .cpp.

> That'll be nice. Maybe one can the the PUBLIC_HEADER property for this purpose. Of course each target then has to explicitly specify its public headers.

In header-only mode the `INTERFACE_INCLUDE_DIRECTORIES` are set on each target, but the source files are not explicitly listed (since those are `INTERFACE` targets). So probably what I'm advocating is not possible at the moment.

> Exporting the third-party dependencies it problematic in the current setup.
> For the dependencies included with find_package everything is fine, but add_subdirectory is a problem.
> One solution would be to use a package manager like Hunter.

Yeah, I know Hunter. It's nice but there are some drawbacks to it: you get to depend on a whole new ecosystem, you have less control on how you can update each library, etc. There are advantages to compiling everything statically in a single project with `add_subdirectory()` (consistent compilation flags between debug/release, etc.).

ithron:
One way to solve the dependency problem could be this one:
add `find_dependency(CGAL)` (e.g.) calls to the libigl-config.cmake. This way e.g. a `find_package(libigl COMPONENTS CGAL)` will fail if CGAL cannot be found. So the user is responsible for managing the dependencies.
A `make install` would only install libigl and no dependencies. But maybe providing and option for each dependency for automatic installation might be convenient, e.g. add
`option(libigl_INSTALL_<DEPENDENCY>, "Install <DEPENDENCY> dependency", Off)` for each dependency.

If libigl is installed system wide, this might be the only clean way to do it.

If the libigl-config.cmake is used from within a build tree the user must still be able to 'inject' their version of the dependency into libigl. This is currently the case because `find_package` is called before `add_subdirectory`.
In the package configuration file one could then use `find_dependency` for all dependencies where `find_package` succeeded and directly use the dependencies target when `find_package` failed.
Then, of course, the dependencies also need to export their targets.

In summary, there would be two separate package configuration files:
one for installation with `make install`, that uses `find_dependency` for all dependencies and
one for a build package that uses 'local' builds of dependencies that couldn't be found by `find_package`.

jdumas:
Anyway regarding your current PR, I'd be willing to accept it as a first step if:
1. You remove the `list(APPEND ...)` that you've added to the different modules. As discussed this is more complicated to do properly, so let's work on that in a separate PR.
2. You can find a way to export the target as `igl::core` instead of `igl::igl_core`, or let me know if you're stuck so we can try to find a workaround.

And do you know why the Windows build is failing on AppVeyor? I do not see anything obvious in the log.

jdumas:
Thanks! That looks cool. I'll take a closer look probably next week. @alecjacobson @danielepanozzo any particular thoughts on this? In any case, this is bound to change a bit once we deal with other modules/refactor the main CMake accordingly.

> The .cpp files are not installed when in header only mode.
I think you meant the other way around?

ithron:
Ah, yes, of course. The cpp files are installed when in header only mode.

I'm also currently trying to figure out how to sole the Eigen dependency problem.
So when a user tries to use libigl with `find_package` then it might be better to issue an error when no Eigen installation can be found.

martinruenz:
Hey there,
I have a question regarding this merge: To me it appeats like only the `igl::core` and `igl::common` component are exposed using the config.cmake. Is this the case? Is there a way to access other components, such as `igl::opengl_glfw`?

I tried something like this, but without success yet:
```cmake
find_package(LIBIGL REQUIRED QUIET COMPONENTS opengl glfw )
target_link_libraries(${PROJECT_NAME}
  igl::core
  igl::common
  igl::opengl
  igl::opengl_glfw
  Eigen3::Eigen
)
```

jdumas:
Right now only `igl::core` is exported. Other components depend on other libraries, which make it more annoying to export with CMake. If you need more than the basic setup the recommended approach for now is to build libigl with you project (cf. [libigl-example-project](https://github.com/libigl/libigl-example-project)).

PR #840, opened 2018-07-20
Eigen3.3 compatibility

ithron:
Fixed some errors that appear when using Eigen3.3.
Still compatible with Eigen 3.2

### DIFF
--- a/include/igl/sparse_cached.cpp
+++ b/include/igl/sparse_cached.cpp
@@ -117,7 +117,7 @@ IGL_INLINE void igl::sparse_cached(
 
 
 #ifdef IGL_STATIC_LIBRARY
-#if EIGEN_VERSION_AT_LEAST(3,2,9)
+#if EIGEN_VERSION_AT_LEAST(3,3,2)
 // SparseMatrixBase::StorageIndex was introduced in eigen 3.2.9
   template void igl::sparse_cached<double>(std::vector<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::StorageIndex>, std::allocator<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::StorageIndex> > > const&, Eigen::Matrix<int, -1, 1, 0, -1, 1> const&, Eigen::SparseMatrix<double, 0, int>&);
   template void igl::sparse_cached_precompute<double>(std::vector<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::StorageIndex>, std::allocator<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::StorageIndex> > > const&, Eigen::Matrix<int, -1, 1, 0, -1, 1>&, Eigen::SparseMatrix<double, 0, int>&);
--- a/include/igl/sparse_cached.cpp
+++ b/include/igl/sparse_cached.cpp
@@ -117,6 +117,6 @@ IGL_INLINE void igl::sparse_cached(
 
 
 #ifdef IGL_STATIC_LIBRARY
-template void igl::sparse_cached<double>(std::vector<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::Index>, std::allocator<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::Index> > > const&, Eigen::Matrix<int, -1, 1, 0, -1, 1> const&, Eigen::SparseMatrix<double, 0, int>&);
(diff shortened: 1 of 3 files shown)


ithron:
Note that this PR does not touch issue https://github.com/libigl/libigl/pull/801 (DynamicSparseMatrix deprecation). Only *hard* errors were fixed.

jdumas:
Does `SparseMatrix::StorageIndex` already exists in Eigen 3.2? Shouldn't code use `SparseMatrix::Index` with Eigen <= 3.2?

ithron:
It exists but is undocumented.

jdumas:
Do you know which in version it was introduced? Otherwise, we can go with a `#if EIGEN_VERSION_AT_LEAST(3,3,0)` to conditionally compile the right template instantiation.

include/igl/sparse_cached.cpp
```diff
@@ -117,6 +117,12 @@ IGL_INLINE void igl::sparse_cached(
 
 
 #ifdef IGL_STATIC_LIBRARY
-template void igl::sparse_cached<double>(std::vector<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::StorageIndex>, std::allocator<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::StorageIndex> > > const&, Eigen::Matrix<int, -1, 1, 0, -1, 1> const&, Eigen::SparseMatrix<double, 0, int>&);
-template void igl::sparse_cached_precompute<double>(std::vector<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::StorageIndex>, std::allocator<Eigen::Triplet<double, Eigen::SparseMatrix<double, 0, int>::StorageIndex> > > const&, Eigen::Matrix<int, -1, 1, 0, -1, 1>&, Eigen::SparseMatrix<double, 0, int>&);
+#if EIGEN_VERSION_AT_LEAST(3,2,9)
```
jdumas: May be better to start at v3.3, where the type has been documented.

ithron:
It's documented since [3.3.2](http://eigen.tuxfamily.org/index.php?title=ChangeLog#Eigen_3.3.2)

jdumas:
I think it's fine to start at v3.3.0 actually. See [Eigen's changelog](http://eigen.tuxfamily.org/index.php?title=ChangeLog#Eigen_3.2.9) and note about [Index type here](http://eigen.tuxfamily.org/index.php?title=3.3#Index_typedef). The thing is that `SparseMatrix::Index` is defined as an `int` in the 3.2 branch, but not in the 3.3 branch. So code that is using Eigen 3.3.0 or higher should use `StorageIndex` instead. Sorry for the confusion and thanks for your patience :p

include/igl/vertex_triangle_adjacency.cpp
```diff
@@ -70,7 +70,7 @@ IGL_INLINE void igl::vertex_triangle_adjacency(
   // vfd now acts as a counter
   vfd = NI;
 
-  VF = Eigen::VectorXi(3*F.rows());
+  static_cast<DerivedVF &>(VF )= Eigen::VectorXi(3*F.rows());
```
jdumas: I'm sorry can you remind me why is this line necessary? If you need to access the derived type you can write `VF.derived() = `. I remember having issues with templated types and allocation in Eigen 3.3, not sure what's the typical way to go.

ithron:
It's required because operator= of PlainObjectBase is protected in 3.3. However, `VF.derived() = ...`
does the same. Maybe using `VF.derived()` is better safer.

jdumas:
Right. Maybe we should be using `MatrixBase` instead of `PlainObjectBase` instead then? See [this note](http://libigl.github.io/libigl/style-guidelines/#templating-with-eigen) on the libigl website. @alecjacobson I never remember which one should be used as template, and I don't think we have a FAQ entry on that. I think it got discussed in previous issues but I don't remember which ones. Can you refresh my memory on that?

ithron:
Related: https://github.com/libigl/libigl/pull/414

ithron:
From ["The class hierarchy](https://eigen.tuxfamily.org/dox/TopicClassHierarchy.html):

> These classes serve as base classes for the five core classes mentioned above. They are more internal and so less interesting for users of the Eigen library.

- PlainObjectBase
- DenseCoeffsBase
- EigenBase

So I think either MatrixBase (matrix expression), VectorBase (vector expression) or DenseBase (matrix or vector expression) should be preferred over PlainObjectBase.

But maybe that's beyond the scope of this PR?

jdumas:
@danielepanozzo Can we merge this?

PR #841, opened 2018-07-20
Rename `igl::components` to `igl::vertex_components`

fwilliams:
we have the functions `igl::components()` and `igl::facet_components(…)` for computing connected components on vertices and faces respectively.

The naming here is super confusing (and someone wasted a couple of hours because they didn't know that facet_components existed).

This change improves discoverability in 2 ways:
1 - More consistent naming by changint `components()` to `vertex_components()` which hints that there may exist a `facet_component()` function.
2 - Reference `facet_components()` in the docs for `vertex_components()` and vice versa


danielepanozzo:
I agree with the change. @alecjacobson can you merge if you also agree?

jdumas:
Filename should also be changed accordingly btw. And maybe the docstring of each function could reference each other.

fwilliams:
bump

PR #861, opened 2018-08-04
Enable nested viewers.

jdumas:
So far it was possible to spawn new nested viewers, but the application was crashing when the user exits the innermost GLFW viewer. This PR fixes that, removed some of the global variables that were used in `Viewer.cpp`, and properly disable the copy/move constructor of the Viewer class to prevent any surprise.

Right now there is still a bug if we spawn a nested viewer with ImGui menu, but I'm working on fixing that (when #859 will be merged).


jdumas:
Alright I've also fixed the code so that we can have nested viewers with ImGui menus! The code is currently in my [nested-imgui](https://github.com/jdumas/libigl/tree/nested-imgui) branch. I'll open a PR once my currently opened PRs are merged.

alecjacobson:
Needs time to review changes to imgui/cmake.

PR #863, opened 2018-08-05
Fix travis build of python and introduce setup.py

jiangzhongshi:


jdumas:
Looks good to me. @danielepanozzo shall we merge?

PR #875, opened 2018-08-19
[WIP] Fancy point shaders.

jdumas:
Been working on this feature this week-end. I'm still working out some details but I want to open the PR so we can discuss the changes. Note that this PR already contains #874 so it should be merged afterwards. Also referencing #683 for navigation purposes. 

[frame0101](https://my.mixtape.moe/fboxsu.webm)

I want to look at drawing line segment as well, and maybe will refactor the current changes accordingly to have something more homogeneous. For lines we have both the wireframe, and line segment overlay, which are sent differently to the GPU, so we might different codepath. I also need to factor our the GLSL shading function (that right now is duplicated between the point shader and the mesh shader).

Currently point radii can be defined either in world space or in camera space. The former means that points will get bigger as we zoom, the latter means that point will not change size as we zoom. This is illustrated in the video above (click on the picture).

Here is some code to test the new shader:

```cpp
#include <igl/readOFF.h>
#include <igl/bounding_box_diagonal.h>
#include <igl/avg_edge_length.h>
#include <iostream>
#include "tutorial_shared_path.h"

Eigen::MatrixXd V;
Eigen::MatrixXi F;

// This function is called every time a keyboard button is pressed
bool key_down(igl::opengl::glfw::Viewer& viewer, unsigned char key, int modifier)
{
  std::cout<<"Key: "<<key<<" "<<(unsigned int)key<<std::endl;
  if (key == '1')
  {
    // Clear should be called before drawing the mesh
    viewer.data().clear();
    viewer.data().set_mesh(V, F);
    viewer.core.align_camera_center(V,F);
    viewer.core.radius_in_screen_space = false;
    Eigen::VectorXd radius(V.rows());
    radius.setConstant(igl::avg_edge_length(V, F) / 4.0 * viewer.core.camera_base_zoom);
    viewer.data().set_points(V, Eigen::RowVector3d(1, 0, 0), radius);
  }
  else if (key == '2')
  {
    viewer.data().clear();
    viewer.data().set_mesh(V, F);
    viewer.core.align_camera_center(V,F);
    viewer.core.radius_in_screen_space = true;
    Eigen::VectorXd radius(V.rows());
    radius.setConstant(0.05 * viewer.core.camera_base_zoom);
    viewer.data().set_points(V, Eigen::RowVector3d(1, 0, 0), radius);
  }

  return false;
}


{
  // Load two meshes
  igl::read_triangle_mesh(TUTORIAL_SHARED_PATH "/bumpy.off", V, F);
  igl::opengl::glfw::Viewer viewer;
  // Register a keyboard callback that allows to switch
  viewer.callback_key_down = &key_down;
  viewer.data().set_mesh(V, F);
  viewer.launch();
}
```

### DIFF
--- a/include/igl/opengl/MeshGL.cpp
+++ b/include/igl/opengl/MeshGL.cpp
@@ -139,17 +139,15 @@ IGL_INLINE void igl::opengl::MeshGL::draw_mesh(bool solid)
   glPolygonMode(GL_FRONT_AND_BACK, solid ? GL_FILL : GL_LINE);
 
   /* Avoid Z-buffer fighting between filled triangles & wireframe lines */
-  if (solid)
+  if (!solid)
   {
-    glEnable(GL_POLYGON_OFFSET_FILL);
-  }
-  else
-  {
-    glPolygonOffset(1.0, -1.0);
+    glEnable(GL_POLYGON_OFFSET_LINE);
+    glPolygonOffset(-1.0, -1.0);
   }
   glDrawElements(GL_TRIANGLES, 3*F_vbo.rows(), GL_UNSIGNED_INT, 0);
 
-  glDisable(GL_POLYGON_OFFSET_FILL);
+  if (!solid)
+    glDisable(GL_POLYGON_OFFSET_LINE);
   glPolygonMode(GL_FRONT_AND_BACK, GL_FILL);
 }
 
@@ -213,7 +211,7 @@ uniform mat4 proj;
 
 // Lights
 uniform vec4 fixed_color;
-uniform vec3 light_position_eye;
+uniform vec3 light_vector_eye;
 uniform bool orthographic;
 
 const vec3 Ls = vec3 (1, 1, 1);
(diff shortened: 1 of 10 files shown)


fwilliams:
Looks great! 

With regards to this item: "Maybe make a default shader without 3D shading, but add this as a user-plugin.", I'd rather have the option to disable 3d shading for points built directly into the viewer. Maybe `viewer.core.enable_3d_point_shading = false;` as opposed to a user plugin.

jdumas:
I'd say this depends on performances. The shader for cylinders is a bit more complicated and emit 4 triangles per lines instead of 2 (for the spheres). I think a 2D line with flat endpoints would require a single quad instead of 2 (for cylinders). My shader is also not really optimized at the moment, so we need to test on low-end graphics cards what is the performance drop for larger models before we decide if we want to make it the default or not.

2018-08-21-143555_1280x800_scrot

jdumas:
Finally managed to debug my shader for cones, so now we can make nice arrows likes this (click on the image to open the video):

[2018-08-21-223726_1280x800_scrot](https://my.mixtape.moe/skjbvb.mp4)

Still pretty much WIP, as I need to cleanup the shaders and refactor the API around it (to pass radius to lines), etc. Also need to add a utility `arrow.cpp` to create arrows from lines (i.e. splitting a line segment in 4 points with their respective radii).

Also notice how I've now separated the shader codes in separate files. These in fact C++ code, which are `#include`d in `MeshGL.cpp`, but it makes development easier (syntax highlighting, easier navigation ...).

PR #908, opened 2018-09-28
Added optional weights to Dijkstra's algorithm + some template definitions

zero-impact:
Dijkstra's can now be called with a vertex-weight parameter. Should be backwards-compatible.

Also added template definitions for static build.


include/igl/dijkstra.h
```diff
@@ -33,10 +34,31 @@ namespace igl {
   IGL_INLINE int dijkstra_compute_paths(const IndexType &source,
```
alecjacobson: `dijkstra_compute_paths` --> `dijkstra`

PR #922, opened 2018-10-14
Multiple viewports

jdumas:
Continuation of #748


jdumas:
Postponed for now. Weird behaviour when moving the mouse over the other viewport + window size broken on mac.

jdumas:
We need to reopen it later after the new history has been pushed.

PR #947, opened 2018-10-16
update for embree 3

ikoruk:
Generated by running `./scripts/cpp-patch.py --patch embree2_to_embree3.patch` as suggested in the upgrade section of https://embree.github.io/api.html and then fixing up the remaining issues.

Update of PR #834 .


alecjacobson:
This is a great opportunity to add some embree unit tests (e.g., as a separate PR that we can merge before this one).

ikoruk:
OK it seems both gcc7 and gcc8 in these linux systems have this issue. I was able to compile on gcc7 on mac without an issue. I don't have a linux machine handy so I'm not sure what's happening.

jdumas:
Looks like an issue with avx512 instructions not being available or something? Maybe there's a way to disable those on embree? Maybe @fwilliams or I can try to build this on a Linux and see if it goes through.

fwilliams:
So I am able to compile locally with no issues at all. My guess is the Travis VMs don't have AVX instructions and so things fail. 

A quick look at the CMakeLists.txt file (https://github.com/embree/embree/blob/master/CMakeLists.txt#L231) for embree reveals there are options to disable it. Can you try disabling all vectorized instructions for *only* travis builds?

jdumas:
Maybe passing  `-DEMBREE_ISA_AVX=OFF -DEMBREE_ISA_AVX2=OFF` on `.travis.yml` should do the trick?

jdumas:
Ok I've restarted the Travis job hoping it will pass this time. @alecjacobson do you want us to have unit tests before merging this one into `dev`? @ikoruk since you're using this maybe you can help us set up some unit tests for the functions you modified?

jdumas:
For reference if it timeouts again I'll need to take a closer look at Travis's cache system and see if tools like [mtime_cache](https://github.com/iboB/mtime_cache) can help on top of ccache.

jdumas:
Ah crap it still hits the 50min mark. @ikoruk can you push a commit that disables CGAL and other libraries beside embree to trigger a build of embree 3 on Travis, so it can be cached, and then push another commit that re-enables them (the cached build of embree should be faster). Sorry this is really annoying.

ikoruk:
@jdumas it didn’t seem to work maybe mtime_cache would help.

fwilliams:
What are we deciding to do with this? On the one hand, I'm fairly certain it works, on the other, building embree everytime is going to timeout or CI. @jdumas is build caching not working?

jdumas:
It is working to some extend (fetching stuff from the cache still takes a non-zero amount of time). For example it seems the VM for macOS builds faster than the Linux one. I want to take a look at this `mtime_cache` option that seems to preserve modification times and avoid calling `gcc` entirely if a file hasn't been modified across builds (provided flags are the same). Or we might decide to ditch gcc 4.8 if it's to slow (but gcc 7 is only marginally faster so it's not a real solution).

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams As requested in #952, I added a commit to template EmbreeIntersector. Tell me if I should template RowVector3f on the raytrace calls as well, I'm not sure what the best option is there, it's definitely not as easy as there is a lot of RowVector3f's in intersectBeam for example.

include/igl/embree/EmbreeIntersector.h
```diff
@@ -74,9 +72,10 @@ namespace igl
       //   isStatic  scene is optimized for static geometry
       // Side effects:
       //   The first time this is ever called the embree engine is initialized.
+      template <typename DerivedV, typename DerivedF>
       inline void init(
-        const std::vector<const PointMatrixType*>& V,
-        const std::vector<const FaceMatrixType*>& F,
+        const std::vector<const Eigen::MatrixBase<DerivedV>*>& V,
+        const std::vector<const Eigen::MatrixBase<DerivedF>*>& F,
```
fwilliams: So `DerivedV` and `DerivedF` should always have shape `(3, 1)` (regardless of type). It's possible that the user passes in the wrong type and gets difficult-to-understand compiler errors. Let's add some `static_asserts` on the size of `DerivedV` and `DefrivedF`. Maybe also check that `DerivedF::Scalar` is an integral type (I don't think we should support anything with fewer bits than an `int32`).

include/igl/embree/EmbreeIntersector.h
```diff
@@ -287,43 +287,49 @@ inline void igl::embree::EmbreeIntersector::init(
...
-    geomID = rtcNewTriangleMesh(scene,RTC_GEOMETRY_STATIC,F[g]->rows(),V[g]->rows(),1);
+    RTCGeometry geom_0 = rtcNewGeometry (g_device, RTC_GEOMETRY_TYPE_TRIANGLE);
+    rtcSetGeometryBuildQuality(geom_0,buildQuality);
+    rtcSetGeometryTimeStepCount(geom_0,1);
+    geomID = rtcAttachGeometry(scene,geom_0);
+    rtcReleaseGeometry(geom_0);
 
     // fill vertex buffer
-    vertices = (Vertex*)rtcMapBuffer(scene,geomID,RTC_VERTEX_BUFFER);
+    vertices = (Vertex*)rtcSetNewGeometryBuffer(geom_0,RTC_BUFFER_TYPE_VERTEX,0,RTC_FORMAT_FLOAT3,4*sizeof(float),V[g]->rows());
```
fwilliams: This looks like a possible bug if the user passes in doubles. Change the `RTC_FORMAT_FLOAT3` to be something generic and `sizeof(float)` to be `sizeof(DerivedV::Scalar)`. 

Also let's static assert that `DerivedV::Scalar` is one of `std::float_t` or `std::double_t`.

include/igl/embree/EmbreeIntersector.h
```diff
@@ -287,43 +287,49 @@ inline void igl::embree::EmbreeIntersector::init(
...
       vertices[i].x = (float)V[g]->coeff(i,0);
       vertices[i].y = (float)V[g]->coeff(i,1);
       vertices[i].z = (float)V[g]->coeff(i,2);
     }
-    rtcUnmapBuffer(scene,geomID,RTC_VERTEX_BUFFER);
+    
 
     // fill triangle buffer
-    triangles = (Triangle*) rtcMapBuffer(scene,geomID,RTC_INDEX_BUFFER);
+    triangles = (Triangle*) rtcSetNewGeometryBuffer(geom_0,RTC_BUFFER_TYPE_INDEX,0,RTC_FORMAT_UINT3,3*sizeof(int),F[g]->rows());
```
fwilliams: Same comment as above.

include/igl/embree/EmbreeIntersector.h
```diff
@@ -564,21 +594,23 @@ ::intersectSegment(const Eigen::RowVector3f& a, const Eigen::RowVector3f& ab, Hi
 
 inline void
 igl::embree::EmbreeIntersector
-::createRay(RTCRay& ray, const Eigen::RowVector3f& origin, const Eigen::RowVector3f& direction, float tnear, float tfar, int mask) const
+::createRay(RTCRayHit& ray, const Eigen::RowVector3f& origin, const Eigen::RowVector3f& direction, float tnear, float tfar, int mask) const
```
fwilliams: We probably want to template `RowVector3f` to `MatrixBase<DerivedDirection>`.

ikoruk:
@fwilliams I can do this here, but I feel like this should be part of a separate PR as it involves significant changes, no?

include/igl/embree/EmbreeIntersector.h
ikoruk: Right now, it always casts to float in this buffer anyway. Embree does not support doubles https://github.com/embree/embree/issues/45, so the question is do we make this to the user (by having them .cast<float>()) or whether it's implicitly cast inside like in this commit.

jdumas:
I'm fine with a separate PR once we merge this.

ikoruk:
Success! Turning AVX off completely seemed to drastically reduce compile times.

fwilliams:
I'm glad the AVX stuff worked out. It'll be nice to get this PR landed. 

Feel free to make templating this a separate PR. If you do end up making that second PR, then make sure to `static_assert` that the scalar input types are always floats (we still want to be general to handle row and column major matrices as input).

jdumas:
Nice! Maybe we should even disable AVX by default if the CMake support for it isn't working for everyone (is there anyone else having issues with it, e.g. on old macbooks?). Let's merge this PR after we merge the unit tests (right now merging is blocked for stupid github reasons, but it should be fixed soon).

PR #952, opened 2018-10-17
add unit tests for embree

ikoruk:
Unit test for embree, in preparation of #947


qnzhou:
Thanks @ikoruk , just FYI, we are in the process of migrating to Catch2 instead of google unit test.

alecjacobson:
This looks good. Is there a special way I'm suppose to mark this as reviewed?

alecjacobson:
I agree @qnzhou , but that looks like a separate PR to me.

fwilliams:
@qnzhou We're going to need to fix that for the Python bindings. Is there an open issue?

ikoruk:
@fwilliams OK, I added a commit to #947. Tell me if I should template RowVector3f as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams OK, I added a commit to #947. Tell me if I should template RowVector3f as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams OK, I added a commit to #947. Tell me if I should template RowVector3f as well, I'm not sure what the best option is there.

ikoruk:
@fwilliams OK, I added a commit to #947. Tell me if I should template RowVector3f as well, I'm not sure what the best option is there.

ikoruk:
I discovered a bug in intersectBeam and issued a fix into this PR yesterday, so this should be re-reviewed before merging.

teseoch:
there is a script (posted on the other PR) to convert a google test into Catch2 test

teseoch:
I updated the test to catch2. @ikoruk can you check?

jdumas:
Alright, I managed to compile embree statically on Windows, the build on AppVeyor looks ok now!

As for why unit-tests were not showing on AppVeyor before, this has to do with the way Catch2 discovers unit tests and populates the CTest config. Indeed, the macro `catch_discover_tests()` will run internally `./igl_tests` to get the list of available unit tests. Now, because of the new CMake setup, the executable `igl_tests.exe` was not finding the `embree3.dll` at runtime, and thus was not able to create the list of tests for CTest. Which is why `igl_tests.exe` was compiled correctly but no tests were found on AppVeyor. `catch_discover_tests()` interprets the [return code](https://github.com/catchorg/Catch2/blob/03d122a35c3f5c398c43095a87bc82ed44642516/contrib/CatchAddTests.cmake#L34) of the unit test program as the number of tests to be defined (and then proceeds to parse the stdout). I don't have the energy to test this, but I suspect that a missing dll produces a return code > 0, which is why no error was reported by AppVeyor and no test was run.

PR #954, opened 2018-10-18
Make libigl targets installable only for toplevel projects by default…

jdumas:
Follow-up on #839.

I got an error in a project using a different target for Eigen, due to how CMake targets are exported:
```
CMake Error: install(EXPORT "igl-export" ...) includes target "igl_common" which requires target "eigen" that is not in the export set.
```

I propose to disable export by default in `libigl.cmake`, and enable it for the toplevel project (like tests, tutorials and python bindings).

- [x] This is a minor change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ endif()
 option(LIBIGL_BUILD_TESTS      "Build libigl unit test"        ${LIBIGL_TOPLEVEL_PROJECT})
 option(LIBIGL_BUILD_TUTORIALS  "Build libigl tutorial"         ${LIBIGL_TOPLEVEL_PROJECT})
 option(LIBIGL_BUILD_PYTHON     "Build libigl python bindings"  ${LIBIGL_TOPLEVEL_PROJECT})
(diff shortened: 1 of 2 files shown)


jdumas:
Updated the commit to also fix a warning with recent CMake version about FindOpenGL:

```
CMake Warning (dev) at /usr/share/cmake-3.12/Modules/FindOpenGL.cmake:270 (message):
  Policy CMP0072 is not set: FindOpenGL prefers GLVND by default when
  available.  Run "cmake --help-policy CMP0072" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  FindOpenGL found both a legacy GL library:

    OPENGL_gl_LIBRARY: /usr/lib/libGL.so

  and GLVND libraries for OpenGL and GLX:

    OPENGL_opengl_LIBRARY: /usr/lib/libOpenGL.so
    OPENGL_glx_LIBRARY: /usr/lib/libGLX.so

  OpenGL_GL_PREFERENCE has not been set to "GLVND" or "LEGACY", so for
  compatibility with CMake 3.10 and below the legacy GL library will be used.
```

alecjacobson:
Do I understand correct that this does not change the behaviour of `make`? But it will enable a `make install`?

PR #955, opened 2018-10-18
Dev delaunay tests

alecjacobson:
This adds unit tests for the Delaunay triangulation function and exposes the `is_delaunay` function, with an overload that computes "Delaunay-ness" per (half-)edge. Along the way, I fixed some PlainObjectBase to MatrixBase and added necessary templates.

- [x] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Looks like you started from a clone with the old history, so it's pulling all the duplicated commits that were cleaned by BFG -- really annoying that github doesn't prevent you from creating the PR in this case.

jdumas:
Ok I've force pushed the tip of `dev` onto your branch `dev-delaunay-tests` to remove the old history and avoid weighting down a fresh `git clone` of libigl again, then re-applied your changes (hope I haven't forgotten anything).

alecjacobson:
Thanks for doing that. In the meantime, I just copied the changes on a clean PR (https://github.com/libigl/libigl/pull/956). So we can just leave this one closed.

jdumas:
No worries, looks like we were just minutes apart =)

PR #956, opened 2018-10-18
delaunay unit tests, expose is_delaunay, templates, PlainObject -> Matrix

alecjacobson:
This adds unit tests for the Delaunay triangulation function and exposes the `is_delaunay` function, with an overload that computes "Delaunay-ness" per (half-)edge. Along the way, I fixed some PlainObjectBase to MatrixBase and added necessary templates.

- [x] Adds new .cpp file.
- [ ] This is a minor change.


alecjacobson:
Trying again (https://github.com/libigl/libigl/pull/955#issuecomment-431197258). I think this worked correctly.

tests/include/igl/copyleft/cgal/order_facets_around_edges.cpp
```diff
@@ -127,23 +121,23 @@ TEST(copyleft_cgal_order_facets_around_edges, DuplicatedFaces) {
...
+//    Eigen::MatrixXi F(6, 3);
+//    F << 0, 1, 2,
+//         1, 2, 0,
+//         2, 1, 3,
+//         1, 3, 2,
+//         1, 2, 4,
+//         4, 1, 2;
+//
+//    assert_order(V, F, 1, 2, {1, 0, 2, 3, 5, 4});
+//}
```
qnzhou: This seems to be undoing the bug fix from https://github.com/libigl/libigl/pull/944.  Did you pull dev?

include/igl/is_delaunay.cpp
```diff
@@ -0,0 +1,112 @@
...
+  D.setConstant(F.rows(),F.cols(),false);
+  const auto D_at = [&D,&num_faces](const int he)->typename DerivedD::Scalar&
+  {
+    const int f = he%num_faces;
+    const int c = he/num_faces;
+    return D(f,c);
+  };
+  typedef typename DerivedV::Scalar Scalar;
+  // Should use Shewchuk's predicates instead.
+  const auto float_incircle = [](
```
qnzhou: This is an inexact predicate?  Shewchuk's exact predicates is not easily templated, which does not fit the header-only convention.  What about Geogram's predicates?

jdumas: Geogram's predicate are also not header-only (but should be easier to isolate, see the [predicate psm](http://alice.loria.fr/software/geogram/doc/html/geogram_PSM.html)). But why not treat it as an external dependency that should be compiled as a pair of .cpp/.h? (Like we do for `stb_image` for example.)

tests/include/igl/copyleft/cgal/order_facets_around_edges.cpp
alecjacobson: Fixed.

jdumas:
Shall we merge #961 before merging any new test-related PR? Just so that have a common framework to work with.

jdumas:
Ah well I guess not then :D

PR #960, opened 2018-10-19
enabled testing on windows

teseoch:
Fixed compilation of test on windows.

I disable Comiso because the tests fails because of boost dll and bbw test on windows because it takes too long

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


tests/include/igl/bbw.cpp
```diff
@@ -8,6 +8,9 @@
 
 TEST(bbw, decimated_knight)
 {
+  #ifdef WIN32
+    return;
```
jdumas: I'm pretty sure this is not the right fix :p

teseoch: this thing was never finishing.
The tests where never run on windows.
My guess it that it comes from the parallel loop

CMakeLists.txt
```diff
@@ -32,10 +32,41 @@ option(LIBIGL_WITH_XML               "Use XML"                      ON)
 option(LIBIGL_WITH_PYTHON            "Use Python"                   ${LIBIGL_BUILD_PYTHON})
 ### End
 
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})		
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})		
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 
+if(MSVC)
```
jdumas: You can make the code shorter using `CMAKE_CONFIGURATION_TYPES`, e.g. (to be adapted):
```
foreach(config ${CMAKE_CONFIGURATION_TYPES})
			string(TOUPPER ${config} config)
			string(REPLACE /MD /MT CMAKE_C_FLAGS_${config} "${CMAKE_C_FLAGS_${config}}")
			string(REPLACE /MD /MT CMAKE_CXX_FLAGS_${config} "${CMAKE_CXX_FLAGS_${config}}")
		endforeach()
```

I would also move this into a different file `cmake/LibiglWindowsFix.cmake` or something.

tests/include/igl/bbw.cpp
teseoch: this is where it gets stuck:
`parallel_for(m,optimize_weight,2);`

danielepanozzo: Can you try to disable (with an assert) the parallelization inside parallel_for on windows? I had troubles with C++11 threads on VS before.

teseoch: you mean an ifdef and skip the parallel for?

teseoch:
I disabled the parallel_for inside the bbw.cpp in libigl.
Now the test takes around 3 minutes to finish.

teseoch:
I dont have access to a windows machine to test the changes.
@jdumas if you do, please fix the cmake as you please.
I think this PR urgent since it will enable *running* code with CI on windows.

jdumas:
I can give it a shot this week-end. But you don't need a Windows machine to move the cmake boilerplate to a different file :p

PR #975, opened 2018-10-23
Avoid possible incompatible type casting

copyme:
clock_t can be implemented as long int and in the code int was used instead to capture values returned by clock(). 

- [-] Adds new .cpp file.
- [-] Adds corresponding unit test.
- [-] Adds corresponding python binding.
- [X] This is a minor change.


copyme:
The other question is if this code should really be there. It looks like a side effect of the early development stage.

jdumas:
@jiangzhongshi do you want to review this or we can merge? (You "assigned" yourself as a reviewer)

PR #977, opened 2018-10-24
Bump version of glfw.

jdumas:
Current version of glfw produces the following warning with recent versions of CMake:

```
CMake Deprecation Warning at external/glfw/CMakeLists.txt:10 (cmake_policy):
  The OLD behavior for policy CMP0042 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
```

The new version has CMake 3.0 as minimum version instead of 2.8.12, but that's ok since our CMake requires 3.1.

- [x] This is a minor change.

### DIFF
--- a/cmake/LibiglDownloadExternal.cmake
+++ b/cmake/LibiglDownloadExternal.cmake
@@ -77,7 +77,7 @@ endfunction()
 function(igl_download_glfw)
 	igl_download_project(glfw
 		GIT_REPOSITORY https://github.com/glfw/glfw.git
-		GIT_TAG        58cc4b2c5c2c9a245e09451437dd6f5af4d60c84
+		GIT_TAG        5afcd0981bf2fe9b9550f24ba298857aac6c35c2
 	)
 endfunction()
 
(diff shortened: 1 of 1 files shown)


fwilliams:
This is totally orthogonal, but does this fix the transparency issue on Mac that @teseoch was working on?

jdumas:
No this the issue is still open in GLFW (see [here](https://github.com/glfw/glfw/issues/1334) and [there](https://github.com/glfw/glfw/pull/1346). What teseo pushed is a less hacky fix that doesn't involve messing up the transparency or moving the window.

jdumas:
Argh of course it fails to build on Windows =( (but glfw has CI on Windows, so we may be missing something).

jdumas:
Updated again so the compilation bug on Windows is fixed. Anybody on macOS can confirm there's no trouble with the viewer so we can merge? (@teseoch with the multiple dpi?).

PR #978, opened 2018-10-24
Multiple viewports

jdumas:
Follow-up on #922 and #748, initial code by @Roipo

I removed the post_resize() code for multiple views, since it had a division by zero and crashed on my machine (@jiangzhongshi and @hankstag). I think it's best to leave it to the user how to update the viewports.

There is still this weird behavior with tutorial 108 where rotating one view, and dragging the mouse to the other side will suddenly start rotating the second view. If anyone wants to take a shot at this, this is the last bug to fix before we can integrate this.

- [x] Adds new .cpp file.
- [x] ~This is a minor change~.


include/igl/opengl/glfw/Viewer.cpp
```diff
@@ -123,16 +123,16 @@ namespace opengl
 namespace glfw
 {
 
-  IGL_INLINE int Viewer::launch(bool resizable,bool fullscreen)
+  IGL_INLINE int Viewer::launch(bool resizable /*= true*/,bool fullscreen /*= false*/, int windowWidth /*= 1280*/, int windowHeight /*= 800*/)
   {
     // TODO return values are being ignored...
-    launch_init(resizable,fullscreen);
+    launch_init(resizable,fullscreen,windowWidth,windowHeight);
```
teseoch: add a parameter for the window name

include/igl/opengl/glfw/Viewer.cpp
```diff
@@ -123,16 +123,16 @@ namespace opengl
...
     launch_rendering(true);
     launch_shut();
     return EXIT_SUCCESS;
   }
 
-  IGL_INLINE int  Viewer::launch_init(bool resizable,bool fullscreen)
+  IGL_INLINE int  Viewer::launch_init(bool resizable,bool fullscreen, int windowWidth, int windowHeight)
   {
```
teseoch: propagated here

teseoch: I am getting a strange behaviour when dragging outside for rotating and for moving

jdumas: Dragging on another viewport is still buggy, I suspect might the same issue when dragging outside.

teseoch: the dragging outside keeps the last view port selected
for the moving, i feel that is a problem of mouse offset, when you move to the new viewport it keeps the delta and the second object "jumps"

teseoch:
I played around with `is_visible` and it doesnt hide labels on that layer.
It *does* hide the points and edges

My code looks like:
```
viewer.append_mesh();
viewer.data(1).add_label(V.row(0), "label");
viewer.data(1).is_visible
```

jdumas:
Yeah good point. That's because the labels are drawn by the ImGui menu plugin. I need to update it to handle the visibility option, and also make it easier to draw arbitrary labels with your own `ViewerData` (that is not in the viewer's `data_list`).

teseoch:
> Yeah good point. That's because the labels are drawn by the ImGui menu plugin. I need to update it to handle the visibility option, and also make it easier to draw arbitrary labels with your own `ViewerData` (that is not in the viewer's `data_list`).

The imgui labels are problematic, there is no alternative to show text in opengl?
here there is a decent list of libs to do it: https://www.opengl.org/archives/resources/features/fontsurvey/
among others there is glut...

jdumas:
Lol we're not going back to glut. We'd have to pull another dependency to draw text anyway, so I don't see the issue with using ImGui, apart from the lack of depth (but it was like that before). I mean there is a bug with too many labels, but there are ways to fix it.

teseoch:
> Lol we're not going back to glut. We'd have to pull another dependency to draw text anyway, so I don't see the issue with using ImGui, apart from the lack of depth (but it was like that before). I mean there is a bug with too many labels, but there are ways to fix it.

There is another bug if you open 2 windows...

jdumas:
But it's not related to labels in particular (we need to fix it also for general menus). It was supposedly fixed too, but apparently not on macOS...

include/igl/opengl/glfw/Viewer.cpp
teseoch: no window name?

teseoch:
When pressing `t` for toggling wireframe both viewport get updated.

I dont really know which is the expected/correct behaviour

jdumas:
The reason that pressing `t` toggle the visualization option on both viewports is that those options are stored [per-mesh](https://github.com/jdumas/libigl/blob/8eac86c1633955268b886ff38a35001fc4a78762/include/igl/opengl/ViewerData.h#L189) as booleans. I can turn them into `unsigned int` and use bitmask operations to provide more control over how each mesh is displayed on each viewport. I think this change wouldn't be too intrusive (single viewport can still use `true` and `false` that will be implicitly cast to the correct integer value).

But we need to make the decision. @alecjacobson @danielepanozzo would you be ok with this change?

danielepanozzo:
Yes, I think it is a good solution and it makes sense to have more fine grained control.

jdumas:
Ok done. There are a couple of options that are still shared across viewports:
- `face_based` and `invert_normals`, since they require updating the mesh VBO on the GPU. We could use a uniform flag for the latter though, but I don't think it's worth the trouble.
- `show_vertid` and `show_faceid`, since showing these is currently handled by the ImGui plugin (if enabled), and right now tihs will only draw labels on the active viewport. The code for the labels could use a refactor anyway, so I'll leave that for later.

teseoch:
I dont know if it is an expected behaviour but the "Toggle between models" acts on both viewports

jdumas:
Yes, the `selected_data_index` is an attribute of the viewer, not individual viewports. Putting it into the viewports would complexify the API for the basic viewer. It's something that the user can implement themselves if they want selection per viewport.

alecjacobson:
Unless https://github.com/libigl/libigl/issues/1133 is easy to fix, this will need to be rolled back out of dev.

PR #984, opened 2018-10-25
Update cut_to_disk.cpp

evouga:
Fixed a bug in my implementation (did not correctly update all data structures when reversing the direction of traversal of a cut when hitting a dead end).

- [ ] Adds new .cpp file.
- [x] This is a minor change.


fwilliams:
This looks good. Is there a unit test which triggered the bug you fixed? If it's easy, could you add something?

evouga:
Where do I put the .obj files for the unit tests?

Here is an example where the old code produces wrong cuts (the mesh is in a .zip file because GitHub does not allow uploading .obj for some reason).

[annulus.zip](https://github.com/libigl/libigl/files/2544655/annulus.zip)

fwilliams:
@evouga any chance you could add a test so we can merge this? It'd be nice to have your bugfix in the code :)

fwilliams:
Ah, it wasn't cross referenced from here so I didn't catch it. What's the plan then? If we merge the test PR it should fail without the bugfix. If we merge this first, then we have untested code. Can we merge the two PRs into one? @qnzhou ?

jdumas:
It was cross-refed, just above the message you posted :D

fwilliams:
Oh then it appears I'm just blind :) Questions above still hold though.

evouga:
@fwilliams You're asking me? The current code is broken. Whether that's better or worse than untested is up to y'all...

I haven't had time to look into how to install and configure the other repository, and link it to the main repository's test cases. I've posted the example .obj and it looks like qnzhou is taking care of it, though.

qnzhou:
@fwilliams I have a unit test on it, but we probably need more.  This function is not well tested.  I also think it may over-cut a mesh (i.e. annulus only need one cut to convert into disk, but the results shows more than one cut).  @evouga, please correct me if my understanding is wrong.

However, PR https://github.com/libigl/libigl/pull/1007 is currently blocked because I am still missing permission to push to the [libigl-test-data repo](https://github.com/libigl/libigl-tests-data).  If you have push permission, can you just upload the [annulus model](https://github.com/libigl/libigl/files/2544655/annulus.zip) to the repo?

evouga:
@qnzhou There should not be any extra cuts for the annulus: just the two boundary cycles, and one cut connecting them.

For meshes of higher genus however there are indeed multiple ways of cutting to disk with different number of cuts. Maybe newer algorithms can guarantee optimality of the number of cuts, but the paper this code implements does not. I believe that's fully documented in the .h, though.

fwilliams:
@qnzhou Looks like I don't have write access either. @danielepanozzo @alecjacobson could you please give myself and @qnzhou access to the [libigl-tests-data](https://github.com/libigl/libigl-tests-data) repo?

PR #988, opened 2018-10-29
Heat Geodesics and Intrinsic Delaunay Triangulation

alecjacobson:
Fixed up a lot of style mistakes and PlainObjectBase -> MatrixBase prototypes along the way.

Added Heat Geodesics tutorial entry (716).

- [x] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Lol, when I saw the number of commit I was thinking "wow, is this using the old history again", but no, it's just your regular crazy PR touching 73 files :p 

My only comment regarding the new tutorial so far is: can we do the without the custom shader? Maybe by loading a texture onto the mesh? We could add a function `texture_from_colormap` (if we don't have one already) like this:

```
void texture_from_colormap(const Eigen::MatrixXd &rgb, GLuint & id)
{
  int width = 1;
  int height = (int) rgb.rows();

  Eigen::Matrix<unsigned char, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> cmap;
  if (rgb.maxCoeff() > 1.0) {
    cmap = rgb.cast<unsigned char>();
  } else {
    cmap = (rgb.array() * 255.f).cast<unsigned char>();
  }
  assert(cmap.cols() == 3);
  cmap.conservativeResize(cmap.rows(), 4);
  cmap.col(3).setConstant(255);

  if (id == 0) {
    glGenTextures(1, &id);
  }
  glBindTexture(GL_TEXTURE_2D, id);
  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR_MIPMAP_LINEAR);
  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
  glTexImage2D(
    GL_TEXTURE_2D, 0, GL_RGBA,
    width, height, 0, GL_RGBA, GL_UNSIGNED_BYTE, cmap.data());
  glGenerateMipmap(GL_TEXTURE_2D);
  glBindTexture(GL_TEXTURE_2D, 0);
}
```

alecjacobson:
I like this idea. We should do it as a separate PR and change _all_ of the tutorials that use scalar field visualization. We should also try to expose it nicely in the viewer with something like `set_per_vertex_data`.

jdumas:
Which tutorials require scalar field visualization? I don't think there are that many. We could name the function `ViewerData::set_scalars` to be coherent with `set_normals`/`set_colors`/`set_uv`. It would simply set the `v` texture coordinate per vertex, with a `u=0`? I mean we could also reuse `set_uv` if the matrix has only 1 column.

There are two questions to answer I think:
- What about normalization? I think it should be left to the user to normalize the scalar field, and values are simply clamped to the range [0, 1].
- We can also add support for periodic colormap. [Twilight](https://github.com/bastibe/twilight) is a perceptually uniform colormap introduced in the latest version of matplotlib. Could be useful too :p

alecjacobson:
From the tutorial there are quite a few:

- 202_GaussianCurvature
- 203_CurvatureDirections
- 204_Gradient
- 206_GeodesicDistance
- 302_Sort
- 303_LaplaceEquation
- 304_LinearEqualityConstraints
- 305_QuadraticProgramming
- 306_EigenDecomposition
- 403_BoundedBiharmonicWeights
- 702_WindingNumber
- 704_SignedDistance
- 712_DataSmoothing
- 716_HeatGeodesics

I agree about the normalization, though it's easy enough to provide parameters and overloads to control that.

Periodic colormap makes sense to add, too.

PR #990, opened 2018-10-30
fix wrong reference in cotmatrix_intrinsic

ikoruk:
Noticed a bug while looking at the new dev branch, `cotmatrix_intrinsic.h` references the wrong cpp so it gives errors in compile in header-only.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/cotmatrix_intrinsic.h
+++ b/include/igl/cotmatrix_intrinsic.h
@@ -34,7 +34,7 @@ namespace igl
 }
 
 #ifndef IGL_STATIC_LIBRARY
-#  include "cotmatrix.cpp"
+#  include "cotmatrix_intrinsic.cpp"
 #endif
 
 #endif


alecjacobson:
yikes! Glad we caught that.

What about adding something like this to one of the travis builds?

```
THROW=0;find . -type f -iname "*.h" -print0 | { while IFS= read -r -d $'\0' h; do if grep "# *include *\".*\.cpp\"" "$h" | grep -vq "# *include *\"$(basename -- "$h" .h).cpp\"" ; then echo "Error: bad .cpp found in $h" ; THROW=1 ;fi; done && exit $THROW; }
```

It spits out error messages and sets an error status if it finds any YYY.cpp files included in XXX.h, where YYY != XXX.

jdumas:
I don't see any harm in that.

PR #968, opened 2018-10-22
fix: segments_intersect bug.

xarthurx:
also test if t is larger than 1.

[Describe your changes and what you've already done to test it.]

#### Check all that apply (change to `[x]`)
- [x] All changes meet [libigl style-guidelines](https://libigl.github.io/style-guidelines/).
- [ ] Adds new .cpp file.
- [ ] Adds corresponding unit test.
- [ ] Adds corresponding python binding.
- [x] This is a minor change.

PR #994, opened 2018-11-03
fix segments_intersection bug.

xarthurx:
fix segments_intersection bug, with #968 #957  as unit tests.

- [x] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
Interesting. It seems the test gives different numerical results on Windows and Linux/macOS? How could that be?

xarthurx:
I have no idea... I don't have a Linux machine at hand.
So I'll close the PR on master branch, will that ok?

jdumas:
Yes keep this one open. We'll try it on linux/macOS whenever we have some time :D

teseoch:
It wasn't a numerical problem, it was an initialization problem.

In fact the test wasn't testing anything. @xarthurx I moved the lines `a_t = t;` and `a_u = u;` before the if so the method always returns something. Can you confirm that's correct?

xarthurx:
@teseoch 
I don't think this modification is necessary.
the two params only make sense if the return value is true, and for using this function, you should always test if the return value is true first.

(Even you move it, you should move the assignment above the first condition line.)

teseoch:
Well in the test you are using the two return values which are never assigned. So you are testing just the defaul initialisation of vars...

xarthurx:
sorry, not sure what you’re talking about exactly.

Did you test it with your coord? It should return a False in your case
because it didn’t intersect. In this case, what value u and t contains
doesn’t make any difference. You shouldn’t use them to decide if the two
segments intersect or not.

Or did you look at the test unit?

jdumas:
@xarthurx the problem is that in the unit test you wrote, you first test that the function returns `false` (line `REQUIRE(sect1 == false);`. That's ok. But then right after you also test that `REQUIRE(intersectCondition1 == false);`. But to set the boolean `intersectCondition1` you are using the output variable `a1` and `a2`, which may have undefined values if you return early. For example it's possible that the variables were initialized with `0` on macOS, which is why it worked on Travis. But it may be initialized with random values on Linux, and thus the test would fail.

Teseo's fix is correct because even when there is no intersection, the output values of `a1` and `a2` are still correctly set. One may argue this is not needed, but it may be useful to test on which "side" of the segments the lines intersect, even when the segments do not. And also removes such uninitialized behavior.


Now, please fix your pull-request by changing your indentation to 2-space to match with the [Style Guidelines](https://libigl.github.io/style-guidelines/) so we can merge this!

xarthurx:
@jdumas  OK. Then the first return should also be put downwards. I fix this, as well as the spaceing issue.

include/igl/segment_segment_intersect.cpp
```diff
@@ -11,55 +11,55 @@
...
+  // t = (q - p) x s / (r x s)
 
-    // (r x s) ~ 0 --> directions are parallel, they will never cross
-    Eigen::RowVector3d rxs = r.cross(s);
-    if (rxs.norm() <= eps)
-        return false;
+  // (r x s) ~ 0 --> directions are parallel, they will never cross
+  Eigen::RowVector3d rxs = r.cross(s);
+  if (rxs.norm() <= eps)
+	return false;
```
jdumas: Thanks. This guy is still a tab though!

xarthurx: Oops, sorry.
Fixed.

PR #995, opened 2018-11-04
Fix Eigen compile error when using setup.py

phg1024:
Adding igl and Eigen include path hints so `setup.py` can find igl and Eigen headers, otherwise compiling the python bindings using `setup.py` would fail. I've tested the changes on ubuntu+python2.7, MacOS+python2.7, and MacOS+conda+python3.5.

Using `setup.py` seems to be necessary on MacOS, since the `pyigl.so` file generated by cmake build from project root is not usable due to import error:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: dlopen(./pyigl.so, 2): Symbol not found: __ZN3igl42internal_angles_using_squared_edge_lengthsIN5Eigen6MatrixIdLin1ELin1ELi0ELin1ELin1EEES3_EEvRKNS1_10MatrixBaseIT_EERNS1_15PlainObjectBaseIT0_EE
  Referenced from: ./pyigl.so
  Expected in: flat namespace
 in ./pyigl.so
```
Building the python bindings from `python` directory using cmake also works with this fix.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -13,6 +13,11 @@ set(Python_ADDITIONAL_VERSIONS 3.4 3.5 3.6)
 find_package(PythonInterp REQUIRED)
 find_package(PythonLibs REQUIRED)
 
+# Force setup.py to find igl headers and Eigen under external
+set(EIGEN3_INCLUDE_DIR "${CMAKE_CURRENT_LIST_DIR}/../external/eigen")
+include_directories(${EIGEN3_INCLUDE_DIR})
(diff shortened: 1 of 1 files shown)


phg1024:
So far this is the most straightforward fix, though it is definitely a hacky solution. I tried to fix it the other way but it seems to require a lot more effort. As you said, new python bindings will be released soon so it may not be very useful to fix this issue anymore. I'll just leave it as is so anyone else seeing similar issue can refer to this.

jdumas:
Yeah, let's close this then, I'd rather not merge hacky solutions in the main repo =)

Issue #963, opened 2018-10-21
Terminals in Wire Mesh


twodee:
The new wire mesh feature is great. I've been looking at implementing this for a long time, and then I saw Alec's blog post about it. Thank you!

My understanding of the algorithm is that the regular polygons must be pushed away from the endpoints a bit to ensure a reasonable surface. This offset works well for generating wire meshes of solid models.

For networks that have a lot of terminal endpoints, this offset isn't so agreeable. We end up with lots of pointy spears. See https://twodee.org/forothers/pointed.png.

Would it be possible to eliminate this offset when a vertex is connected to the network by only one edge? So that the endcaps are flat? If there's interest but not time, I'm happy to look into implementing it myself and submitting a pull request.

jdumas:
A quick workaround would be to place spheres at the endpoints, and use libigl's boolean to "resolve" the self-intersections (this is already what the code is doing if you pass the option `solid = true`). I personnaly don't really use this feature so I can't comment on usability.

alecjacobson:
@twodee could you describe what you'd like? Flat tops? round tops?

unioning spheres would work but would be quite messy compared to the wire meshes we currently make. If you want round caps it might be better to come up with a procedurally way so it perfectly matches the existing vertex from the edge-parts.

alecjacobson:
yeah, that would do it. To find the vertices with only a single edge, I'd create an array of length #Vertices, initialize it to zero and then iterate over all of the edges, incrementing the counter of each vertex that is touched.

PR #999, opened 2018-11-07
Flatten wiremesh caps when vertex on single edge.

twodee:
In #963, I asked about flattening the caps on any terminal vertices of wiremeshes. Following @alecjacobson's implementation suggestions, I zero out the offset when a vertex is incident on only one edge.

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/include/igl/copyleft/cgal/wire_mesh.cpp
+++ b/include/igl/copyleft/cgal/wire_mesh.cpp
@@ -65,11 +65,6 @@ IGL_INLINE void igl::copyleft::cgal::wire_mesh(
 
   // Count each vertex's indicident edges.
   std::vector<int> nedges(WV.rows());
-  for(int v = 0;v<WV.rows();v++)
-  {
-    nedges[v] = 0;
-  }
-
   for(int e = 0;e<WE.rows();e++)
   {
     ++nedges[WE(e, 0)];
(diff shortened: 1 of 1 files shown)


include/igl/copyleft/cgal/wire_mesh.cpp
```diff
@@ -62,6 +62,20 @@ IGL_INLINE void igl::copyleft::cgal::wire_mesh(
     v = v-c*(PV.rows());
     p = v;
   };
+
+  // Count each vertex's indicident edges.
+  std::vector<int> nedges(WV.rows());
+  for(int v = 0;v<WV.rows();v++)
```
fwilliams: This loop is not necessary. The constructor for `std::vector` with a size parameter will initialize all the values to their default (0 for int). Take a look at constructor (2) in https://en.cppreference.com/w/cpp/container/vector/vector.

twodee: I didn't realize std::vector did this. Thanks!

jdumas: Personally I prefer to put `0` for the second argument to be more explicit (it doesn't hurt), but that's just my preference.

twodee: I've added the 0.

twodee:
Fixes #963.

PR #996, opened 2018-11-04
Enable CGAL on Travis + AppVeyor.

jdumas:
It just occurred to me that since CGAL is now disabled by default, we need to explicitly enable it on Travis and AppVeyor...

#### Check all that apply (change to `[x]`)
- [x] All changes meet [libigl style-guidelines](https://libigl.github.io/style-guidelines/).
- [x] This is a minor change.

PR #1001, opened 2018-11-08
Add extra exceptions for arithmetic and types operations 

copyme:
This PR allows for extra exceptions in the debug mode (Linux only).

So, without these few lines we have all the test passing:
```

Running tests...
Test project /home/kacper/Projects/libigl/cmake-build-debug
    Start 1: run_igl_tetgen_tests
1/3 Test #1: run_igl_tetgen_tests .............   Passed    0.01 sec
    Start 2: run_igl_comiso_tests
2/3 Test #2: run_igl_comiso_tests .............   Passed   23.02 sec
    Start 3: run_igl_tests
3/3 Test #3: run_igl_tests ....................   Passed   26.59 sec

100% tests passed, 0 tests failed out of 3

Total Test time (real) =  49.61 sec
```

But if we run the tests with the additional exceptions we get:

```
Running tests...
Test project /home/kacper/Projects/libigl/cmake-build-debug
    Start 1: run_igl_tetgen_tests
1/3 Test #1: run_igl_tetgen_tests .............   Passed    0.01 sec
    Start 2: run_igl_comiso_tests
2/3 Test #2: run_igl_comiso_tests .............   Passed   22.68 sec
    Start 3: run_igl_tests
3/3 Test #3: run_igl_tests ....................***Exception: Numerical  0.46 sec

67% tests passed, 1 tests failed out of 3

Total Test time (real) =  23.16 sec

The following tests FAILED:
          3 - run_igl_tests (NUMERICAL)
Errors while running CTest
make: *** [Makefile:119: test] Error 8
```

Sometimes, it is not that easy to find bugs related to arithmetic and type errors so it will take me a while to find them and fix. But as you can see there is something wrong. So, stay tuned.


copyme:
The first problem comes somewhere in tests/include/bbw.cpp. But, it not so clear so far what is going on. You have there multi-thread operations which do not make things easier to debug. I will have a look at this during the weekend.

jdumas:
Hi. It's a good idea. But before this can be merged it needs to be rebased onto the `dev` branch, where we've switched to Catch2 for unit testing. If you want to help you may also want to take a look at #996, which enables also CGAL for compilation on Travis and AppVeyor. Unfortunately in #996 it seems some tests are randomly failing on macOS due to memory issues (at least one buffer overflow in Eigen).

copyme:
@jdumas OK. I will have a look on these things.

copyme:
@jdumas just a quick question. Is libigl supposed to compile with CGAL on system without static boost (only dynamic)? OpenSUSE does not ship static boost and I can compile CGAL or libdgtal + cgal without problem but igl+CGAL even with set( Boost_USE_STATIC_LIBS OFF) does not compile - I have a cmake message saying that some boost components were not found.

jdumas:
We do set the option `CGAL_Boost_USE_STATIC_LIBS` to `ON` by default (because we try to compile everything statically to avoid dll nightmare on Windows). Try to turn it off maybe.

copyme:
@jdumas There is something strange in cmake files when `CGAL_Boost_USE_STATIC_LIBS` is set to OFF then boot libraries are not searched. I had to set them by hand. It works with other libraries and cgal itself.

PR #1007, opened 2018-11-12
Add cut to disk unittest

qnzhou:
Added a failing unit test for `igl::cut_to_disk` based on https://github.com/libigl/libigl/pull/984.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/tests/include/igl/cut_to_disk.cpp
+++ b/tests/include/igl/cut_to_disk.cpp
@@ -130,3 +130,17 @@ TEST_CASE("cut_to_disk: cube", "[igl]")
 
   cut_to_disk_helper::assert_is_disk(V, F, cuts);
 }
+
+TEST_CASE("cut_to_disk: annulus", "[igl][PR984][bug]") {
+  using namespace igl;
+  Eigen::MatrixXd V;
+  Eigen::MatrixXi F;
+  test_common::load_mesh("annulus.obj", V, F);
+
+  std::vector<std::vector<int>> cuts;
+  cut_to_disk(F, cuts);
+  REQUIRE (cuts.size() == 1);
+
+  cut_to_disk_helper::assert_is_disk(V, F, cuts);
+}
+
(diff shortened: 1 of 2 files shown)


jdumas:
Hmm don't you need to update the `libigl-test-data` repo and change the `CMakeLists.txt` accordingly? I would also tag the PR as `#984` and remove the `bug` tag (once we merge the fix, it'll no longer be a bug :p).

qnzhou:
@jdumas I need permission to push to `libigl-test-data`.

jdumas:
Probably needs to be fixed, but in the meantime you can always create a PR and update the url/sha1 in the cmake file of libigl.

qnzhou:
The problem is that I cannot push my branch to create a PR:

```
git push --set-upstream origin cut-to-disk-unittest
ERROR: Permission to libigl/libigl-tests-data.git denied to qnzhou.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Maybe I need to fork this repo?

jdumas:
Yes if you don't have write permissions you need to fork it =)

qnzhou:
@jdumas This PR is ready to merge.  Can you review it? Thanks!

tests/include/igl/cut_to_disk.cpp
```diff
@@ -130,3 +130,16 @@ TEST_CASE("cut_to_disk: cube", "[igl]")
 
   cut_to_disk_helper::assert_is_disk(V, F, cuts);
 }
+
+TEST_CASE("cut_to_disk: annulus", "[igl][PR984][bug]") {
```
jdumas: I would remove the tags `[PR984]` and `[bug]`. It's no longer a bug once it's fixed, and I would rather put the url to the PR/github issue in a comment above.

qnzhou: Done, however, we will lose the ability of running just this one test like the following:
```
$ ./build/libigl_tests [pr984]
```

jdumas: You can run individual tests using their name `cut_to_disk: annulus`. If you want to keep the tag, maybe write it as `#984` (more github-like) rather than `PR984`. But that seems pretty minor to me.

PR #1008, opened 2018-11-12
Fix build python bindings

jdumas:
I know we have new python bindings in the work, but right now the ones we have are not usable anymore.
- If compiled with `LIBIGL_USE_STATIC_LIBRARY=ON`, there will be missing symbols at runtime.
- I also fixed the path of the tutorial data in `shared.py`.
- Added an error message if someone tries to run cmake from the `python/` folder without the correct options.
- Once this is merged, I'll put a big warning on the website that people should compile with `LIBIGL_USE_STATIC_LIBRARY=ON` if they want to use the python bindings.

Testing and integrating the new python bindings will take some time (a couple of weeks at least), so this needs to patched in the meantime.

- [x] This is a minor change.

### DIFF
--- a/cmake/libigl.cmake
+++ b/cmake/libigl.cmake
@@ -88,11 +88,6 @@ if(BUILD_SHARED_LIBS)
   set_target_properties(igl_common PROPERTIES INTERFACE_POSITION_INDEPENDENT_CODE ON)
 endif()
 
-if(UNIX)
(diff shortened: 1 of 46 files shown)


jdumas:
Yes, it builds now! I've had to update a couple of third-party libraries (tetgen, triangle, stb_image, glad and comiso), but I've removed the hack around the `-fPIC` flag (it is now properly propagated without messing with the CXX_FLAGS globally).

jdumas:
Blergh, who uses python 2 anymore nowadays ...

Issue #1025, opened 2018-11-21
randperm.cpp fails to compile in c++17


ycjungSubhuman:
When compiling static library with -std:c++17 flag, the following error occurs.
```
The'random_shuffle': is not a member of 'std'	C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CMakeLists.txt	D:\Projects\fext\lib\libigl\include\igl\randperm.cpp	20	
```
std::random_shuffle is deprecated in c++14, and completely removed in c++17. I think it is a good idea to replace it with std::shuffle

This is the part where it is problematic

```
include/igl/randperm.cpp

template <typename DerivedI>
IGL_INLINE void igl::randperm(
  const int n,
  Eigen::PlainObjectBase<DerivedI> & I)
{
  Eigen::VectorXi II;
  igl::colon(0,1,n-1,II);
  I = II;
  std::random_shuffle(I.data(),I.data()+n);
}
```

#### Check all that apply (change to `[x]`)
- [x] Windows
- [ ] Mac OS X
- [ ] Linux

See <https://libigl.github.io/CONTRIBUTING/#bugreport> for more tips.

PR #1026, opened 2018-11-21
Use std::shuffle instead of std::random_shuffle in igl/randperm.cpp

ycjungSubhuman:
This commit fixes issue #1025

- [ ] Adds new .cpp file.
- [x] This is a minor change.


ycjungSubhuman:
@jdumas That totally makes sense. Embedding RNG inside the function can be malicious...

However, If we modify the signature of this function, it might harm the usability of this library. Also, it is hard to specialize templates for static libraries.

I have an idea.

1. Use cstdlib std::rand to implement c-style shuffling
1. This function does not call srand inside
1. The user can change the behavior of RNG by setting the global seed(srand). This shall be documented in the header file.

I see igl/randperm.h or igl/randquaternion.h using rand() calls... so it might not harm consistency.

What do you think?

jdumas:
I'd prefer to add an extra parameter that is defaulted to a global random if nothing is provided by the user.

ycjungSubhuman:
@jdumas Yes, I think it is a good idea.

Then I'll try writing
```
  template <typename DerivedI>
  IGL_INLINE void randperm(
    const int n,
Eigen::PlainObjectBase<DerivedI> & I, std::optional<unsigned int> seed=std::nulopt);
``
When I have free time.

**Comment by jdumas** [comment]
Unfortunately, `std::optional` is C++17, so we cannot put that in libigl. An alternative is wrapping things up in a `std::function<int64_t()>` or something.

File: include/igl/randperm.h [hunk 1]
```diff
@@ -9,18 +9,23 @@
...
+  //        initialized with random seed generated by std::rand()
   // Outputs:
   //   I  n list of rand permutation of 0:n-1
-  template <typename DerivedI>
+  template <typename DerivedI, typename URBG=std::minstd_rand>
   IGL_INLINE void randperm(
     const int n,
-    Eigen::PlainObjectBase<DerivedI> & I);
+    Eigen::PlainObjectBase<DerivedI> & I,
+    URBG urbg=std::minstd_rand(std::rand()));
```
jdumas: Does this default initialization work when `URBG` is a different type ? Otherwise, could we write `URBG urbg = URBG(std::srand())` ? Another option is to use a function overload to define the optional argument, which is maybe more in line with libigl's style.

tests/include/igl/randperm.cpp
```diff
@@ -0,0 +1,70 @@
+#include <test_common.h>
+#include <igl/randperm.h>
+#include <random>
+
+TEST(randperm, default_rng_reproduce_identity)
```
jdumas: Please take a look at [other tests](https://github.com/libigl/libigl/blob/dev/tests/include/igl/cotmatrix.cpp) in the `dev` branch, since we've switched to Catch2 for the unit tests (this is why it doesn't compile right now).

include/igl/randperm.h
ycjungSubhuman: Thank you for the suggestion. I think using function overload is a better choice.

tests/include/igl/randperm.cpp
ycjungSubhuman: I forgot to merge to code to dev branch in my machine... Thank you for letting me know!

jdumas:
Wow, I'm not sure what's happening exactly here, but I think github is getting confused by the merge you did. Could you rebase your change onto the tip of the current `dev` branch? (Either closing this PR, or force-push onto your branch to eliminate the weird history duplication).

ycjungSubhuman:
Alright, this is weird.. I'm closing this PR and making a new one.

PR #1029, opened 2018-11-23
Eigen Alignment issue

alecjacobson:
https://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html

A memory alignment assertion fired once for the .data_list in the viewer. I'm not sure if we also need the special stl vector allocator for the data list of if this is enough to fix it. It's hard to get this assertion to fire in a reproducible way.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/opengl/ViewerData.h
+++ b/include/igl/opengl/ViewerData.h
@@ -19,6 +19,14 @@
 // Alec: This is a mesh class containing a variety of data types (normals,
 // overlays, material colors, etc.)
 //
+// WARNING: Eigen data members (such as Eigen::Vector4f) should explicitly
+// disable alignment (e.g. use `Eigen::Matrix<float, 4, 1, Eigen::DontAlign>`),
+// in order to avoid alignment issues further down the line (esp. if the
+// structure are stored in a std::vector).
+//
+// See this thread for a more detailed discussion:
+// https://github.com/libigl/libigl/pull/1029
+//
 namespace igl
 {
 
@@ -103,7 +111,7 @@ public:
   //   C  #E|1 by 3 color(s)
   IGL_INLINE void set_edges (const Eigen::MatrixXd& P, const Eigen::MatrixXi& E, const Eigen::MatrixXd& C);
   // Alec: This is very confusing. Why does add_edges have a different API from
-  // set_edges? 
+  // set_edges?
   IGL_INLINE void add_edges (const Eigen::MatrixXd& P1, const Eigen::MatrixXd& P2, const Eigen::MatrixXd& C);
   IGL_INLINE void add_label (const Eigen::VectorXd& P,  const std::string& str);
 
@@ -189,7 +197,7 @@ public:
   // Point size / line width
   float point_size;
   float line_width;
-  Eigen::Vector4f line_color;
+  Eigen::Matrix<float, 4, 1, Eigen::DontAlign> line_color;
 
   // Shape material
   float shininess;
@@ -205,9 +213,6 @@ public:
     const igl::opengl::ViewerData& data,
     const bool invert_normals,
     igl::opengl::MeshGL& meshgl);
-  
-public:
-  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 };
 
 } // namespace opengl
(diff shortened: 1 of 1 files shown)


jdumas:
Honestly I would rather disable alignment for this specific case `Eigen::Matrix<float,4,1,Eigen::DontAlign> v`, since I don't imagine it's performance-critical...

alecjacobson:
I'm worried later on somebody will add another Vector4f field and forget
this...

jdumas:
Fair enough, but I'm not versed enough in Eigen's arcane to know if we need to mess also with the `std::vector` allocator. If we need to specialize it then it's also gonna be easy for the users to forget when they write their own `std::vector<ViewerData>` in their apps.

alecjacobson:
We can be really conservative and do everything.

We should get an official answer from Eigen whether an stl container of
classes containing fixed-size vectorizable Eigen types needs this special
allocator... I'll post something

alecjacobson:
Actually from the first sentence of https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html it seems pretty clear that we _do_ need the special allocator for `std::vector<igl::ViewerData , ...`

jdumas:
Yeah then I vote for really disabling alignment. I may use `std::vector<igl::ViewerData>` in my code (of stuff unrelated to the `data_list` of the viewer), and it would be extremely annoying to have to pay attention to alignment.

alecjacobson:
OK. But let's put a big warning at the top and bottom of the class that any
new members should not be aligned(-able).

jdumas:
It's a warning that should apply on all classes, not only this one right? And in any case, I don't see that new members will be added very often to this class, but a warning wouldn't hurt.

alecjacobson:
The eigen issue applies to all classes, but we're choosing a particular
solution (to not use alignment, rather than to do the macros/std
allocators). For other classes, it might make more sense to use those
options. So the warning here is specific to this class and the decision
we're making.

>

bubnikv:
Speaking of memory alingment issues: We at Prusa Research started to use libigl for bits and pieces on the Slic3r Prusa Edition project (slicer for FDM / SLA fabrication), thank you!

We stumbled over memory alignment issues reported by the Visual Studio 2013, 32bit builds, for example:
igl\AABB.h(73) : error C2719: 'other': formal parameter with __declspec(align('16')) won't be aligned

This is what Microsoft has to say to it:
https://msdn.microsoft.com/en-us/library/373ak2y1.aspx?f=255&MSPPError=-2147217396

I interpret the error as violating the Eigen recommendation to never pass memory aligned structures by value to a function.
https://eigen.tuxfamily.org/dox/group__TopicPassingByValue.html

I understand that the libigl authors do not guarantee 32bit compatibility, but here (likely by a lucky accident) the Visual Studio 2013 32bit compiler gives us a hint where we may have alignment issues.

For our code to work, I had to fix the copy constructor of AABB to accept a refernce
-      AABB& operator=(AABB other)
+      AABB& operator=(const AABB &other)
       {
-        swap(*this,other);
+        this->deinit();
+        m_left  = other.m_left ? new AABB(*other.m_left) : NULL;
+        m_right = other.m_right ? new AABB(*other.m_right) : NULL;
+        m_box   = other.m_box;
+        m_primitive = other.m_primitive;
         return *this;
       }
       AABB(AABB&& other):

and ClosestBaryPtPointTriangle lambda in igl::point_simplex_squared_distance() to accept references as well:
  const auto & ClosestBaryPtPointTriangle = 
    [&Dot](const Point &p, const Point &a, const Point &b, const Point &c, BaryPoint& bary_out )->Point

jdumas:
Hmm, I don't think the current implementation of AABB trees in libigl is very efficient, since it seems to be using pointers. You might want to check out [geogram](http://alice.loria.fr/software/geogram/doc/html/index.html) which has a very efficient implementation of AABB trees.

That being said if you have fixes you want to contribute to the current class, we'd be happy to evaluate PRs. I don't remember exactly what kind of issue there was with 32bit compilation on Windows, but it was probably alignment-related as well. If 64bits work for you (and passed valgrind/addressSanitizer checks), then you might as well do nothing.

PR #1038, opened 2018-12-03
[WIP] Multiple windows

jdumas:
This is a follow-up of PR #861 that we skipped during the hackaton.

This enables multiple viewers, either nested (the main loop of the new one blocking the rendering loop of the previous ones), or concurrent (rendering of the new viewer interleaves with rendering of the parent viewer). This PR gets rid of all global variables in the `Viewer` class, and adds a new tutorial to showcase both behaviors. I think it's a nice feature that allows, like #978, to have multiple viewports (e.g. useful for debugging a texture along with the 3D mesh).

The only remaining issue is that it doesn't work yet with the imgui menu. It's not like we are introducing a bug since multiple windows were not possible before, but this is something I'm looking at fixing in the future.

- [x] Adds new tutorial.


jdumas:
Ok, I believe I've fixed the integration with ImGui when using multiple windows! If someone on macOS can try the new tutorial `109_MultipleWindows_bin` on their system, that'd be nice!

Now there are only two statics variables in the viewer and the `ImGuiMenu` plugin. I believe we could even get rid of the static used in the `ImGuiMenu` class, but this works well enough as it is.

jdumas:
Ah no I spoke too soon. There are still some issues with multiple viewers wrt the capture of keyboard inputs. But it doesn't crash at least! I'll try to update the PR with a corrected version.

teseoch:
image

problems with hidpi windows opening

jdumas:
For the first issue, how is this happening exactly? Does it have to do with your multi-DPI monitor setup? If you stick to a single monitor, does it work as expected?

The resizing issue is normal I think. The application itself is not redrawn/refreshed by OpenGL, so what you see is just macOS resizing the last image that was displayed (on Linux whatever is drawn on top is retained on the window).

For the shortcuts can you detail a bit what is the problem? What is the expected behavior, and what is actually happening? On Linux if I close a child window it seems to behave as expected (i.e. only the child window gets closed). It may be a macOS specific thing where quitting sets the `glfwWindowShouldClose` flag to all open glfw windows. In any case, this looks more like an enhancement that an actual bug, and would need someone with macOS to be able to implement it properly.

teseoch:
1st issue: I just open the second viewer on an high-dpi display

2st issue: I agree,

3rd issue: on macos you have 2 ways of closing: 1. close the window without closing the app (cmd + w) (on windows it is ctrl F4 i think) and 2. quite the app (cmd + q). What is missing on the multi-window is a "shortcut" to close the newly spanned windows.

jdumas:
Ok. I'll see if I can get my hands on a mac to try this out. Unless someone else in the lab volunteers to debug this :D

alecjacobson:
this is old but still a cool feature. Reading above it seems like the remaining issue was imgui which as changed a bit since. If this is salvageable it would be great to freshen up and merge.

alecjacobson:

PR #1051, opened 2018-12-21
Building a CGAL polyhedron via a float-type Eigen matrix.

metorm:
The origin version only accepts Eigen::MatrixXd as vertices.

Add a template parameter to `igl::copyleft::cgal::mesh_to_polyhedron` so it also accepts `Eigen::MatrixXf` as `V`

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/include/igl/copyleft/cgal/mesh_to_polyhedron.cpp
+++ b/include/igl/copyleft/cgal/mesh_to_polyhedron.cpp
@@ -10,9 +10,9 @@
 #include <CGAL/Polyhedron_incremental_builder_3.h>
 
 
-template <typename Polyhedron>
+template <typename Polyhedron, typename DerivedV>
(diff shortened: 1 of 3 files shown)


jdumas:
While you're at it, you might as well add a template parameter to the second argument `F` ...

metorm:
@jdumas What else types can `F` be?

jdumas:
Well signed, unsigned, long, short ... All matrix types in libigl are templated by default, seems like an oversight this one was not.

metorm:
But, in order to support these int types, what shall I do with the STATIC lib? Add 2*4 or even more explicit entries?

jdumas:
No you don't need to create explicit template instantiations for each type, just add the template parameter to the function signature.

metorm:
@jdumas I had updated the PR.

PR #1057, opened 2018-12-24
Bug fix: mesh_to_polyhedron compiles

alecjacobson:
But I can't test unit tests since all the CGAL tests are broken (re: https://github.com/libigl/libigl/pull/996)

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


alecjacobson:
This should fix the first compilation error identified here: https://travis-ci.org/libigl/libigl/jobs/471976535

jdumas:
I thought the build still had compilation errors?

alecjacobson:
My understanding is that there are multiple problems with cgal on the dev branch. This is fixing one of them...

jdumas:
Hmm. The only problem left after #996 was that some tests (at runtime) were failing randomly. But at least it should compile on all platforms ... Nothwistanding compilation errors introduced since then (e.g. #1051).

alecjacobson:
Seems like there are other issues, on ubuntu I'm getting this error:

[ 99%] *Linking CXX shared module ../../python/pyigl.so*

/usr/bin/ld:
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libboost_thread.a(thread.o):
relocation R_X86_64_32 against `.rodata.str1.8' can not be used when making
a shared object; recompile with -fPIC

/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libboost_thread.a:
error adding symbols: Bad value

collect2: error: ld returned 1 exit status

python/CMakeFiles/pyigl.dir/build.make:425: recipe for target
'../python/pyigl.so' failed

make[2]: *** [../python/pyigl.so] Error 1

CMakeFiles/Makefile2:6326: recipe for target
'python/CMakeFiles/pyigl.dir/all' failed

make[1]: *** [python/CMakeFiles/pyigl.dir/all] Error 2

Makefile:149: recipe for target 'all' failed

make: *** [all] Error 2

alecjacobson:
Maybe this https://github.com/libigl/libigl/pull/1008/files change to cmake/libigl.cmake would fix that?

alecjacobson:
do we have instructions for dealing with that somewhere?

jdumas:
Man, I wished I had instructions for dealing with boost :D
If you just need to get it running on your machine, you can try to force usage of shared boost libraries (if the pb is boost not being compiled with -fPIC in your packages). If the pb is on the libigl side, I'd try to sprinkle `-fPIC` flags all over the place, although in theory we already specify a CMake flag for that.

alecjacobson:
This is on a rather fresh ubuntu machine so I imagine it will pop up as a problem for others (not sure why the situation is different on the travis machine). I'll try to track down any missing fPIC flags.

alecjacobson:
huh. So removing boost_thread from the linking command resolves this issue. Is there a clean cmake-ese way of doing that?

jdumas:
Isn't CGAL using boost_thread?

alecjacobson:
apparently not, or at least it will link OK without it.

jdumas:
Yes, but we're also using CGAL in header-only mode, so we might just not be using anything that requires boost thread. The [CGAL documentation](https://doc.cgal.org/latest/Manual/installation.html#title21) explains why boost thread is needed, and I don't think removing it altogether is a very long-term solution.

alecjacobson:
ok, alternatively it seems to also work if boost_thread is listed first in the linked libraries in the call to the linker. Is there a cmake way of forcing that to happen?

jdumas:
Waat? That seems rather brittle and not super inspiring confidence. I know that for `target_include_directories()` you can specify `BEFORE` to prepend the include flag instead of appending at the end, but there's no such construct for linking libraries afaik.

Which branch are you trying to compile exactly? `dev-fix-cgal`? with both `-DLIBIGL_WITH_PYTHON=ON -DLIBIGL_WITH_CGAL=ON`? Do you need to have both enabled at the same time?

alecjacobson:
Yes, `dev-fix-cgal` via `mkdir build; cd build; cmake -DLIBIGL_WITH_CGAL=ON
..; make`. So, `LIBIGL_WITH_PYTHON` is on by default.

>

jdumas:
Ok I've pushed some compilation fixes to the `dev-fix-cgal` branch. I confirm that I'm able to run the python CGAL boolean tutorial on my machine. On Ubuntu, can you try to pass the option [-DCGAL_Boost_USE_STATIC_LIBS=OFF](https://github.com/libigl/libigl/blob/e699d257fe860d4fd82709835e72438bd0cd1533/cmake/libigl.cmake#L198) to the CMake? We're telling CGAL to compile against the static boost libs by default, but the python bindings build a dynamic lib. This poses a problem if the static boost libs are not compiled with `-fPIC`, as I suspect is the case on Ubuntu but not on Archlinux.

PR #1062, opened 2018-12-27
Use std::shuffle instead of std::random_shuffle in igl/randperm

ycjungSubhuman:
This pull request addresses #1025 and #1026

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
Thanks. Well have to wait until dev is passing again. Meanwhile, there are some style issues with this PR. For example if urbg is an input it should be const.

jdumas:
Technically, the random number generator `URBG` should be passed as a rvalue reference, similar to the signature of [std::shuffle](https://en.cppreference.com/w/cpp/algorithm/random_shuffle).

ycjungSubhuman:
@alecjacobson Thanks for reviewing. Yes. I guess I'll have to wait until the issue with CGAL is resolved.  About the style, I forgot to make `urbg` const reference and put it before output parameters. I'll try to change this.

ycjungSubhuman:
@jdumas True... If std::shuffle takes rvalue reference, may be a good idea to follow it. However, I have a feelling that `const URGB urgb` can be a valid option, too. If a user wants to pass a URGB using move semantics, he or she can just call `randperm(n, std::move(urgb), I)`

ycjungSubhuman:
I guess `URBG &&urbg` is better because it is intended to be used that way, as shown in the example of `std::shuffle` signature. I'm more inclined towards this approach.

jdumas:
Yes, let's not try to be smart and follow the signature of the stdlib function :D

Issue #1061, opened 2018-12-27
The recent commit merging"jdumas-fix-cgal" seems to break something about boost lib ...


metorm:
Minutes ago I pulled the latest dev branch to my local repository and my current project fails to build.
The problem is on `boost`. My project uses the `MT` version of `boost` and compiles well with the older versions.
However, with the latest version, a fatal error "Mixing a dll boost library with a static runtime is a really bad idea" is triggered. I checked the history and it turned out [the commit "Merge branch 'jdumas-fix-cgal' into dev"](https://github.com/libigl/libigl/commit/48c7fa98c1e01472141a2640bd237ce18c7d85b6) triggered the error.

![image](https://user-images.githubusercontent.com/12082680/50481185-286ae680-0a1b-11e9-880e-d2ca37d8a7be.png)

Anyone please help with this?

#### Check all that apply (change to `[x]`)
- [x] Windows
- [ ] Mac OS X
- [ ] Linux

jdumas:
This is a known issue, see #1057, we're not sure what's the best way to fix it for now.

alecjacobson:
Hopefully this is fixed now.

Issue #1055, opened 2018-12-24
Template errors in mesh_to_polyhedron


alecjacobson:
New templates are in the wrong order 

https://github.com/libigl/libigl/blob/8df183bc460a5cd90575f399136ffd05f3ffc07d/include/igl/copyleft/cgal/mesh_to_polyhedron.h#L32

Meanwhile, the explicit template instantiations are failing to compile (not tested before merging PR? not being tested in the CI?)

```
/usr/local/libigl/include/igl/copyleft/cgal/mesh_to_polyhedron.cpp:53:36: error: explicit instantiation of 'mesh_to_polyhedron' does not refer to a
      function template, variable template, member function, member class, or static data member
template bool igl::copyleft::cgal::mesh_to_polyhedron<CGAL::Polyhedron_3<CGAL::Simple_cartesian<double>, CGAL::Polyhedron_items_with_id_3, CGAL::Ha...
                                   ^
/usr/local/libigl/include/igl/copyleft/cgal/mesh_to_polyhedron.cpp:16:38: note: candidate template ignored: could not match 'MatrixBase' against 'Matrix'
IGL_INLINE bool igl::copyleft::cgal::mesh_to_polyhedron(
                                     ^
/usr/local/libigl/include/igl/copyleft/cgal/mesh_to_polyhedron.cpp:54:36: error: explicit instantiation of 'mesh_to_polyhedron' does not refer to a
      function template, variable template, member function, member class, or static data member
template bool igl::copyleft::cgal::mesh_to_polyhedron<CGAL::Polyhedron_3<CGAL::Simple_cartesian<float>, CGAL::Polyhedron_items_with_id_3, CGAL::Hal...
                                   ^
/usr/local/libigl/include/igl/copyleft/cgal/mesh_to_polyhedron.cpp:16:38: note: candidate template ignored: could not match 'MatrixBase' against 'Matrix'
IGL_INLINE bool igl::copyleft::cgal::mesh_to_polyhedron(
                                     ^
2 errors generated.
```

jdumas:
Ah, my bad. #1051 introduced this change, and since it was compiling on Travis I didn't think twice before merging it. But you're right that CGAL is currently disabled on Travis + AppVeyor! See my current pending PR #996 on this topic ... (unfortunately some shady stuff is making unit tests randomly fail due to some uninitialized memory issues I guess? So this PR hasn't been merged yet).

alecjacobson:
[...]

PR #1064, opened 2018-12-28
Dev fix cgal

jdumas:
Following discussion in #1057 and . This should also take care of #1055 and #1061.

I've enabled both Python + CGAL on Travis, and they seem to build and run properly!

- [x] Adds corresponding python binding.
- [x] This is a minor change.


jdumas:
The memory issue was bugging me so I took a look and compiled libigl with `-DCMAKE_CXX_FLAGS="-fsanitize=address"` to detect memory leaks/overflow, etc. I've fixed a stack-buffer overflow in tetgen (unlikely to cause any real issue in practice). Then there is a bug in Eigen triggered by `tests/include/igl/cotmatrix.cpp`, as it sems that `L1.col(f).sum()` access some out-of-bound memory (the matrix is stored row-major, and summing over columns seems to cause the issue). I've tried with the latest version of Eigen (3.3.7), and the problem seems to be fixed, although I had to fix a bunch of compilation errors along the way. Then remaining issues detected by AddressSanitizer seem to be memory leaks in the loaders. One in `readDMAT` was easy to fix (a `new` that was never deleted). The others is `ply.h` are more tricky, and I think we should switch to a more modern implementation of this reader, e.g. this [tinyply](https://github.com/ddiakopoulos/tinyply) looks promising.

Still not sure what is causing the CSGTree test to randomly fail, but we should probably try again once we fix the memory leaks in the PLY reader, and with a more recent version of Eigen.

I've also enabled compilation against Eigen 3.3.7 as an additional build on Travis, for good measures (it'll be easier to not introduce code that doesn't compile against new Eigen versions).

alecjacobson:
Thanks. I like the idea of adding Eigen 3.3. to the CI to smooth over the transition. I'll merge this despite the memory issue, since we're not changing that in this PR.

PR #1067, opened 2018-12-28
Extend boundary_facets with outputs revealing which element facet comes from

alecjacobson:
Also made some friendly sort overloads and fixed up some style things along the way. Corresponding unit tests.

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
@alecjacobson compilation seems to fail on Windows on your PR (and this is not a timeout of AppVeyor).

PR #1079, opened 2019-01-27
[WIP] Add voxelization function

cedesu:
The voxelization is partly based on flood fill. And it uses igl::winding_number.

- [x] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
I don't really get it. Why do you need a flood-fill if you are calling `igl::winding_number`?
Also, you need to use 2-space indents, and indent your comments as well. There are other style issues in your code, such as braces.

cedesu:
Sorry I mentioned the wrong algorithm name, it is actually scanline fill algorithm.

This program is basicly through two steps.
1. First solidify a shell that represents the surface.
    (I use winding_number to decide whether it's inside or not, because it is robust in all cases)
2. Fill it by scanline fill algorithm.
image

I can provide more detailed explanation if needed.

And I turn the tab to 2 spaces. What else need to be done?

jdumas:
Ok this makes sense. Your implementation seems dreadfully complicated though, and not very efficient (just eyeballing the code itself). I'll let others comment on the algorithmic side, but on the technical side I will make the following comments:
- Brace style needs to be changed to comply with the [style guidelines](https://libigl.github.io/style-guidelines/).
- Code needs to compile on both Travis and AppVeyor.
- You should remove the `#include <igl/read_triangle_mesh.h>`, it is not needed.
- You should never use a `std::list<>` container, it is really inefficient (lots of fragmented `malloc`). Use a `std::vector`, `std::dequeue`, `std::stack` or `std::queue` instead, depending on what you need.
- Output layout should be documented (which of X, Y or X is the major index).
- Output should be a `std::vector<bool>`, or a `Eigen::Matrix<bool, ...>` (the latter if you need parallel writing into the array).
- Internal variables are poorly named (`trans`, `trans1`, `now`, `now1`, really?).
- Finally, it would be nice to have a unit test since it's a new feature. 

Now, on the algorithmic side, I'm wondering if it might make sense to integrate the [FastWindingNumber](https://github.com/sideeffects/WindingNumber) for triangle meshes to be able to handle imperfect inputs instead. If we assume the input to be watertight, ray-casting will also be a much faster alternative (but it will not necessarily produce a conservative rasterization).

alecjacobson:
closing as stale. feel free to reopen

PR #1088, opened 2019-02-18
Fixed undefined behavior in the SolveP3 function of flip_avoiding_lin…

breannansmith:
…e_search.

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
I don't think this is undefined behavior, but it's probably more clear this way.

PR #1093, opened 2019-02-25
Explicit template instantiation of std::vector<std::vector<int>> for boundary_loop

ShuangLiu1992:
Explicit template instantiation of std::vector<std::vector<int>> for boundary_loop

- [ ] Adds new .cpp file.
- [x] This is a minor change.


ShuangLiu1992:
the compilation errors seem to come from clang + CGAL, it should be from my changes

PR #1097, opened 2019-02-27
Fix igl common cmake config

### LINKED ISSUE #1077: use libigl as external project (modern cmake build with targets)
Hi,

I am using a very recent build of libigl (>=ver. 2) (compiled as a static lib, but this should not be the issue) in visual studio 2017 and I use it as an external dependency:

```cmake
find_package(libigl CONFIG REQUIRED core common)
if(NOT TARGET igl::core OR NOT TARGET igl::common)
  message (FATAL_ERROR "libigl not found!")
endif()

(...)

add_library(MyFooLib ${some_sources})
target_link_libraries(MyFooLib PUBLIC igl::core igl::common)
```

At the configure stage, CMake complains that despite it found libigl config, libigl package is considered as not found:

```
CMake Error at CMakeLists.txt:76 (find_package):
  Found package configuration file:

    C:/Programming/Libs/Git/Build/x64/vs2017/libigl/libigl-config.cmake

  but it set libigl_FOUND to FALSE so package "libigl" is considered to be
  NOT FOUND.
```
By inspecting generated libigl-config.cmake, the issue is that some code like:

```cmake
if(TARGET igl::common)
  set(libigl_common_FOUND TRUE)
endif()
```

is missing after group `if (TARGET igl::core) (...) endif()` to indicate that target igl::common has been properly found (as it is done for target core).

Am I missing something about the use of new cmake targets? The libgl-example-project does something similar (except that I am specifically calling the find_package in CONFIG mode).

Thanks!

ithron:
- Added a missing check for the igl::common target in libigl-config.cmake.in which resulted in an error if a user requested the common component using `find_package(libel CONFIG REQUIRED core common)`. This also fixes #1077 
- libigl-config.cmake was using `target_link_libraries` to add the Eigen3 dependency to the target igl::core and igl::common at configuration time. This results in an error. Fixed by replacing with `set_target_properties`
- Fixed: .c files were not installed

Tested the changes with the following CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.10)

# Use hunter to install Eigen3 for convenience
include("cmake/HunterGate.cmake")
HunterGate(
  URL "https://github.com/ruslo/hunter/archive/v0.23.120.tar.gz"
  SHA1 "7f43361ab4142e7d0eb03488b4b654aad38cf0cb"
)

project(Test)

hunter_add_package(Eigen)
find_package(Eigen3 CONFIG REQUIRED)

find_package(libigl CONFIG REQUIRED core common)
if(NOT TARGET igl::core OR NOT TARGET igl::common)
    message (FATAL_ERROR "libigl not found!")
endif()

add_executable(hello test.cpp)
target_link_libraries(hello igl::core igl::common)
```

Note that the "hunter steps" are not necessary if Eigen3 is installed on the system.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
Why is `set_target_properties()` working, but not `target_link_libraries()`?

ithron:
The error you get using `target_link_libraries()` is

> Cannot specify link libraries for target "igl::core" which is not built by
  this project.

The cmake documentation says:

> The named target must be created by add_library() within the project or as an IMPORTED library.

I think `target_link_libraries()` works on IMPORTED targets but not on INTERFACE IMPORTED libraries.

jdumas:
This looks like a limitation of CMake. I found some discussion [there](https://gitlab.kitware.com/cmake/cmake/issues/17197), but cannot pinpoint an exact thread about this `INTERFACE IMPORTED` business. I mean if you PR works around that, should be fine.

PR #1108, opened 2019-03-07
Extending igl::cat to work with std::vectors AND Allowing N-gons in writeOBJ

lawsonfulton:
I've extended igl::cat to accept a std::vector of matrices. I've also added a test case.

Also in this pull request: Template instantiations and a bug fix for adjacency_list (it wasn't working with std::vector).

- [x] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/adjacency_list.cpp
```diff
@@ -12,7 +12,7 @@
 
 template <typename Index, typename IndexVector>
 IGL_INLINE void igl::adjacency_list(
-    const Eigen::MatrixBase<Index>  & F,
+    const Eigen::PlainObjectBase<Index>  & F,
```
jdumas: Why change this to `PlainObjectBase`? We should use `MatrixBase` for inputs.

include/igl/writeOBJ.h
```diff
@@ -50,6 +51,18 @@ namespace igl
...
 
+  // Write a mesh of mixed tris and quads to an ascii obj file
+  // Inputs:
+  //   str  path to outputfile
+  //   V  #V by 3 mesh vertex positions
+  //   F  #F std::vector of std::vector<Index> of size 3 or 4 mesh indices into V
+  // Returns true on success, false on error
+  template <typename DerivedV>
+  IGL_INLINE bool writeOBJ(
+    const std::string str,
```
jdumas: Should be `const std::string &`.

include/igl/writeOBJ.h
```diff
@@ -50,6 +51,18 @@ namespace igl
...
+    const Eigen::MatrixBase<DerivedV>& V,
+    const std::vector<std::vector<typename DerivedV::Index> >& F);
```
jdumas: This probably ought to work for any integer type `T` (doesn't have to be `DerivedV::Index`).

include/igl/writeOBJ.cpp
```diff
@@ -118,6 +118,35 @@ IGL_INLINE bool igl::writeOBJ(
...
+  if(!s.is_open())
+  {
+    fprintf(stderr,"IOError: writeOBJ() could not open %s\n",str.c_str());
+    return false;
+  }
+  s<<V.format(IOFormat(FullPrecision,DontAlignCols," ","\n","v ","","","\n"));
+  
+  for(const auto& face : F) {
+    int face_size = face.size();
+    assert(face_size == 3 || face_size == 4);
```
jdumas: `.obj` can support polygonal faces with > 4 vertices. We could also support line graphs, by replacing the prefix `f` with `l` in case `face_size == 2`.

include/igl/adjacency_list.cpp
zero-impact: My bad, this is because the templates were written with PlainObjectBase. I will switch it back.

include/igl/writeOBJ.cpp
zero-impact: Good point. It turns out face statements with two vertices (edges) are supported by some programs (Blender). So it might make sense to allow that.

jdumas: Actually, meshlab will open a .obj where edges are marked using the `l` marker, but will complain that faces should have `>= 3` vertices if you use `l`. I'm pretty sure blender supports both, so I would rather use `l` if `face_size == 2`.

include/igl/writeOBJ.cpp
```diff
@@ -118,6 +118,43 @@ IGL_INLINE bool igl::writeOBJ(
...
+    return false;
+  }
+  s<<V.format(IOFormat(FullPrecision,DontAlignCols," ","\n","v ","","","\n"));
+  
+  for(const auto& face : F) {
+    int face_size = face.size();
+    
+    assert(face_size != 0);
+    
+    if(face_size == 2) {
```
jdumas: Btw, the braces in your commit do not respect the [style guideline](https://libigl.github.io/style-guidelines/). Can you please fix that? For this specific line, I would use a trigraph instead: `s << (face_size == 2 ? "l" : "f")`

include/igl/adjacency_list.cpp
```diff
@@ -132,8 +132,16 @@ IGL_INLINE void igl::adjacency_list(
   const std::vector<std::vector<Index> > & F,
   std::vector<std::vector<Index> >& A)
 {
-  A.clear(); 
-  A.resize(F.maxCoeff()+1);
+  A.clear();
+  
+  // Find maxCoeff
+  int maxCoeff = 0;
+  for(const auto &vec : F) {
```
jdumas: Braces need to be update here too.

include/igl/cat.cpp
```diff
@@ -251,6 +251,63 @@ IGL_INLINE void igl::cat(const std::vector<std::vector< Mat > > & A, Mat & C)
...
+    C.resize(tot_rows, A_cols);
+
+    int cur_row = 0;
+    for(int i = 0; i < num_mat; i++)
+    {
+      assert(A_cols == A[i].cols());
+      C.block(cur_row,0,A[i].rows(),A_cols) = A[i];
+      cur_row += A[i].rows();
+    }
+  }else if(dim == 2)
```
jdumas: And go to newline after this brace.

zero-impact: Sorry, thanks for your patience with this.

I guess I don't actually know what the preferred brace-style is completely. I copied this style from earlier in the file. As far as I can tell, the [style guidelines](https://libigl.github.io/style-guidelines/) does not address braces.

include/igl/adjacency_list.cpp
zero-impact: Thanks

include/igl/cat.cpp
jdumas: Hmm you're right. I've been trying to match the already existing codestyle. I think braces on a newline for `for` and `if` are a waste of space, but that's not up to me. I'll see if we can update the guideline accordingly.

zero-impact:
Thanks @jdumas I think this is all set. Is there anything I still need to do?

include/igl/adjacency_list.cpp
```diff
@@ -132,8 +132,18 @@ IGL_INLINE void igl::adjacency_list(
   const std::vector<std::vector<Index> > & F,
   std::vector<std::vector<Index> >& A)
 {
-  A.clear(); 
-  A.resize(F.maxCoeff()+1);
+  A.clear();
+  
+  // Find maxCoeff
+  int maxCoeff = 0;
```
jdumas: This should be of type `Index`.

include/igl/boundary_loop.cpp
```diff
@@ -130,7 +130,7 @@ IGL_INLINE void igl::boundary_loop(
 
 template <typename DerivedF, typename DerivedL>
 IGL_INLINE void igl::boundary_loop(
-  const Eigen::MatrixBase<DerivedF>& F,
+  const Eigen::PlainObjectBase<DerivedF>& F,
```
jdumas: This should be of type `MatrixBase` (see FAQ).

include/igl/adjacency_list.cpp
zero-impact: Thanks

include/igl/boundary_loop.cpp
zero-impact: Thanks

zero-impact:
Hopefully that does it :)

include/igl/boundary_loop.cpp
```diff
@@ -131,7 +131,7 @@ IGL_INLINE void igl::boundary_loop(
 template <typename DerivedF, typename DerivedL>
 IGL_INLINE void igl::boundary_loop(
   const Eigen::MatrixBase<DerivedF>& F,
-  Eigen::PlainObjectBase<DerivedL>& L)
+  Eigen::MatrixBase<DerivedL>& L)
```
jdumas: Hmm  sorry, `MatrixBase` should be used for *inputs*, and `PlainObjectBase` should be used for *output* (see [FAQ](https://libigl.github.io/faq/)).

zero-impact: Yup, just noticed this.. fixing

Issue #919, opened 2018-10-11
readPLY probably leaks memory


jimmyland:
Based on just reading over the code, the functions in ply.h do a lot of mallocs (`myalloc`s) and largely don't appear to have corresponding free calls, so e.g. ply_get_element_description allocates for each description and igl::readPLY doesn't appear to free this memory.

teseoch:
there is also no test/tutorial using it

qnzhou:
I can confirm this issue.  Repeatedly reading a ply file does increase the memory consumption over time on mac.

jdumas:
We should consider removing the current dependency on `ply.h` in favor of a more modern implementation, e.g. I've found https://github.com/ddiakopoulos/tinyply which looks neat.

jimmyland:
tinyply is neat but fwiw note that it does not support loading files with a mix of tris and quads currently (when loading lists of lists, it doesn't seem to tell you how long each sub-list is; it wants to assume they are all the same length), though the developer plans to address this for the next version

jdumas:
Libigl's matrix representation also does not map well to hybrid meshes that are a mix of quads and triangles, so I don't expect this will be an issue (in fact, very few libigl functions support quad meshes at all).

jdumas:
I don't think any of the loader is doing that at the moment. And it's not really the job of the loader to triangulate quad faces on read (after all, there are several ways to triangulate a quad face, e.g. you may decide to add a vertex in each quad barycenter to avoid directional bias, etc.).

jimmyland:
ah, you're right that the other loaders do not triangulate on load, but the OBJ loader does return false rather than allow you to incorrectly load a mixed tri/quad file as if it were just a tri file

alecjacobson:
use the std vector std vector version.

jdumas:
And of course there are tons of ply io libraries out there, e.g. look at [this benchmark](https://github.com/mhalber/ply_io_benchmark). There's all single files without dependencies. I guess tinyply may be easier to use (requires fewer loc), but once it's wrapped up it doesn't matter much.

[...]

PR #1120, opened 2019-03-21
Fixed two memory leaks

robertguetzkow:
As mentioned in #919 the ply.h implementation has several code quality issues. The suggested changes fix two memory leaks. One is caused by a missing free, another from improper handling when realloc fails. Additionally UB has been identified, because a pointer to a local variable is returned.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/ply.h
+++ b/include/igl/ply.h
@@ -539,8 +539,11 @@ inline PlyFile *ply_open_for_writing(
 
   fp = fopen (name, "w");
   if (fp == NULL) {
+    free(name);
     return (NULL);
   }
+
+  free(name);
 
   /* create the actual PlyFile structure */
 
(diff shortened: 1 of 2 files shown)


robertguetzkow:
From what I've seen the implementation needs a major overhaul or a complete replacement. There are also string processing operations that don't look safe.

jdumas:
As discussed in the thread ideally we should replace it by another library entirely, e.g. tinyply. I believe dropping the file in `include/igl/` and making it safe for header-only inclusion (e.g. cleanup the `using namespace std;` in the cpp) is the way to go.

robertguetzkow:
@jdumas that would be the best solution. I just thought in the meantime fixing the most obvious bugs in the current implementation could help a little.

jdumas:
I mean if you want to give it a shot at integrating tinyply or rply, feel free to do so!

PR #1125, opened 2019-03-22
Memory leak fixes

robertguetzkow:
This pull request fixes several memory leaks.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/principal_curvature.cpp
+++ b/include/igl/principal_curvature.cpp
@@ -479,16 +479,16 @@ IGL_INLINE void CurvatureCalculator::getSphere(const int start, const double r, 
 {
   int bufsize=vertices.rows();
   vv.reserve(bufsize);
-  std::list<int>* queue= new std::list<int>();
-  bool* visited = (bool*)calloc(bufsize,sizeof(bool));
-  queue->push_back(start);
+  std::list<int> queue;
+  bool visited[bufsize];
+  queue.push_back(start);
   visited[start]=true;
   Eigen::Vector3d me=vertices.row(start);
-  std::priority_queue<std::pair<int, double>, std::vector<std::pair<int, double> >, comparer >* extra_candidates= new  std::priority_queue<std::pair<int, double>, std::vector<std::pair<int, double> >, comparer >();
-  while (!queue->empty())
+  std::priority_queue<std::pair<int, double>, std::vector<std::pair<int, double> >, comparer > extra_candidates;
+  while (!queue.empty())
   {
-    int toVisit=queue->front();
-    queue->pop_front();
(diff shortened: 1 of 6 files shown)


robertguetzkow:
The improvements should be done now. Made some painfully obvious mistakes in the first commits because having a cold & coding don't match, sorry for that.

robertguetzkow:
AppVeyor fails because it doesn't finish within 60 minutes and aborts the build process.

jdumas:
Yeah it's annoying... It's a hard limit of AppVeyor and I don't know how we can speed up compilation of the Windows build. I'll restart it until it goes through. Thanks for your patience!

include/igl/principal_curvature.cpp
```diff
@@ -522,8 +522,12 @@ IGL_INLINE void CurvatureCalculator::getSphere(const int start, const double r,
       }
     }
   }
-  free(extra_candidates);
-  free(queue);
+  while(!extra_candidates->empty()){
```
jdumas: I believe this while and the clear are unnecessary. The `delete` operator should call the destructor of both variables. Honestly looking at the code you might as well remove the `new` operators altogether (and this is also true for the `bool* visited` which is `calloc`ed.

robertguetzkow: [You're right](https://en.cppreference.com/w/cpp/container/priority_queue/~priority_queue), my C++ skills are a bit rusty. I'll commit the changes to the function. Unnecessary allocations with `new` happen in other code as well, which may need to be fixed in the future.

robertguetzkow: BTW a potential fix for the slow AppVeyor build could be `/m:2` in the `MSBuildOptions` to use two threads for the two cores available in the VMs.

robertguetzkow:
I'll revert the change for the bool array, because it requires a size known at runtime which isn't support by C++ standard (as far as I'm aware). So it has to be allocated or we change it to a vector.

include/igl/principal_curvature.cpp
```diff
@@ -448,7 +448,8 @@ IGL_INLINE void CurvatureCalculator::getKRing(const int start, const double r, s
   int bufsize=vertices.rows();
   vv.reserve(bufsize);
   std::list<std::pair<int,int> > queue;
-  bool* visited = (bool*)calloc(bufsize,sizeof(bool));
+  std::vector<bool> visited;
```
jdumas: Can you initialize it to false as well? Using the constructor `std::vector<bool> visited(bufsize, false);`

robertguetzkow:
Are there any additional changes that you need before merging?

PR #1131, opened 2019-03-28
Templates and overloads

alecjacobson:

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
Can anyone tell what went wrong with the AppVeyor builds? I can't see what the errors are. Did it just time out?

jdumas:
Yes sometimes the build randomly times out on AppVeyor (individual builds are capped at 60min times, even on paid plans). This is annoying and I have no solution outside of restarting the build until it gets on a faster machine... (then it takes usually about 35min).

The fact that you build times out on *both* debug and release however, suggests that the new function you've added take quite some time to compile. On Linux/macOS we have ccache to speed up incremental builds, but I haven't found any such system for Windows, and I'm really out of ideas for how to improve this situation.

jdumas:
We should find a way to fix this issue with AppVeyor. Merging stuff into dev without running the tests and compile everything on Windows is opening the door to trouble in the long term. Inria has a [CI platform](https://ci.inria.fr/) for INRIA projects, where one can configure a Jenkins server. Jenkins is admittedly more complicated than AppVeyor, but hopefully this is the kind of things to do only once...

alecjacobson:
I'm leaning toward just setting up a fleet of machines to test on. AppVeyor is too annoying and even travis is too slow.

jdumas:
Agreed. Do you have some resources at your University you could exploit? Setting up and administrating the machines yourself is  a chore, but if you have some IT department that offer that kind of service it's a great load off the shoulders (and I imagine you're not the only one with an open-source project in need of CI in the CS department). I don't know how it is at NYU for @danielepanozzo though.

robertguetzkow:
@alecjacobson @jdumas Have you tried `/m:2` in the config for a multithreaded build?

danielepanozzo:
Do you know if both AppVeyor and Travis will get better if we use a paid account and how much this will cost? I should be able to cover for it if the cost is not unreasonable.

alecjacobson:
I seem to remember that it's unreasonable.

jdumas:
@robertguetzkow I think it's already enabled in `.appveyor.yml`.

@danielepanozzo Can you ask the IT department if they have any kind of service for continuous integration? On AppVeyor the time limit exists also for paid plans. You might get faster machines by default on a paid plan, but that would only delay the problem.

robertguetzkow:
@jdumas You're right, it's enabled in the dev-branch and I was looking at the master-branch.

PR #1132, opened 2019-03-28
expose label color

alecjacobson:
Can change base:dev after templates-and-overloads is merged into dev.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


include/igl/opengl/glfw/imgui/ImGuiMenu.cpp
```diff
@@ -113,7 +113,9 @@ IGL_INLINE bool ImGuiMenu::mouse_down(int button, int modifier)
 
 IGL_INLINE bool ImGuiMenu::mouse_up(int button, int modifier)
 {
-  return ImGui::GetIO().WantCaptureMouse;
+  //return ImGui::GetIO().WantCaptureMouse;
+  // !! Should not steal mouse up
```
jdumas: Why this change?

alecjacobson: This is just general GUI good manners. Don't steal mouse up. Everyone should be allowed to know if mouse up happened.

jdumas: I'm not sure about that. Why is that different from a mouse down event? This flag means that mouse event happens on a imgui widget, so it should not be dispatched/may be hidden from the rest of the application. But maybe you have a specific example of interaction where this was problematic?

jdumas:
Is there any reason you're in such a rush to merge those changes into `dev`? I'm still not sure about whether `ImGui::GetIO().WantCaptureMouse` should be commented out...

PR #1138, opened 2019-04-03
Update issue_template.md

jdumas:
We've been having a couple of issues reported against the master branch, which are already fixed in dev. We already have a [checklist](https://libigl.github.io/CONTRIBUTING/) on the website but nobody seems to read it. Seems to me we should at least add a line in the issue template about whether or not the user is on the `dev` branch. I'm not sure if we should make it another checklist though (people seems to read checklists).

- [x] This is a minor change.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,8 +1,10 @@
 [Describe your issue]
 
+[**If you are using the `MASTER` branch, please check out the `DEV` branch before reporting any issue**]
+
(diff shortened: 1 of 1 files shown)


jdumas:
Yeah but what's the point of users reporting issues that are already fixed in the `dev` branch?

alecjacobson:
to encourage regular merges of dev into master

jdumas:
What are we waiting for then :D

Issue #1133, opened 2019-03-29
Compilation errors using Viewer since multi-viewport changes.


alecjacobson:
In this compilation, `IGL_STATIC_LIBRARY` is _not_ defined and I'm in debug mode.

```
In file included from /Users/ajx/Dropbox/joints/designer/designer.cpp:22:
In file included from /usr/local/libigl/cmake/../include/igl/opengl/glfw/Viewer.h:17:
In file included from /usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerCore.h:12:
In file included from /usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerData.h:300:
/usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerData.cpp:128:5: error: member access into incomplete type 'const igl::opengl::ViewerCore'
  to.set(show_overlay      , from.is_set(show_overlay)      );
    ^
/usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerData.h:38:7: note: forward declaration of 'igl::opengl::ViewerCore'
class ViewerCore;
      ^
In file included from /Users/ajx/Dropbox/joints/designer/designer.cpp:22:
In file included from /usr/local/libigl/cmake/../include/igl/opengl/glfw/Viewer.h:17:
In file included from /usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerCore.h:12:
In file included from /usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerData.h:300:
/usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerData.cpp:129:5: error: member access into incomplete type 'const igl::opengl::ViewerCore'
  to.set(show_overlay_depth, from.is_set(show_overlay_depth));
    ^
/usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerData.h:38:7: note: forward declaration of 'igl::opengl::ViewerCore'
class ViewerCore;
      ^
In file included from /Users/ajx/Dropbox/joints/designer/designer.cpp:22:
In file included from /usr/local/libigl/cmake/../include/igl/opengl/glfw/Viewer.h:17:
In file included from /usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerCore.h:12:
In file included from /usr/local/libigl/cmake/../include/igl/opengl/glfw/../ViewerData.h:300:
[...]

PR #1139, opened 2019-04-04
[travis-ci] added build configs to ensure both, static as well as hea…

BruegelN:
…der-only mode is build for gcc-7 (ubunutu) and clang(osx).

Results can be viewed at https://travis-ci.org/BruegelN/libigl.
Currently only gcc-7 and clang on macOS are tested with both config(static + header-only).
For sure the both newly introduced builds fail, because the actual bug wasn't fixed so far (see #1133)
Is this what you suggested @jdumas?

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,10 @@ matrix:
     compiler: gcc-7
     env:
     - MATRIX_EVAL="export CC=gcc-7 CXX=g++-7 CONFIG=Release PYTHON=python3"
+  - os: linux # same config like above but with -DLIBIGL_USE_STATIC_LIBRARY=OFF to test static and header-only builds
+    compiler: gcc-7
+    env:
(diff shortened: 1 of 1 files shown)


jdumas:
The bug #1133 is fixed by #1134, I'm just waiting for someone else to merge my PR ...

BruegelN:
@jdumas sorry I didn't meant to offend you or anybody else! The 
> For sure the both newly introduced builds fail, because the actual bug wasn't fixed so far (see #1133)

was only intended as a note, that my changes will fail because the bugfix is not yet merged into dev. For that I used the issue and not the PR to fix this issue. I may have been unclear, sorry :smiley:

jdumas:
No worries. We can merge your PR once the fix #1134 is merged.

PR #1141, opened 2019-04-05
 robust intersection test for 2d segments using cgal orient2d

hankstag:
A more efficient and robust way to test whether 2d segments intersect each other without solving a linear system

- [x] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Why closing this?

hankstag:
Since James have already extracted the predicate check from CGAL, i think it makes more sense to directly use that instead of using orient2D from CGAL. I'm gonna create a new pr for that.

hankstag:
Ah i see, actually i don't need to create a new pr but just stick to this one

Issue #1019, opened 2018-11-19
CMake error on Ubuntu 18.04


gulib:
Hello,

I was trying to install newer version of libigl while faced the following error

```
CMake Error at .cache/eigen/eigen-download-prefix/src/eigen-download-stamp/download-eigen-download.cmake:159 (message):
  Each download failed!

    error: downloading 'http://bitbucket.org/eigen/eigen/get/3.2.10.tar.gz' failed
         status_code: 1
         status_string: "Unsupported protocol"
         log:
         --- LOG BEGIN ---
           Trying 18.205.93.0...

  TCP_NODELAY set

  Connected to bitbucket.org (18.205.93.0) port 80 (#0)

  GET /eigen/eigen/get/3.2.10.tar.gz HTTP/1.1

  Host: bitbucket.org

  User-Agent: curl/7.58.0

  Accept: */*

  

  HTTP/1.1 301 Moved Permanently

  Content-Type: text/html

  Date: Mon, 19 Nov 2018 05:50:28 GMT

  Location: https://bitbucket.org/eigen/eigen/get/3.2.10.tar.gz

  Connection: Keep-Alive

  Content-Length: 0

  

  Connection #0 to host bitbucket.org left intact

  Issue another request to this URL:
  'https://bitbucket.org/eigen/eigen/get/3.2.10.tar.gz'

  Protocol "https" not supported or disabled in libcurl

  Closing connection -1

  

         --- LOG END ---
         error: downloading 'http://bitbucket.org/eigen/eigen/get/3.2.10.tar.gz' failed
         status_code: 1
         status_string: "Unsupported protocol"
         log:
         --- LOG BEGIN ---
           Trying 18.205.93.0...

  TCP_NODELAY set

  Connected to bitbucket.org (18.205.93.0) port 80 (#0)

  GET /eigen/eigen/get/3.2.10.tar.gz HTTP/1.1

  Host: bitbucket.org

  User-Agent: curl/7.58.0

  Accept: */*

  

  HTTP/1.1 301 Moved Permanently

  Content-Type: text/html

  Date: Mon, 19 Nov 2018 05:50:28 GMT

  Location: https://bitbucket.org/eigen/eigen/get/3.2.10.tar.gz

  Connection: Keep-Alive

  Content-Length: 0

  

  Connection #0 to host bitbucket.org left intact

  Issue another request to this URL:
  'https://bitbucket.org/eigen/eigen/get/3.2.10.tar.gz'

[...]

PR #1153, opened 2019-04-16
Update GLFW to 3.3

jdumas:
Stable version of GLFW 3.3 came out today, so I thought we might update our dependency accordingly.
I've also updated Embree and use git urls instead of direct download due to #1019.

- [x] This is a minor change.


alecjacobson:
Has this been tested on Mac, Ubuntu and Windows ? (not just CI but actually testing that the tutorial windows appear correctly etc.)

jdumas:
I have tested tutorials 102, 107 and 108 on my Archlinux and they work as expected. I can test on Windows too since we don't have many Windows users, but I don't have a Mac.

jdumas:
Also tested on Windows, and it works fine. I also fixed a cmake issue with top-level project detection (somehow was not working on Windows).

BruegelN:
@jdumas @alecjacobson just wanted to let you know: everything looks fine and behaves as I'd expected it on my MacBook(macOS Mojave 10.14.4)!
I've tested 102, 107 and 108 as well as 106 & 708.

To do so I took the following steps:
```
git clone https://github.com/libigl/libigl.git
cd libigl
git checkout dev
git pull origin update-libs
mkdir build
cd build
cmake ../
make -j4
```

PR #1159, opened 2019-04-17
Fixing bug preventing compilation of sparse marching cubes

zero-impact:
The sparse index version of marching cubes was trying to access an Eigen vector like a std::vector, preventing compilation.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/copyleft/marching_cubes.cpp
+++ b/include/igl/copyleft/marching_cubes.cpp
@@ -223,7 +223,7 @@ public:
       // determine cube type
       for (int i=0; i<8; ++i)
       {
-        if (values[cube[i]] > isovalue)
+        if (values(cube[i]) > isovalue)
         {
           cubetype |= (1<<i);
         }
@@ -303,8 +303,8 @@ public:
     const Eigen::Matrix<typename DerivedPoints::Scalar, 1, 3> & p0 = points.row(i0);
     const Eigen::Matrix<typename DerivedPoints::Scalar, 1, 3> & p1 = points.row(i1);
 
-    typename DerivedValues::Scalar s0 = fabs(values[i0]);
-    typename DerivedValues::Scalar s1 = fabs(values[i1]);
+    typename DerivedValues::Scalar s0 = fabs(values(i0));
+    typename DerivedValues::Scalar s1 = fabs(values(i1));
     typename DerivedValues::Scalar t  = s0 / (s0+s1);
 
     num_vertices++;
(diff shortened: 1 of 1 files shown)


jdumas:
Eigen matrices with one dimension == 1 can be accessed using the `[]` operator, so I am surprised this would prevent compilation. Can you paste the error message you are getting?

zero-impact:
You're right, when I switch my definition of values from `MatrixXd` to `VectorXd` it compiles fine. But, I'd say it is still reasonable to let values take a rows*1 `MatrixXd`. Up to you.

zero-impact:
Oh and the error is
```
/Users/lawson/Workspace/lattice_meshing_workspace/libigl/libigl/cmake/../external/eigen/Eigen/src/Core/DenseCoeffsBase.h:153:7: error: static_assert failed due to requirement 'Matrix<double, -1, -1, 0, -1, -1>::IsVectorAtCompileTime' "THE_BRACKET_OPERATOR_IS_ONLY_FOR_VECTORS__USE_THE_PARENTHESIS_OPERATOR_INSTEAD"
      EIGEN_STATIC_ASSERT(Derived::IsVectorAtCompileTime,
      ^                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /Users/lawson/Workspace/lattice_meshing_workspace/libigl/libigl/include/igl/copyleft/marching_cubes.cpp:26:
In file included from /Users/lawson/Workspace/lattice_meshing_workspace/libigl/libigl/include/igl/copyleft/marching_cubes.h:12:
In file included from /Users/lawson/Workspace/lattice_meshing_workspace/libigl/libigl/cmake/../external/eigen/Eigen/Core:254:
/Users/lawson/Workspace/lattice_meshing_workspace/libigl/libigl/cmake/../external/eigen/Eigen/src/Core/util/StaticAssert.h:32:40: note: expanded from macro 'EIGEN_STATIC_ASSERT'
    #define EIGEN_STATIC_ASSERT(X,MSG) static_assert(X,#MSG);
                                       ^             ~
/Users/lawson/Workspace/lattice_meshing_workspace/libigl/libigl/include/igl/copyleft/marching_cubes.cpp:226:13: note: in instantiation of member function 'Eigen::DenseCoeffsBase<Eigen::Matrix<double, -1, -1, 0, -1, -1>, 0>::operator[]' requested here
        if (values[cube[i]] > isovalue)
```

jdumas:
If you have a `RowVectorXd` it will also work. But maybe you have a point that the matrix can be dynamic, as long as it has one dimension == 1 at runtime (but we'd need to assert that maybe?).

zero-impact:
It looks like there is already an assert on line 196:
```
assert(values.cols() == 1);
```
My vote is to make the change, if only because it took me several minutes of debugging to realize what was wrong, and this could happen to others.

PR #1163, opened 2019-04-19
Shewchuk's predicates support

qnzhou:
Add support for Shewchuk's exact predicates.

- [x] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/predicates/predicates.cpp
```diff
@@ -0,0 +1,268 @@
...
+    else if (r < 0) return Orientation::OUTSIDE;
+    else return Orientation::COSPHERICAL;
+}
+
+}
+}
+
+#ifdef IGL_STATIC_LIBRARY
+// Explicit template instantiation
+template
```
jdumas: Can you put each template instantiation on a single line?

include/igl/predicates/predicates.cpp
```diff
@@ -0,0 +1,268 @@
...
+using REAL = double;
+
+template<typename Vector2D>
+IGL_INLINE Orientation orient2d(
+        const Eigen::MatrixBase<Vector2D>& pa,
+        const Eigen::MatrixBase<Vector2D>& pb,
+        const Eigen::MatrixBase<Vector2D>& pc) {
+
+    using Point = Eigen::Matrix<REAL, Eigen::Dynamic, Eigen::Dynamic>;
+    const Point& a = pa.template cast<REAL>();
```
jdumas: Should we assert that `std::is_same<Vector2d::Scalar, REAL>::value` instead of casting the type of the argument? I wish there was a template trait for detecting narrowing conversion and avoid them, but in the meantime maybe it would be better to prevent automatic cast altogether? (Or simply allow both `float` and `double`, but not `long long double` for example?

include/igl/predicates/predicates.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+template<typename Vector3D>
+IGL_INLINE Orientation orient3d(
+        const Eigen::MatrixBase<Vector3D>& pa,
+        const Eigen::MatrixBase<Vector3D>& pb,
+        const Eigen::MatrixBase<Vector3D>& pc,
+        const Eigen::MatrixBase<Vector3D>& pd) {
+
+    using Point = Eigen::Matrix<REAL, Eigen::Dynamic, Eigen::Dynamic>;
+    const Point& a = pa.template cast<REAL>();
```
jdumas: Since you're evaluating it, wouldn't it be better to store a `Point a = ...` and avoid the `const_cast<>`?

include/igl/predicates/predicates.cpp
```diff
@@ -0,0 +1,268 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public License 
+// v. 2.0. If a copy of the MPL was not distributed with this file, You can 
+// obtain one at http://mozilla.org/MPL/2.0/.
+#include <igl/predicates/predicates.h>
+#include <predicates.h>
+
+namespace igl {
+namespace predicates {
+
+using REAL = double;
```
jdumas: I see that the `predicates.h` has a `LIBIGL_PREDICATES_USE_FLOAT` define, so maybe we need to put a static assert to make sure predicates are compiled for `double` here?

include/igl/predicates/predicates.h
```diff
@@ -0,0 +1,95 @@
...
+      COLLINEAR=0, COPLANAR=0, COCIRCULAR=0, COSPHERICAL=0, DEGENERATE=0
+    };
+
+    // Compute the orientation of the triangle formed by pa, pb, pc.
+    //
+    // Input:
+    //   pa, pb, pc  2D points.
+    //
+    // Output:
+    //   Return POSITIVE if pa, pb, pc is counterclockwise oriented.
```
hankstag: 'is' counterclockwise -> 'are' counterclockwise

include/igl/predicates/predicates.cpp
qnzhou: When `pa`'s Scalar type is the same as `REAL`, the `cast` does not do any copying.  `Point a = ...` will force an unnecessary copy.

jdumas: If you pass a matrix expression it will still allocate a temporary.

cmake/libigl.cmake
```diff
@@ -411,6 +412,19 @@ if(LIBIGL_WITH_TRIANGLE)
...
+### Compile the predicates part ###
+if(LIBIGL_WITH_PREDICATES)
+  set(PREDICATES_DIR "${LIBIGL_EXTERNAL}/predicates")
+  if(NOT TARGET predicates)
+    igl_download_predicates()
+    add_subdirectory("${PREDICATES_DIR}" "predicates")
+  endif()
+  compile_igl_module("predicates")
+  target_link_libraries(igl_predicates ${IGL_SCOPE} predicates)
+  target_include_directories(igl_predicates ${IGL_SCOPE} ${PREDICATES_DIR})
```
hankstag: could you add a compilation option here: `target_compile_definitions(igl_predicates ${IGL_SCOPE} -DLIBIGL_WITH_PREDICATES)`?

Issue #1165, opened 2019-04-20
octree bug


xarthurx:
The `igl::octree` seems to have a bug in calculating the centre of each cell. The centre of the sub-cells are located in the corner of the parent cell (red dots):

![image](https://user-images.githubusercontent.com/1921878/56456149-87f15700-6368-11e9-9c7a-353c2a8e6ac7.png)

It seems this line:
https://github.com/libigl/libigl/blob/e699d257fe860d4fd82709835e72438bd0cd1533/include/igl/octree.cpp#L109

should be 
```
          centers.emplace_back(translate_center(curr_center,h/2,i));
```
where for translating centre, the distance should be the half of width of the sub cell instead of the width `h`.

After correcting it, the result looks fine:

![image](https://user-images.githubusercontent.com/1921878/56456156-a48d8f00-6368-11e9-9acc-023e72b17783.png)

And for more complex examples:

![image](https://user-images.githubusercontent.com/1921878/56456167-d3a40080-6368-11e9-88ef-4f6fcecc7358.png)

Please let me know if there're any questions or you need a PR.

#### Check all that apply (change to `[x]`)
- [x] Windows
- [ ] Mac OS X
- [ ] Linux

See <https://libigl.github.io/CONTRIBUTING/#bugreport> for more tips.

alecjacobson:
@GavinBarill could you take a look at this?

GavinBarill:
Yes, I see the issue now in reviewing my code. I'll fix it now.

GavinBarill:
It reads better if I divide the first width by 2, so I'll change it at the beginning

GavinBarill:
@xarthurx Yes, you're right, I've reviewed the code and your proposed solution is the correct one. Would you mind making a pull request?

xarthurx:
@GavinBarill Thanks for the verification.
PR here: https://github.com/libigl/libigl/pull/1171

PR #1171, opened 2019-04-21
Fix a issue in octree.

xarthurx:
For translating centre, the distance should be the half of width of the sub cell instead of the width h.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


xarthurx:
This request is added due to https://github.com/libigl/libigl/issues/1165#issuecomment-485210639.

PR #1175, opened 2019-04-23
Solve fragmentation issue with remesh_along_isoline

n-taka:
Previous remesh_along_isoline function causes fragmentation issue (the resulting mesh gets disconnected).
This commit solves it.

- [x] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Do you have a test-case where the problem happens? If so, could you also add this as a unit test for this function?

include/igl/remesh_along_isoline.cpp
```diff
@@ -71,6 +71,7 @@ template <
   }
 
   // Loop over each face
+  std::unordered_map<int, std::unordered_map<int, int>> edgeToBirthVert;
```
BruegelN: Even though unordered_map is a C++11 feature you need to `#include <unordered_map>` especially on gcc 4.8 which is used by default in Ubuntu 14.04 (travis-ci).

n-taka: Thank you! I added #include<unordered_map> line.

n-taka:
Hi, the previous implementation causes fragmentation issue almost all time remesh happens.
To make it clear, I added simple unit test for remesh_along_isoline, which checks that the number of face connected components is unchanged.

BruegelN:
I wonder if there are any news on this PR? It's a bit impractical if a PR contains many different things. To play around with it I've cherry-picked the relevant commits into a branch of mine: https://github.com/BruegelN/libigl/tree/fix-remesh_along_isoline. There are case where the current implementation for `igl::remesh_along_isoline()` causes fragmentation and this PR resolves it.

jdumas:
I believe it would help if we were to add such test cases as a unit test.

BruegelN:
Hi @jdumas I'll have a look at this during the weekend.
Maybe @n-taka wasn't aware that his branch is not yet merged?!

jdumas:
Superseded by #1440

Issue #1126, opened 2019-03-23
Segfault during cotmatrix: tetrahedron test with CLANG


robertguetzkow:
There appears to be a bug in one of the functions called in [cotmatrix.cpp](https://github.com/libigl/libigl/blob/e699d257fe860d4fd82709835e72438bd0cd1533/tests/include/igl/cotmatrix.cpp#L97) that occasionally causes a segfault during testing with CLANG (apparently only with CLANG). Here are two build jobs that failed: 

- [Job #1737.4](https://travis-ci.org/libigl/libigl/jobs/509626667)
- [Job #1746.4](https://travis-ci.org/libigl/libigl/jobs/510265400)

#### Check all that apply (change to `[x]`)
- [ ] Windows
- [x] Mac OS X
- [ ] Linux

See <https://libigl.github.io/CONTRIBUTING/#bugreport> for more tips.

jdumas:
The fact that it fails randomly on your PR (which is not touching cotmatrix at all), and seems fixed with Clang + Eigen 3.3 indicates a possible memory/segfault issue within an Eigen operation (like working with empty matrices). Would be nice to catch it still.

robertguetzkow:
I see @alecjacobson already tried to fix it in PR #1066 which didn't seem to work.

PR #1197, opened 2019-05-14
Fix on linux path_to_executable()

BruegelN:
Added symboles to check for in path_to_executable() because capital case UNIX is not defined under Linux (i.e. Ubuntu).

Lowercase unix is defined however. See: https://stackoverflow.com/questions/142508/how-do-i-check-os-with-a-preprocessor-directive
Also added missing include for readlink().
readlink() does not append a null byte, thus use information about the written byted when building the resulting string.

- [x] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/path_to_executable.cpp
+++ b/include/igl/path_to_executable.cpp
@@ -13,6 +13,7 @@
 #  include <windows.h>
 #endif
 #include <stdint.h>
+#include <unistd.h>
 IGL_INLINE std::string igl::path_to_executable()
 {
   // http://pastebin.com/ffzzxPzi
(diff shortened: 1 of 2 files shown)


BruegelN:
```
The following tests FAILED:
	 47 - cotmatrix: tetrahedron (SEGFAULT)
Errors while running CTest
The command "ctest --verbose" exited with 8.
```
Which isn't even affected by this PR.
It seems somehow related to #1126 

See here for failed test: https://travis-ci.org/libigl/libigl/jobs/532716414#L2191

jdumas:
We will make Eigen 3.3 the default soon, when we move to the new python bindings. Our deadline for that is a SIGGRAPH course this summer, so it will happen within the next two months =)

PR #1210, opened 2019-06-03
file_dialog_open/save fixes

### LINKED ISSUE #1209: igl::file_dialog_save does not correctly handle canceling
#### Describe your issue

The header claims that "[t]he string is returned empty if no file is selected" but that does not seem correct---on Linux the function returns garbage, since buffer[] is never null-terminated in this case. I didn't test other operating systems, but at a glance the code looks dubious for those cases as well.

Also I couldn't find documentation on whether fgets() is portably safe to call with a NULL stream pointer; it doesn't crash on my compiler but makes me nervous.

- [ ] Windows
- [ ] Mac OS X
- [X] Linux
- [X] I have tried the `dev` branch and the problem persists

evouga:
Attempt to fix several issues with the libigl file dialogs, so that they correctly return an empty string rather than garbage when:
 - the user cancels the dialog
 - the user selects a file somehow longer than the 1024-character limit
 - the file selection using zenity etc. fails for any other reason.

There is a strange erasure of the last character in the Linux version
(presumably to remove some trailing newline?) that I did not touch.

Fixes #1209.

- [ ] Adds new .cpp file.
- [X] This is a minor change.

### DIFF
--- a/include/igl/file_dialog_open.cpp
+++ b/include/igl/file_dialog_open.cpp
@@ -33,7 +33,7 @@ IGL_INLINE std::string igl::file_dialog_open()
     "   end tell\n"
     "   set existing_file_path to (POSIX path of (existing_file))\n"
     "\" 2>/dev/null | tr -d '\n' ","r");
-  if( !output || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) == NULL || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) != NULL)
+  if( !output || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) == NULL || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) != NULL || ferror(output))
   {
     buffer[0] = '\0';
   }
@@ -76,7 +76,7 @@ IGL_INLINE std::string igl::file_dialog_open()
   
   // For linux use zenity
   FILE * output = popen("/usr/bin/zenity --file-selection","r");
-  if( !output || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) == NULL || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) != NULL)
+  if( !output || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) == NULL || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) != NULL || ferror(output))
   {
     buffer[0] = '\0';
   }
--- a/include/igl/file_dialog_open.cpp
+++ b/include/igl/file_dialog_open.cpp
@@ -1,9 +1,9 @@
 // This file is part of libigl, a simple c++ geometry processing library.
-// 
+//
 // Copyright (C) 2014 Daniele Panozzo <daniele.panozzo@gmail.com>
-// 
-// This Source Code Form is subject to the terms of the Mozilla Public License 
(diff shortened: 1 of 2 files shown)


include/igl/file_dialog_open.cpp
```diff
@@ -32,8 +33,9 @@ IGL_INLINE std::string igl::file_dialog_open()
     "   end tell\n"
     "   set existing_file_path to (POSIX path of (existing_file))\n"
     "\" 2>/dev/null | tr -d '\n' ","r");
-  while ( fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) != NULL )
+  if( !output || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) == NULL || fgets(buffer, FILE_DIALOG_MAX_BUFFER, output) != NULL)
```
jdumas: I'm not sure I understand the point of the check `fgets(...) == NULL || fgets(...) != NULL`? I don't understand why there was a while loop in he first place (maybe because [fgets](http://www.cplusplus.com/reference/cstdio/fgets/) stops reading upon a newline? But even then the old code looks incorrect).

Would the following fix also work in the situations you are describing?
```cpp
buffer[0] = '\0';
FILE * output = ...
if (output) {
  fgets(...);
  buffer[FILE_DIALOG_MAX_BUFFER - 1] = '\0';
}
```

evouga: There are several potential failure cases:
1. popen() fails and sets output to NULL;
2. output is empty (i.e. the user canceled the file dialog without picking a file);
3. the pipe breaks while reading the filename (so that fgets() fails and sets an error flag);
4. the chosen file is too long to fit in buffer (and hence will become invalid if naively truncated to FILE_DIALOG_MAX_BUFFER-1 characters.)
In all these cases the buffer needs to be cleared of its contents so that the function returns an empty string (per the specification in the header).

Your version handles case 1 only. Mine handles 1, 2, 4, and almost 3 (I just noticed that there is a corner case I missed: the user picks an overlong file, which is partially read by the first fgets() call, and then the pipe breaks for some reason before the second fgets() call, whose NULL return is incorrectly interpreted as an EOF. This (presumably, very rare) eventually can be detected by appending `|| ferror(output)` to the conditional expression.

jdumas: Ok I see. Cramming all those checks in a single `if()` statement makes it difficult to follow the flow. Also, with your double `fgets()` you are reading the stream multiple times, which is a bit weird.

How about the following:
```cpp
buffer[0] = '\0';
buffer[FILE_DIALOG_MAX_BUFFER - 1] = 'x'; // Initialize last character with a char != '\0'
FILE * output = ...
if (output) {
  auto ret = fgets(buffer, FILE_DIALOG_MAX_BUFFER, output);
  if (ret == NULL || ferror(output)) {
    // I/O error
    buffer[0] = '\0';
  }
  if (buffer[FILE_DIALOG_MAX_BUFFER - 1] == '\0') {
    // File name too long, buffer has been filled, so we return empty string instead
    buffer[0] = '\0';
  }
}
```

evouga: Sure that seems fine (other than disallowing names of length exactly `FILE_DIALOG_MAX_BUFFER - 1`, which may be unimportant).

jdumas: We can replace `FILE_DIALOG_MAX_BUFFER` by `1024+1` if you prefer :D

PR #1218, opened 2019-06-08
Fast winding number soups

alecjacobson:
Fast winding number for soups + new tutorial entry for FWN (soups and clouds).

Should add a unit test to this...

- [x] Adds new .cpp file.
- [ ] This is a minor change.


alecjacobson:
Seems to be including changes from https://github.com/libigl/libigl/pull/1216

alecjacobson:
Is there a way to fix this PR so it doesn't include the changes from #1216 ? I'm not sure how this happens. Maybe I created this branch starting with the wrong branch.

jdumas:
#1216 has been merged already so I don't see why this is an issue? There is always the option to rebase or squash the commit on your side if you want to keep the history clean, but this can also be done through github when you merge the PR into the `dev` branch (select "squash and merge"), which I find easier to do.

jdumas:
I've `git merge dev` from the `fast-winding-number-soups` branch, and now the diff doesn't show changes to the marching cube files. It's a bit weird, maybe github didn't refresh the diff when we merged #1216 into dev?

jdumas:
Tssk, looks like Travis updated something in their macOS image and now `ccache` is not installed by `brew`? And there is zero announcement on the Travis website/blog. Seriously?

BruegelN:
@jdumas [this line](https://github.com/BruegelN/libigl/blob/master/.travis.yml#L57) did the trick for me.
However I do not know why the documentation of travis CI states to run `brew install ccache` (https://docs.travis-ci.com/user/caching/#ccache-on-macos) but so far these lines worked fine:
```
homebrew:
    packages:
    - ccache
```
I'd assume they should be equivalent but somehow the are not 😅

jdumas:
Yeah I had the `brew install ccache` before, but I replaced it by
```
homebrew:
    packages:
    - ccache
```
at some point, after reading the Travis documentation, as I thought it was more clean. I'll try the former again if this still doesn't build.

include/igl/fast_winding_number.cpp
```diff
@@ -4,321 +4,415 @@
...
+  W.resize(Q.rows(),1);
+  igl::parallel_for(Q.rows(),[&](int p)
+    {
+    FastWindingNumber::HDK_Sample::UT_Vector3T<float>Qp;
+          Qp[0] = Q(p,0);
+          Qp[1] = Q(p,1);
+          Qp[2] = Q(p,2);
+      W(p) = fwn_bvh.ut_solid_angle.computeSolidAngle(
+        Qp,
+        accuracy_scale)
```
daviesthomas: _accuracy_scale not being used in computing solid angle. This blocks control of accuracy for this function and all that call it (triangle soup precomp+comp overload ln 334). 

Thought I'd point out simple (hopefully...) fix before merge.

alecjacobson:
Appveyor says this has failed. Looking at the dump on https://ci.appveyor.com/project/danielepanozzo/libigl-6hjk1/builds/28599834/job/evwlhl1gk05pu7d8 I can't see any errors.... What's going wrong?

jdumas:
Jobs that take > 1h are killed on AppVeyor. It's a hard limit than even paid plans are subject to (which is why I was suggesting to look into self-hosted options that AppVeyor provide, which probably won't have this limitation).

I usually re-launch the job and if we're lucky it will run on a faster VM and finish under 1h.

PR #1162, opened 2019-04-18
Change matrices and other fixes for python bindings

teseoch:
[Describe your changes and what you've already done to test it.]

#### Check all that apply (change to `[x]`)
- [x] All changes meet [libigl style-guidelines](https://libigl.github.io/style-guidelines/).
- [ ] Adds new .cpp file.
- [ ] Adds corresponding unit test.
- [ ] Adds corresponding python binding.
- [x] This is a minor change.

PR #1228, opened 2019-06-14
Python

teseoch:
New cleaned PR or https://github.com/libigl/libigl/pull/1162


cmake/LibiglDownloadExternal.cmake
```diff
@@ -64,7 +64,10 @@ function(igl_download_cork)
 endfunction()
 
 ## Eigen
-set(LIBIGL_EIGEN_VERSION 3.2.10 CACHE STRING "Default version of Eigen used by libigl.")
+# 3.2.10 8ad10ac703a78143a4062c9bda9d8fd3
+# 3.3.7  f2a417d083fe8ca4b8ed2bc613d20f07
+set(LIBIGL_EIGEN_VERSION 3.3.7                            CACHE STRING "Default version of Eigen used by libigl.")
+set(LIBIGL_EIGEN_MD5     f2a417d083fe8ca4b8ed2bc613d20f07 CACHE STRING "md5sum of the tar.gz archive corresponding to this version.")
```
jdumas: md5 is not needed anymore.

jdumas:
Ok let me add the warning message then, and then we can merge.

jdumas:
Also, it would be nice to squash those commits before merging into `dev` to cleanup the history, if that doesn't impede the development in your fork.

teseoch:
I a bit worried about the squash. If something bokes we cannot trace it back

danielepanozzo:
Yes, I think we should not squash this PR. @alecjacobson could you please also approve (or directly merge) this PR? Since it is massive I want to be sure that we are all ok before proceeding.

include/igl/arap_linear_block.h
```diff
@@ -43,32 +43,32 @@ namespace igl
...
-    Eigen::SparseMatrix<Scalar> & Kd);
+    MatK & Kd);
   // Helper functions for each energy type
-  template <typename MatV, typename MatF, typename Scalar>
+  template <typename MatV, typename MatF, typename MatK>
   IGL_INLINE void arap_linear_block_spokes(
     const MatV & V,
     const MatF & F,
     const int d,
-    Eigen::SparseMatrix<Scalar> & Kd);
```
alecjacobson: How come this can't be `SparseCompressedBase` like `arap_rhs`? Switching a templated to a class to a generic template is a bad practice in general.

include/igl/boundary_loop.cpp
```diff
@@ -142,7 +142,7 @@ IGL_INLINE void igl::boundary_loop(
   vector<int> Lvec;
   boundary_loop(F,Lvec);
 
-  L.resize(Lvec.size());
```
alecjacobson: This will fail for RowVector (but that's not very popular so perhaps it's ok if it's a necessary change; not sure how it is).

include/igl/ears.cpp
```diff
@@ -15,15 +15,15 @@ IGL_INLINE void igl::ears(
   Eigen::PlainObjectBase<Derivedear_opp> & ear_opp)
 {
   assert(F.cols() == 3 && "F should contain triangles");
-  Eigen::Array<bool,Eigen::Dynamic,3> B;
```
alecjacobson: I don't understand this change. `Array`s are fine...

include/igl/invert_diag.h
```diff
@@ -22,10 +22,10 @@ namespace igl
   //   X  an m by n sparse matrix
   // Outputs:
   //   Y  an m by n sparse matrix
-  template <typename T>
+  template <typename DerivedX, typename MatY>
   IGL_INLINE void invert_diag(
-    const Eigen::SparseMatrix<T>& X, 
-    Eigen::SparseMatrix<T>& Y);
```
alecjacobson: Same as above. It's bad practice to simply make things generic templates.

include/igl/scaf.cpp
```diff
@@ -435,8 +435,8 @@ void build_surface_linear_system(const SCAFData &s, Eigen::SparseMatrix<double>
   {
     MatrixXd bnd_pos;
     igl::slice(s.w_uv, bnd_ids, 1, bnd_pos);
-    ArrayXi known_ids(bnd_ids.size() * dim);
```
alecjacobson: Why this change?

include/igl/slice_into.cpp
```diff
@@ -34,7 +34,7 @@ IGL_INLINE void igl::slice_into(
 #endif
 
   // create temporary dynamic sparse matrix
-  Eigen::DynamicSparseMatrix<T, Eigen::RowMajor>  dyn_Y(Y);
```
alecjacobson: This is an incorrect change. Was it tested? I guess it will be significantly slower. `DynamicSparseMatrix` is an obsolete class that had much better performance. Changing this to `SparseMatrix` doesn't make sense with the current code and introduces _at least_ a copy and possible a lot of overhead.

jdumas: Coming back to #801, I feel we should really try to get rid of this `DynamicSparseMatrix` at some point. I haven't had time to come up with a benchmark test case however, but I can try to give it a shot next week.

include/igl/arap_linear_block.h
teseoch: > no member named 'resize' in 'Eigen::SparseCompressedBase<Eigen::SparseMatrix<double, 0, int> >'

include/igl/boundary_loop.cpp
teseoch: it dosent compile otherwise

include/igl/ears.cpp
teseoch: it created many problems, i dont remember exactly anymore which ones.

It was something with incompatibility with MatrixBase

include/igl/invert_diag.h
teseoch: > no member named 'coeffRef' in 'Eigen::SparseCompressedBase<Eigen::SparseMatrix<double, 0, int> >'

include/igl/scaf.cpp
teseoch: i dont remember

include/igl/slice_into.cpp
teseoch: ok i revert that change. I changed it because it was deprecated...

teseoch:
In general we cannot use `Eigen:: SparseCompressedBase` for output. it lacks most of the write functions. That's why I use a generic template

include/igl/ears.cpp
jdumas: Probably some templating issues with the `Array<>` type being propagated to an function that expects a `MatrixBase<>` as input, which is not possible because of inheritance.

include/igl/boundary_loop.cpp
jdumas: Probably doesn't compile because forwarding from a numpy-allocated array maybe the number of colums at compile time is not always known to be 1 (can be 0 and needs to be allocated). In any case, since the convention in libigl is to use column vectors, I would argue that passing a RowVector to a function returning a N x 1 matrix should be illegal anyway. No need to add an assert since the `resize()` will complain if the user does so.

jdumas:
Apart from the issue of `Eigen::DynamicMatrix` that has been reverted, I don't think see any big issue left on this PR. It may be better to merge it and quickly unforeseen issues as they arise that to merge it right before the SIGGRAPH course and find out there are still minor issues to iron out.

alecjacobson:
It'd be great to understand the issue with Eigen::Array before making this
change.

jdumas:
My understanding is that `Array` is not a descendant of `MatrixBase`, so you cannot use it in combination with functions that expect a `MatrixBase` as input (see the [Eigen class hierarchy](https://eigen.tuxfamily.org/dox/TopicClassHierarchy.html)). For example in `ears.cpp`, the matrix `B` is given to `igl::slice()`, whose input was changed from a `DenseBase<>` to a `MatrixBase`.

Now, maybe you could argue that `igl::slice()` should still work on `DenseBase<>` types, and I don't know why those were changed to `MatrixBase<>`. However it seems more clean to use the same calling convention for all libigl functions, i.e. `MatrixBase<>` for inputs and `PlainObjectBase<>` for outputs.

jdumas:
@alecjacobson is the `Array<>` change still an issue for you, or do you think we can merge this PR?
As we uniformize input types of libigl funfctions, we should decide to either stick with `MatrixBase<>` or `DenseBase<>` as input, but having a mix would be kinda confusing.

include/igl/unproject.cpp
```diff
@@ -38,13 +38,13 @@ IGL_INLINE void igl::unproject(
...
     typedef typename Derivedscene::Scalar Scalar;
-    Eigen::Matrix<Scalar,4,4> Inverse = 
+    Eigen::Matrix<Scalar,4,4> Inverse =
       (proj.template cast<Scalar>() * model.template cast<Scalar>()).inverse();
 
     Eigen::Matrix<Scalar,4,1> tmp;
     tmp << win.row(i).head(3).transpose(), 1;
-    tmp(0) = (tmp(0) - viewport(0)) / viewport(2);
-    tmp(1) = (tmp(1) - viewport(1)) / viewport(3);
+    tmp(0) = (tmp(0) - viewport(0, 0)) / viewport(2, 0);
```
jdumas: `viewport` is supposed to be a column vector. Why was this change needed?

include/igl/unproject_ray.cpp
```diff
@@ -27,8 +27,8 @@ IGL_INLINE void igl::unproject_ray(
   using namespace Eigen;
   // Source and direction on screen
   typedef Eigen::Matrix<typename Deriveds::Scalar,3,1> Vec3;
-  Vec3 win_s(pos(0),pos(1),0);
-  Vec3 win_d(pos(0),pos(1),1);
+  Vec3 win_s(pos(0, 0),pos(1, 0),0);
```
jdumas: Same here, `pos` is a column-vector, so it should be possible to use the single-index acessors.

include/igl/isolines.cpp
```diff
@@ -38,29 +38,29 @@ IGL_INLINE void igl::isolines(
...
         iso(i) = Scalar(i)/Scalar(n)*(max-min) + min;
-    
+
     typedef Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic> Matrix;
     std::array<Matrix,3> t{{Matrix(nFaces, np1),
         Matrix(nFaces, np1), Matrix(nFaces, np1)}};
     for(int i=0; i<nFaces; ++i) {
         for(int k=0; k<3; ++k) {
-            const Scalar z1=z(F(i,k)), z2=z(F(i,(k+1)%3));
+            const Scalar z1=z(F(i,k), 0), z2=z(F(i,(k+1)%3), 0);
```
jdumas: `z` is also `V x 1`, so should be no need to add `, 0)`.

teseoch: if it comes from python it is not. we already had this discussion

Issue #1198, opened 2019-05-17
Cross platform serialization


jhwang7628:
I have some class/data that I want to serialize on a Linux machine (64-bit machine with little endian byte order) and deserialize on a Mac machine (64-bit machine I think also with little endian). I found that the data in std::vector is somehow corrupted. Any idea why? 

The simplest way for me to reproduce this is to use the provided example:

    bool b;
    unsigned int num;
    std::vector<float> vec;

    // use overwrite = true for the first serialization to create or overwrite an
    // existing file
    igl::serialize(b,"B","filename",true);
    // append following serialization to existing file
    igl::serialize(num,"Number","filename");
    igl::serialize(vec,"VectorName","filename");

    // deserialize back to variables
    igl::deserialize(b,"B","filename");
    igl::deserialize(num,"Number","filename");
    igl::deserialize(vec,"VectorName","filename");

In this cases vec deserialized on my Mac will have a zero size.

I know in the documentation it says the serialization does not support cross platform, but in my case I thought it should work. Not sure if its a bug.

jdumas:
Can you post the content of the serialized file when dumped on both machines? You can also inspect by stepping into [this function](https://github.com/libigl/libigl/blob/e33f534ffd2c04b5c82e9f8de432ddb61e2e2d65/include/igl/serialize.h#L526) during deserialization, to see what's going on (what are the size/type names, etc. that it is dumping).

I ran into an issue recently because type names are serialized *mangled*, so if you use different compilers the type for `std::vector` might be mangled in different ways, so the deserialization part will "skip" over the entry that it should read from the binary file. My issue is even worse than that since it involves `std::string`, which have different aliases (`std::__cxx11::basic_string<...>`, etc.) depending on which C++ ABI you compile with ...

In the meantime you can try another serialization library like [this one](https://github.com/USCiLab/cereal). This probably won't solve ambiguity if you are serializing a `std::string`, but it does demangle type names before serializing them.

jdumas:
Ok so for reference we've found that on clang, stl types are under the namespace `std::__1`, so we need to replace all occurrences of `std::__1` by `std` when serializing/deserializing for this to work. See this [SO thread](https://stackoverflow.com/questions/25103885/how-to-demangle-stdstring-as-stdstring) for additional details.

[...]

PR #1233, opened 2019-06-20
cross platform serialization

BruegelN:
As mentioned in #1198 
Removed the `replaceString()` in `demangle.h`.
Check if type you want to demangle std::is_same as std::string.
Added minimal unit test for serialize.
I'm aware that `Viewer::save_scene()` and `Viewer::load_scene()` use serialization but I'm not sure how to test this headless? 

It might be nice to change demangleName to:
``` C++
template<class T> inline
std::string demangledName( const T& var)
{

}
```
Because then template specification can makes things easier or added when needed e.g. like:
``` C++
template< > inline
std::string demangledName( const std::string& var)
{
    return std::string("std::string");
}

template<class T> inline
std::string demangledName( const std::vector<T>& var)
{
    return std::string("std::vector<"+demangle(typeid(T).name()))+">";
}
```

@jdumas could you please let me know what you think? Any further sugesstion or did I miss anything?

The two serialized files (used in `tests/include/igl/serizalize.cpp`) need to be added to libigl-test-data before merging this PR. Then I can change back the repo url to `https://github.com/libigl/libigl-tests-data`
I'll add the corresponding PR to libigl-test-data soon if no rework regarding the test is needed.

- [x] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
Hmm, I've thought about changing the type of `demangle<>()` to use a recursive template to remove the ambiguity for nested types (e.g. a `std::vector<std::string>`). The problem is that it will not work with user types without extra definitions by the user (e.g. a `Eigen::Matrix`, user-templated types, etc.). Thus I'm a bit hesitant to add this as is, as it would require a more careful refactoring of the interface.

Maybe @danielepanozzo would have some design suggestions, but for now I would say to leave `std::string` as they are, even if it means dropping cross-platform compatibility (we can add an entry in the documentation about that).

include/igl/demangle.h
```diff
@@ -0,0 +1,78 @@
...
+
+namespace igl
+{
+
+#ifdef _MSC_VER
+  //! Demangles the type encoded in a string
+  /*! @internal */
+  inline std::string demangle(std::string mangledName)
+  {
+    std::string prefix = "class "; // MSVC does "class myClass" see: https://docs.microsoft.com/en-us/cpp/cpp/typeid-operator
```
jdumas: Not sure if this works with nested class names. I would use an auxiliary function to replace all instances of one string by another, e.g. [this](https://stackoverflow.com/questions/5343190/how-do-i-replace-all-instances-of-a-string-with-another-string).

include/igl/demangle.h
```diff
@@ -0,0 +1,78 @@
...
+  {
+    int status = 0;
+    char *demangledName = nullptr;
+    std::size_t len;
+
+    demangledName = abi::__cxa_demangle(mangledName.c_str(), 0, &len, &status);
+
+    std::string retName(demangledName);
+    free(demangledName);
+	  // MSVC does not add spaces after semicolon in gcc and clang sth. like this: "class<T1, T2, T3>"
```
jdumas: You are mixing tabs and space for indentation. Also, if the problem is space after comma, then we should replace `", "` by `","` (using the same `replace_all()` function defined above).

include/igl/demangle.h
```diff
@@ -0,0 +1,78 @@
...
+    retName.erase(end_pos, retName.end());
+    return retName;
+  }
+
+#endif // clang or gcc branch of _MSC_VER
+
+  //! Gets the demangled name of a type
+  /*! @internal */
+  template<class T> inline
+  std::string demangledName()
```
jdumas: I know I've used `demandedName()` in my fork, but the naming convention of libigl is `lower_case`, so we should update the name.

include/igl/demangle.h
```diff
@@ -0,0 +1,78 @@
...
+  {
+    if (std::is_same<T, std::string>::value)
```
jdumas: This will only work if the primary type is `std::string`, not if there are nested types (e.g. `std::vector<std::string>`). I don't have a good solution for that yet.

include/igl/demangle.h
BruegelN: Thanks! done with 8d286fe3e3d4ae17dd4a01104b2deeb31be7aeae

include/igl/demangle.h
BruegelN: Sorry for the indentation mixture. (Don't use my VS to often, it has to be somehow miss configured....)
With 8d286fe3e3d4ae17dd4a01104b2deeb31be7aeae `replace_all()` is used

include/igl/demangle.h
BruegelN: Fixed with b5707cf795eb5447705523587b848e5b786ce3ba

include/igl/demangle.h
BruegelN: Yes your right, currently it doesn't work.
I've an idea in my mind using RTTI to check for `std::vector` as well.
Are there any other mandatory datatype I'm missing out?

jdumas: Yes there could be a `std::array<std::string, N>`, a `Eigen::Matrix<std::string>`, or even a user-defined templated function that takes a `std::string` as template parameter. Unlikely for sure, but I think it's a bit weird to have different behaviour for `std::vector` and for other types of data.

BruegelN:
I'm so sorry, I totally forgot about this one. :disappointed:
Thank you @hankstag for your improvements on the testcase! Is there something left to do?

include/igl/serialize.h
```diff
@@ -105,21 +106,21 @@ namespace igl
   //   obj        object to load back serialization to
   //
   template <typename T>
-  inline bool deserialize(T& obj,const std::string& filename);
+  inline bool deserialize(T& obj,const std::string& filename,const bool warning = false);
```
jdumas: I thought we should default to having `warning = true`? Also, no need to make it `const bool` (`bool` is enough).

jdumas:
@danielepanozzo I feel there should also be a way to programmatically check if the types match. Right now the return type is just a bool success/failure. Maybe we should make it an enum or something to indicate mismatched types/error, etc.?

@hankstag Can you also update the documentation (comments) about the extra `warning` flag?

jdumas:
@danielepanozzo @hankstag I think I have another idea regarding the type mismatch. There should be a way to programmatically check if the type matches or not, as well as show the warning on stderr. So how about we pass a `int flag` instead of a simple `bool warning`? The bit flag would be able to use the following enum:

```cpp
enum Serialization : int {
   SHOW_WARNING = 1,
   ERROR_ON_MISMATCH = 2,
}
```

We can find better names for the flag, but by default the argument would be `flag = SHOW_WARNING | ERROR_ON_MISMATCH`, and the user is responsible for calling deserialization with other options.

@hankstag for the unit test, can you produce a file serialized under Windows/macOS, and Linux (both on the HPC + a recent Ubuntu), that contains both a `std::string` and a `std::vector<std::string>`? I have modified the demangling function to handle (I think) nested string in type names. Thus we should be able to deserialize those 4 files with the flag `ERROR_ON_MISMATCH` and assert for success in the unit test.

hankstag:
@BruegelN No worries, I can take care of the rest

hankstag:
> We can find better names for the flag, but by default the argument would be `flag = SHOW_WARNING | ERROR_ON_MISMATCH`, and the user is responsible for calling deserialization with other options. If `ERROR_ON_MISMATCH` is on, we retain the old behavior (but still demangle type names).
@jdumas Could you elaborate more on the `old behavior` - what do we do when there is a mismatch and `ERROR_ON_MISMATCH` is on? Do we throw a runtime error?

jdumas:
The function returns a boolean indicating success, so if the flag ERROR_ON_MISMATCH is set, deserialization should stop and return false when types do not match.

hankstag:
i see. I will make the change and prepare the unit tests

jdumas:
Awesome, thanks!

hankstag:
@jdumas i tried to load a serialized string and a vector of strings from windows - however the types still seem to be inconsistent and the check failed. 
`inconsistent types <<< std::basic_string<char,struct std::char_traits<char>,std::allocator<char> >
                                   >>> std::string`
I'm not sure how the demangle works, could you check? The files for the test case are here

[serialized_files.zip](https://github.com/libigl/libigl/files/3828027/serialized_files.zip)

jdumas:
@hankstag I think I've pushed a fix regarding the demangling of std::string on Windows. The problem is that when you as `typeid(T).name()` to the compiler you don't always get `"T"` as the answer. This applies to Windows/Linux and macOS. Let me know if my latest changes fixes the serialization issues for Windows as (you will need to reserialize).

alecjacobson:
I don't use the serialize functions so my heads not really in this implementation. Is this PR close to being ready to merge? If not, what's the remaining work?

alecjacobson:

PR #1235, opened 2019-06-21
Added ply reader based in tinyply

jmespadero:

- [x] Adds new .cpp file.
- [ ] This is a minor change.

I have not done python binding, but suppose that the old binding is working because I have just reimplemented some methods, no new methods have been created.


jdumas:
Nice! I think there's no point in keeping the old code in comments though, you can just delete it.

jmespadero:
It was just to ensure it has the same functionality. I will delete soon.

include/igl/tinyply.cpp
```diff
@@ -0,0 +1,720 @@
...
+//   tinyply implementation   //
+////////////////////////////////
+
+#include <algorithm>
+#include <functional>
+#include <type_traits>
+#include <iostream>
+#include <cstring>
+
+namespace igl::tinyply {
```
jdumas: I think this form of nested namespace is only available in C++17.
Also, there should be no need for a `using namespace igl::tinyply` just below.

include/igl/read_triangle_mesh.h
```diff
@@ -29,20 +29,20 @@ namespace igl
...
+  //   path  path to file
   // Outputs:
   //   V  eigen double matrix #V by 3
   //   F  eigen int matrix #F by 3
-  // Returns true iff success
+  // Returns true if success
   template <typename Scalar, typename Index>
   IGL_INLINE bool read_triangle_mesh(
-    const std::string str,
+    const std::string path,
```
jdumas: Maybe we should also change those to `const std::string &` while we're at it.

include/igl/readPLY.cpp
```diff
@@ -16,20 +17,20 @@ template <
...
-  // Largely follows ply2iv.c
-  FILE * ply_file = fopen(filename.c_str(),"r");
-  if(ply_file == NULL)
-  {
+  try {
+    std::ifstream ply_stream(filename, std::ios::binary);
+    if (ply_stream.fail())
+          return false;
+    return readPLY(ply_stream,V,F,N,UV);
+  } catch (...) {
```
jdumas: Maybe we should catch `std::exception()` and display the `.what()` on `stderr`. Using `(...)` can catch stuff like `throw 2;` which is a bit weird.

include/igl/readPLY.cpp
```diff
@@ -38,140 +39,313 @@ template <
...
+  // Tinyply treats parsed data as untyped byte buffers.
+  std::shared_ptr<igl::tinyply::PlyData> vertices, normals, faces, texcoords;
+
+  // The header information can be used to programmatically extract properties on elements
+  // known to exist in the header prior to reading the data. For brevity of this sample, properties
+  // like vertex position are hard-coded:  
+  try {
+    vertices = file.request_properties_from_element("vertex", { "x", "y", "z" });
   }
+  catch (const std::exception & e) { }
```
jdumas: Shouldn't we return `false` if an exception occurs?

jmespadero: This is the suggested way to ask if a .ply file contains (or not) a property associate to vertex. 
In this case, we could argument that if a file does not contain (x,y,z) properties in vertex, then you can not really store a mesh. Also, somebody could argument that "an empty mesh is still a mesh", so I would let that code as is.

jdumas: Ah, I see. I don't really like that it uses exceptions to signal the absence of optional attributes, but I guess that's how it is.

jmespadero:
I have solved all the template instantiation errors using the instructions in https://libigl.github.io/static-library/#explicit-instantiations-of-templated-functions , but I have a new problem that I can't solve.

I have written a test case for readPLY that reads a file named "bunny.ply" and check the results. The test is working on my machine but not in travis, which launch this error:

```
106: Test command: /home/travis/build/libigl/libigl/build/libigl_tests "readPLY: bunny.ply"
106: Test timeout computed to be: 1500
106: ReadPLY: Error opening file /home/travis/build/libigl/libigl/cmake/../external/../tests/data/bunny.ply
106: 
```

I can't understand why, because the file `bunny.ply`has been correctly pushed to ' /tests/data/bunny.ply'

BruegelN:
@jmespadero the folder `libigl/tests/data` is another git repo (https://github.com/libigl/libigl-tests-data).
This does not contain a `bunny.ply` (nor `mesh_error.ply`).
It is automatically downloaded when running cmake (see: https://github.com/libigl/libigl/blob/dev/cmake/LibiglDownloadExternal.cmake#L173)
To add your files you might want to fork the libigl-tests-data repo and submit a PR there with your bunny.ply(+mesh_error.ply) and update the `GIT_TAG` for the repo afterwards.

jmespadero:
Thanks!

jdumas:
@jmespadero can you add the `.ply` test files in the `libigl-test-repo`, create a PR on this repo, and update this PR accordingly to point to the commit containing those, in order to pass the CI?

ddiakopoulos:
@jdumas @jmespadero It would be super neat to see this merged in! Let me know if there's anything I can do to help. I've also updated tinyply to version 2.3.1 which has some performance improvements.

vfonov:
Oops , looks like I forgot to include right files in commit

jdumas:
Merged with #1422

PR #1237, opened 2019-06-24
fix #1230: Heat geodesics example crashes

### LINKED ISSUE #1230: Heat geodesics example crashes
#### Describe your issue
Running the tutorial `716_HeatGeodesics_bin` on the default file `beetle.off` crashes with a `vector subscript out of range` error. 

It seems to be triggered by `slice()` in `min_quad_with_fixed.cpp` while running the function `heat_geodesics_precompute()`.

- [X] Windows
- [ ] Mac OS X
- [ ] Linux
- [X] I have tried the `dev` branch and the problem persists

xinyazhang:

- [ ] Adds new .cpp file.
- [x] This is a minor change.

This change swap the calling of transpose() and sparseView() and fixed the igl::heat_geodesics_precompute on Windows.

tutorial/716_HeatGeodesics has been tested on Windows with `cmake -DLIBIGL_USE_STATIC_LIBRARY=OFF ..`

P.S. I'm not an expert of Eigen so I've no idea this comes from misusing of Eigen, a Windows bug in Eigen, or a compiler bug from Visual C++.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,14 +41,17 @@ matrix:
     env:
     - MATRIX_EVAL="export CC=gcc-7 CXX=g++-7 CONFIG=Release PYTHON=python3 CMAKE_EXTRA='-DLIBIGL_EIGEN_VERSION=3.3.7'"
   - os: osx
+    osx_image: xcode10.2
     compiler: clang
     env:
     - MATRIX_EVAL="export CONFIG=Debug PYTHON=python3 LIBIGL_NUM_THREADS=1"
   - os: osx # same config like above but with -DLIBIGL_USE_STATIC_LIBRARY=OFF to test static and header-only builds
+    osx_image: xcode10.2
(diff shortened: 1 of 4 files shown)


alecjacobson:
Thanks for doing the hard working of tracking down this issue. Looks like a great candidate to submit as a possible but to Eigen, too.

jdumas:
Hmm why are we even transposing a diagonal sparse matrix in the first place?

alecjacobson:
We're not.

`M.diagonal()` returns the digaonal of M as a (dense) column vector.

`Aeq` needs this as a sparse row vector.

jdumas:
Ah, right, I got confused with `.asDiagonal()`. So it looks like one of the typical temporary evaluation with Eigen then, and should be reported.

jdumas:
Fixed by #1288

PR #1202, opened 2019-05-20
Hunterize libigl

rbsheth:
This PR adds basic support for the Hunter package manager in libigl. Because libigl is already following modern CMake practices, the changes needed to make this package Hunter-ready are minimal.

Changes:
1) Add `HunterGate.cmake` and corresponding CMake code in `CMakeLists.txt`
2) Pull `Eigen` from Hunter
3) Add `HUNTER_ENABLED` flag and default to `OFF` to preserve current behavior.

Once this change is merged, a new (minor?) GitHub release is needed to generate a source tarball that Hunter can pull in.

#### Check all that apply (change to `[x]`)
- [x] All changes meet [libigl style-guidelines](https://libigl.github.io/style-guidelines/).
- [ ] Adds new .cpp file.
- [ ] Adds corresponding unit test.
- [ ] Adds corresponding python binding.
- [x] This is a minor change.

PR #1242, opened 2019-06-25
Hunterize libigl, attempt 2

rbsheth:
Original PR with missing branch here: #1202 

This PR adds basic support for the Hunter package manager in libigl. Because libigl is already following modern CMake practices, the changes needed to make this package Hunter-ready are minimal.

Changes:
1) Add `HunterGate.cmake` and corresponding CMake code in `CMakeLists.txt`
2) Pull `Eigen` from Hunter
3) Add `HUNTER_ENABLED` flag and default to `OFF` to preserve current behavior.

Once this change is merged, a new (minor?) GitHub release is needed to generate a source tarball that Hunter can pull in.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/cmake/libigl.cmake
+++ b/cmake/libigl.cmake
@@ -90,9 +90,14 @@ if(BUILD_SHARED_LIBS)
   set_target_properties(igl_common PROPERTIES INTERFACE_POSITION_INDEPENDENT_CODE ON)
 endif()
 
-if(UNIX)
+if(UNIX AND NOT HUNTER_ENABLED)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
+endif()
+
+if(HUNTER_ENABLED)
+  hunter_add_package(Eigen)
+  find_package(Eigen3 CONFIG REQUIRED)
 endif()
 
 # Eigen
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,14 +41,17 @@ matrix:
     env:
     - MATRIX_EVAL="export CC=gcc-7 CXX=g++-7 CONFIG=Release PYTHON=python3 CMAKE_EXTRA='-DLIBIGL_EIGEN_VERSION=3.3.7'"
   - os: osx
+    osx_image: xcode10.2
     compiler: clang
     env:
     - MATRIX_EVAL="export CONFIG=Debug PYTHON=python3 LIBIGL_NUM_THREADS=1"
   - os: osx # same config like above but with -DLIBIGL_USE_STATIC_LIBRARY=OFF to test static and header-only builds
+    osx_image: xcode10.2
     compiler: clang
     env:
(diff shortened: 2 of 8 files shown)


cmake/libigl.cmake
```diff
@@ -80,6 +80,10 @@ if(MSVC)
   # Enable parallel compilation for Visual Studio
   target_compile_options(igl_common INTERFACE /MP /bigobj)
   target_compile_definitions(igl_common INTERFACE -DNOMINMAX)
+  target_compile_options(igl_common INTERFACE /Zc:__cplusplus)
```
jdumas: Can you explain why this is needed? Maybe it's worth discussing in a separate PR.

cmake/libigl.cmake
```diff
@@ -80,6 +80,10 @@ if(MSVC)
...
+  if(LIBIGL_USE_STATIC_LIBRARY)
+    target_compile_definitions(igl_common INTERFACE -D_USE_MATH_DEFINES)
```
jdumas: This shouldn't be necessary, as we take care to not use `M_PI` directly in libigl. Is there any reason to enable this flag?

cmake/libigl.cmake
rbsheth: Background on the flag: https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

I could swear there was some code gated on the `__cplusplus` version (which I subsequently used here for porting `master`: https://github.com/hunter-packages/libigl/commit/5f2a02d3eec78b94b701907f66ffbaafd0f6d0a1#diff-379356ad4dff6b985fbb126dc1a6fd32R11) but I can't find it. I can remove it for this PR.

cmake/libigl.cmake
rbsheth: Same issue, perhaps it was fixed earlier but in my experiment `M_PI` had crept in somehow. I can remove this too.

jdumas: Yes, we should instead remove all usage of `M_PI` from libigl's code instead, without requiring additional compilation flags from the user.

cmake/libigl.cmake
jdumas: Libigl's minimum requirement in terms of C++ version is C++11. We don't seem to be relying on the `__cplusplus` macro to induce some specific behavior, so I don't see the hurt in not specifying this flag.

rbsheth:
@jdumas Any ideas on this Travis failure? https://travis-ci.org/libigl/libigl/jobs/550530668

Looks like `ccache` is missing?

jdumas:
Yes. I'm waiting for another maintainer to approve #1239 (I can't merge it myself) which fixes the CI issues. Once this is done we can merge this PR.

jdumas:
@rbsheth I'm trying to merge the tip of the `dev` branch with your PR branch to trigger a new build on Travis, but it doesn't look like you've given us write access to the branch when you open your PR. Could you see to it?

rbsheth:
@jdumas Invited you to collaborate

jdumas:
@rbsheth do you mind if I squash and merge instead of just merging? Or will it affect your fork in hunter for future versions?

rbsheth:
@jdumas Squash merge should be fine, I can work with that for Hunter.

PR #1247, opened 2019-07-04
Fix igl::is_irregular_vertex()

BruegelN:
Only in DEBUG mode Eigen does all boundary checks when accessing elements.
Thus when using `igl::is_irregular_vertex()` fails if `CMAKE_BUILD_TYPE=Debug`.
I've added a test for igl::is_irregular_vertex to showcase that it segfaults with CMAKE_BUILD_TYPE=Debug due to a missing F.maxCoeff()+1 in https://github.com/libigl/libigl/blob/master/include/igl/is_irregular_vertex.cpp#L16 see build https://travis-ci.org/BruegelN/libigl/builds/550519463 (Note only macOS builds with `CMAKE_BUILD_TYPE=Debug` and it's not related to the recent changes on Travis CI for macOS)
After fixing the off-by-one issue everything works as expected (https://travis-ci.org/BruegelN/libigl/builds/550521102)

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/tests/include/igl/is_irregular_vertex.cpp
+++ b/tests/include/igl/is_irregular_vertex.cpp
@@ -6,16 +6,10 @@ TEST_CASE("is_irregular_vertex: simple", "[igl]")
 {
     Eigen::MatrixXd V;
     Eigen::MatrixXi F;
-    // Known non-manifold mesh
+    // Known "bad" mesh (many boundaries + irregular vertices, non-manifold)
     test_common::load_mesh("truck.obj", V, F);
     std::vector<bool> vec = igl::is_irregular_vertex(V,F);
-    int ret = false;
-    for (auto val : vec)
-    {
-        // true corresponds to 1
-        ret += val;
-    }
-    // some verticies are irregular thus the sum over all verticies should evaluate to a value > 0
-    REQUIRE( ret > 0 );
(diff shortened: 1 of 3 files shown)


tests/include/igl/is_irregular_vertex.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+
+TEST_CASE("is_irregular_vertex: simple", "[igl]")
+{
+    Eigen::MatrixXd V;
+    Eigen::MatrixXi F;
+    // Known non-manifold mesh
+    test_common::load_mesh("truck.obj", V, F);
+    std::vector<bool> vec = igl::is_irregular_vertex(V,F);
+    int ret = false;
```
jdumas: Note that you can use `std::any_of` to write this check in 1 line.

tests/include/igl/is_irregular_vertex.cpp
```diff
@@ -0,0 +1,21 @@
...
+    for (auto val : vec)
+    {
+        // true corresponds to 1
+        ret += val;
+    }
+    // some verticies are irregular thus the sum over all verticies should evaluate to a value > 0
```
jdumas: `vertices`

PR #1259, opened 2019-07-25
Path to edges + various template instantiations

lawsonfulton:
I've added a new function path_to_edges which acts as the converse of the existing igl::edges_to_path.

- [x] Adds new .cpp file.
- [x?] This is a minor change.


lawsonfulton:
Ok looks like all the checks are passing now.

PR #1263, opened 2019-08-01
Fix a potential crash when initializing the map for meshes with holes

martin-marinov77:
* Fix a potential crash when initializing the map for meshes with holes
* Fix a minor conversion warning

Without this change we can get crashes on models with holes.

### DIFF
--- a/tutorial/710_SCAF/main.cpp
+++ b/tutorial/710_SCAF/main.cpp
@@ -98,9 +98,8 @@ int main(int argc, char *argv[])
     all_bnds.erase(primary_bnd);
     Eigen::MatrixXi F_filled;
     igl::topological_hole_fill(F, bnd, all_bnds, F_filled);
-    igl::harmonic(F_filled, bnd, bnd_uv, 1, uv_init);
-    Eigen::MatrixXd tmpr_uv_init = uv_init.topRows(V.rows()); // copy to a temporary
-    uv_init = std::move(tmpr_uv_init); // and move over to prevent another copy
+    igl::harmonic(F_filled, bnd, bnd_uv ,1, uv_init);
+    uv_init.conservativeResize(V.rows(), 2);
   }
 
   Eigen::VectorXi b; Eigen::MatrixXd bc;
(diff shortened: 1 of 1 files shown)


jiangzhongshi:
@jdumas After changing the base branch, some more commits are appended. Is there an easy way from our side to dismiss them?

martin-marinov77:
I can rebase this on top of dev and resubmit if that makes your life easier?
I did the change on master only because it's a crash fix and people consuming libigl might want to get this immediately rather than wait for dev to get merged into master...

jdumas:
If people want the fix they should use the dev branch. You can rebase onto dev and force-push into this PR, no need to create a new PR.

martin-marinov77:
OK, done.

jiangzhongshi:
Hi Martin,

Thanks for the PR, but I don't quite understand why it would potentially crash, at least on my desktop. Maybe I missed something, can you point it out?
I admit this is not the best way to write it. `uv_init.conservativeResize(V.rows(), 2)` could be a more succinct way.

martin-marinov77:
This did crash in Eigen on some of the _large_ models we've tried, and corrupted the init uv data in other cases. I did not dig into it too deep, but I did see Eigen doing some weird stuff internally in `uv_init = uv_init.topRows(V.rows())`. I cannot share the test models, but maybe i can get a call-stack to attach to this to convince you. The resize approach might work as well, I did not think of it.

Bottom line is that after I've replaced `uv_init = uv_init.topRows(V.rows())` with the statements in the commit, everything started to work...

martin-marinov77:
Here you go, this is what I get with a large mesh with many holes and a very large outer boundary. 

image

I will not be surprised if Eigen somehow decides to free the existing memory in `uv_init` in the assignment operator and allocate new storage before the copy, possibly to save some memory. Then the wrapper of `topRows()` would be reading discarded memory which may (or may not) work.

martin-marinov77:
`uv_init.conservativeResize(V.rows(), 2)` is better indeed, it does the same (free, alloc and copy) but it's more concise. So I've pushed another commit with the change. Thanks for tip.

PR #1277, opened 2019-08-19
avoid nan in barycentric_coordinates

ShuangLiu1992:
avoid nan in barycentric_coordinates

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
we don't have a consistent policy for throwing/returning errors. For functions like this, returning nan is easily detectable by the caller. So I'd opt to keep it as is.

jdumas:
So just close this PR then§

Issue #1276, opened 2019-08-19
Keep mtl reference in OBJReader


DerouineauNicolas:
Hi,
From what I see here https://github.com/libigl/libigl/blob/f6b406427400ed7ddb56cfc2577b6af571827c8c/include/igl/readOBJ.cpp#L209, reference to faces material are ignored when reading obj. I think there could be some usecase (for example mesh reduction with uv) where reading materials/faces association would be useful (maybe with a std::map ?). Does it make sense ? 

#### Describe your issue

...

#### Check all that apply (change to `[x]`)
- [ ] Windows
- [ ] Mac OS X
- [ ] Linux
- [ ] I have tried the `dev` branch and the problem persists

See https://libigl.github.io/CONTRIBUTING/#bugreport for more tips.

jdumas:
Well, pull-requests are welcome =)

PR #1280, opened 2019-08-23
Keep reference to multiple material when reading obj

DerouineauNicolas:
This PR adress issue #1276. It goes along with PR 2 on libigl-testdata.

I choose a tuple list to store (faces vertices index, ..., material). I'm not sure it's a right choice and I'll be happy to have a second opinion.

I did a function overload for readOBJ. This leads to code duplication (which is always bad).


jdumas:
So this code is node reading the material file at all? Apart from that, I would suggest returning the list of face-material associations as a vector or ranges into face indices. So a `vector<tuple<string, int, int>>` maybe? Where each a entry `(material_name, f0, f1)`  means faces `f0 <= f < f1` are using `material_name`.

To avoid code duplication, you could also consider having each function use the mode generic overload (that reads material associations) and drops the material info if not needed.

DerouineauNicolas:
Thanks for the suggestion, your solution for the face/materail association is indeed simpler.

I don't understand your suggestion to avoid code duplication. Could you point me to an example about generic function overload in C++ ?

jdumas:
Just look at [this one](https://github.com/libigl/libigl/blob/b0d7740e0b7e887a7e93601c4c557ecf762b389b/include/igl/readPLY.cpp#L207) for example. Use an overloaded function and create temporary arguments to pass to the *more generic* version of the function (the one with the most arguments).

DerouineauNicolas:
@jdumas Should be ok now for the vector<tuple<string, int, int>>

jdumas:
Looks like the CI didn't pass, probably you need to update the SHA1 commit for the test data repository?

DerouineauNicolas:
Hum ... I think I didn't catch how to update the SHA1 for the test data repository.

jdumas:
It's [here](https://github.com/libigl/libigl/blob/c645aac0c5852fad6fabe8c192fdc8675e607263/cmake/LibiglDownloadExternal.cmake#L178).

DerouineauNicolas:
Correct me if I'm wrong, but I guess the PR on the test data repo needs to be merged first.

jdumas:
You need to create a PR on the `libigl-test-data` repo, so that your SHA1 becomes known to github under this URL. Maybe you need to change also the URL in the CMake to point to your fork (for the purpose of this PR). Once this passes the CI, we can merge both the PR in the test data repo, and update this one to point to the new tip in libigl-test-data's master.

jdumas:
So you may need to also update the URL in the CMake to point to your own fork. Once this passes the CI, we can merge the PR in the test-data repo, and update this one to point to the new tip of the test-data master branch.

Issue #1059, opened 2018-12-26
Unit test documentation on webpage is out of date


alecjacobson:
https://libigl.github.io/unit-tests/ still references gtest setup. 

We should have a description of the new setup and instructions on how to isolate tests etc.

For example, the current dev branch is failing for one of the "test on a bunch of meshes" tests ( https://travis-ci.org/libigl/libigl/jobs/472451023 ). Is it possible to isolate the failure or figure out which exact test is failing?

#### Check all that apply (change to `[x]`)
- [X] Windows
- [X] Mac OS X
- [X] Linux

See <https://libigl.github.io/CONTRIBUTING/#bugreport> for more tips.

jdumas:
Lots of pages may be outdated on the website, I've listed pages to check in [libigl.github.io/#7](https://github.com/libigl/libigl.github.io/issues/7).

PR #1288, opened 2019-09-16
# fix issue #1237

stnoh:
- [ ] Adds new .cpp file.
- [x] This is a minor change.

Hello, I tested the fix by @xinyazhang on #1230 and #1237 in my environment (VS2017/Windows 10). Unfortunately, it still doesn't work for me. I checked it with the debugger and found that Aeq is still empty after running the original line of code (maybe this should be Eigen bug).  
To fix this, I made one additional line to create M.diagonal().transpose() and make it sparse representation in the second line. Additionally, I also set the type as Scalar, not "double" to follow the original template.  
I certificated that it now works in my environment.

### DIFF
--- a/include/igl/heat_geodesics.cpp
+++ b/include/igl/heat_geodesics.cpp
@@ -84,7 +84,8 @@ IGL_INLINE bool igl::heat_geodesics_precompute(
         return false;
       }
     }
-    const Eigen::SparseMatrix<double> Aeq = M.diagonal().transpose().sparseView();
+    const DerivedV M_diag_tr = M.diagonal().transpose();
+    const Eigen::SparseMatrix<Scalar> Aeq = M_diag_tr.sparseView();
     L *= -0.5;
     if(!igl::min_quad_with_fixed_precompute(
       L,Eigen::VectorXi(),Aeq,true,data.Poisson))


jdumas:
There is definitely a Eigen bug to be reported there. I was waiting for someone to create a MWE and report the bug upstream before I merge the PR. We should also have a unit test to catch this use-case, to make sure we don't break it again inadvertently.

jdumas:
@stnoh @xinyazhang this has been hanging for a while. Has anyone produced a MWE and reported the bug upstream to Eigen?

stnoh:
Nope, I didn't because @xinyazhang said he was writing a bug report in #1230 ...
However, I still cannot find his report from Eigen Bugzilla, neither.
http://eigen.tuxfamily.org/bz/buglist.cgi?component=Sparse

stnoh:
Urgh... Do you have an Eigen bugzilla account, @jdumas ?  
Just now I tried to report it, but their system does not permit self-registration.  
Before reporting a bug, I need to send an e-mail for Bugzilla account request. But I'm not sure how much it will take time to get an account.

jdumas:
Yeah it says self-registration is disabled due to spam. But if you email the address shown on the webpage they should create an account pretty quickly. I think I got my account before they disabled self-registration due to spam.

xinyazhang:
> It doesn't seem @xinyazhang ever reported it upstream, so maybe you can do it @stnoh? (Worst case we'll have it reported twice...).
Sorry for the delay, I'm working on Windows for the most of my time. @stnoh @jdumas have you filed the report? I could file one tonight with my home laptop.

jdumas:
I haven't, and I'm working on Linux most of the time, so I would appreciate if one of you could do it.

xinyazhang:
@stnoh @jdumas submitted as https://eigen.tuxfamily.org/bz/show_bug.cgi?id=1754

stnoh:
Oh, I got Eigen account now, but @xinyazhang already reported a bug.
(In my case, I AM working in Windows for the most of time 😃)

jdumas:
Hmm I realized I merged this but we still need to add a unit test to make sure we don't reopen the bug with a future update to Eigen. Would any of you Windows guys be willing to add that?

stnoh:
Sorry for late reply... I may write something, but where should I start?  
I looked out some information from the repo, but it was slightly confusing, because #1059 meant some part of the instruction page (https://libigl.github.io/unit-tests/) is out-of-date...

jdumas:
I would suggest to start from an existing unit test, e.g. [this one](https://github.com/libigl/libigl/blob/dev/tests/include/igl/per_face_normals.cpp) (I picked randomly). Copy the structure and write your own test. Let me know if you have any question!

PR #1303, opened 2019-10-02
update nrosy to match the MIQ paper

hankstag:
The current nrosy implementation is not consistent with the MIQ paper, and the singularity index has wrong sign. This PR is to fix this.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/copyleft/comiso/nrosy.cpp
+++ b/include/igl/copyleft/comiso/nrosy.cpp
@@ -482,7 +482,7 @@ void igl::copyleft::comiso::NRosyField::computek()
       // Map the two triangles in a new space where the common edge is the x axis and the N0 the z axis
       Eigen::MatrixXd P(3,3);
       Eigen::VectorXd o = V.row(F(fid0,fid0_vc));
-      Eigen::VectorXd tmp = -N0.cross(common_edge);
+      Eigen::VectorXd tmp = N0.cross(common_edge);
       P << common_edge, tmp, N0;
       P.transposeInPlace();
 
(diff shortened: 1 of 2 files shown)


danielepanozzo:
This looks ok, but it does not pass the build.

include/igl/copyleft/comiso/nrosy.cpp
```diff
@@ -494,32 +494,32 @@ void igl::copyleft::comiso::NRosyField::computek()
...
 
-      assert(V1(fid1_vc,2) < 10e-10);
-      assert(V1((fid1_vc+1)%3,2) < 10e-10);
+      assert(V1(fid1_vc,2) < 1e-10);
+      assert(V1((fid1_vc+1)%3,2) < 1e-10);
 
       // compute rotation R such that R * N1 = N0
       // i.e. map both triangles to the same plane
-      double alpha = -std::atan2(V1((fid1_vc + 2) % 3, 2), V1((fid1_vc + 2) % 3, 1));
+      double alpha = -std::atan2(-V1((fid1_vc + 2) % 3, 2), -V1((fid1_vc + 2) % 3, 1));
```
jiangzhongshi: What difference does the sign make here?

include/igl/copyleft/comiso/nrosy.cpp
```diff
@@ -528,17 +528,25 @@ void igl::copyleft::comiso::NRosyField::computek()
...
+        return (0 == y) ? x : x - y * floor(x/y);
+      };
+      ktemp = pos_fmod(ktemp, 2*igl::PI);
+      if (ktemp > igl::PI) 
+        ktemp -= 2*igl::PI;
+      
       // just to be sure, rotate ref0 using angle ktemp...
       Eigen::MatrixXd R2(2,2);
-      R2 << std::cos(ktemp), -std::sin(ktemp), std::sin(ktemp), std::cos(ktemp);
+      R2 << std::cos(-ktemp), -std::sin(-ktemp), std::sin(-ktemp), std::cos(-ktemp);
```
jiangzhongshi: This line looks strange, should it be equivalent to 
`R2 << std::cos(ktemp), std::sin(ktemp), -std::sin(ktemp), std::cos(ktemp);`

include/igl/copyleft/comiso/nrosy.cpp
jdumas: If you flip both points coordinates you should get a rotation or pi.

PR #1305, opened 2019-10-03
fixes #1295

### LINKED ISSUE #1295: & vs && in Viewer.cpp
#### Describe your issue

The line below comment "// Set default windows width" in Viewer.cpp is like [this](https://github.com/libigl/libigl/blob/b0d7740e0b7e887a7e93601c4c557ecf762b389b/include/igl/opengl/glfw/Viewer.cpp#L162):

>       if (windowWidth <= 0 & core_list.size() == 1 && core().viewport[2] > 0)

, which causes a warning in Visual Studio 2015. Is there any special reason why & is used here instead of &&? If not, I think it might be better to change the line to:

>       if (windowWidth <= 0 && core_list.size() == 1 && core().viewport[2] > 0)

Thanks. 

...

- [x] Windows
- [x] Mac OS X
- [x] Linux
- [x] I have tried the `dev` branch and the problem persists

jiangzhongshi:

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/opengl/glfw/Viewer.cpp
+++ b/include/igl/opengl/glfw/Viewer.cpp
@@ -159,12 +159,12 @@ namespace glfw
     else
     {
       // Set default windows width
-      if (windowWidth <= 0 & core_list.size() == 1 && core().viewport[2] > 0)
+      if (windowWidth <= 0 && core_list.size() == 1 && core().viewport[2] > 0)
         windowWidth = core().viewport[2];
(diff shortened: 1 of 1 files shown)


jiangzhongshi:
@jdumas  it seems that travis is behaving weird. How can I trigger the build again?

jdumas:
If you click on "detail" and then on the job that failed, there is a button on the top-right to re-start the build.

PR #1313, opened 2019-10-10
octree + knn imp

mtaoaugmenta:
There are a  variety of minor improvements to KNN/octree to allow it to be run with floats and so KNN can be run between two different point clouds.

There is an API change is that "const KType&" was changed to int because it didn't need to be template deduced. The new knn call uses P and V point clouds as the points that need neighbors found and V for the points that will be neighbors. P==V reduces to the original implementation.

When |V| < k, the code now shortcuts to just writing out a list of indices, sorted by distances.

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/include/igl/knn.cpp
+++ b/include/igl/knn.cpp
@@ -1,88 +1,120 @@
 #include "knn.h"
+#include "sort.h"
 #include "parallel_for.h"
 
 #include <cmath>
 #include <queue>
+#include <set>
+#include <algorithm>
 
 namespace igl {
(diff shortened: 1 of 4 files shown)


jdumas:
@mtaoaugmenta Can you give us permission to write to your fork (it's a checkbox to enable on this PR) so we can update your PR?

mtao:
done! thanks

jdumas (approved):

PR #1319, opened 2019-10-17
Revert "Set default value for ViewerCore::rotation_type"

alecjacobson:
Reverts libigl/libigl#1318

Two axis should be default


jdumas:
It still is. `set_rotation_type(ViewerCore::ROTATION_TYPE_TWO_AXIS_VALUATOR_FIXED_UP);
` is called in the constructor of the `ViewerCore`. But `set_rotation_type()` is relying on a previous value of `rotation_type` being defined. May be a bit counter-intuitive though. Maybe best to initialize `rotation_type` in the `.cpp` and not in the header to avoid confusing people?

PR #1325, opened 2019-10-22
fix embree library dir and `define` macro

xarthurx:

- [ ] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
Thanks but the include path is correct. You need to update your CMake to link against `igl::embree` instead.

PR #1332, opened 2019-10-27
cut mesh refactor

hankstag:

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

Current implementation of cut_mesh function add duplicated vertices directly after the original one. Breaking the original indices of vertices is undesirable sometimes. This new function will append all new vertices at the bottom of V matrix and provides an index mapping to find the correspondence after cut.

Added some simple unit test. Since the mapping of indices changed, the test data for MIQ needs to be updated.

### DIFF
--- a/include/igl/cut_mesh.cpp
+++ b/include/igl/cut_mesh.cpp
@@ -54,12 +54,12 @@ IGL_INLINE void igl::cut_mesh(
   igl::triangle_triangle_adjacency(F,FF,FFi);
 
   // store current number of occurance of each vertex as the alg proceed
-  Eigen::Matrix<Index,Eigen::Dynamic,1> o(V.rows());
-  o.setConstant(1);
+  Eigen::Matrix<Index,Eigen::Dynamic,1> occurence(V.rows());
+  occurence.setConstant(1);
   
-  // set target number of occurance of each vertex
-  Eigen::Matrix<Index,Eigen::Dynamic,1> g(V.rows());
-  g.setZero();
+  // set eventual number of occurance of each vertex expected
+  Eigen::Matrix<Index,Eigen::Dynamic,1> eventual(V.rows());
+  eventual.setZero();
   for(Index i=0;i<F.rows();i++){
     for(Index k=0;k<3;k++){
       if(C(i,k) == 1){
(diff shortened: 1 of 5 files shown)


jiangzhongshi:
MIQ unit test failed...

hankstag:
the two DMAT files in MIQ test data also needs to be updated since the index are changing

include/igl/cut_mesh.cpp
```diff
@@ -1,330 +1,142 @@
...
+  const Eigen::MatrixBase<DerivedC>& C,
+  Eigen::PlainObjectBase<DerivedI>& I
+){
+
+  typedef typename DerivedF::Scalar Index;
+  DerivedF FF, FFi;
+  igl::triangle_triangle_adjacency(F,FF,FFi);
+
+  // store current number of occurance of each vertex as the alg proceed
+  Eigen::Matrix<Index,Eigen::Dynamic,1> o(V.rows());
```
jiangzhongshi: Consider change this `o` and `g` to a more meaningful name, for future maintenance.

include/igl/cut_mesh.cpp
```diff
@@ -1,330 +1,142 @@
...
-      if (Fcut(fi,k)==-1)
-      {
-        //we need to add a vertex
-        Fcut(fi,k) = newNumV;
-        newNumV ++;
-        Vcut.conservativeResize(newNumV, Eigen::NoChange);
-        Vcut.row(newNumV-1) = V.row(F(fi,k));
+  }
+  // add one extra occurance for boundary vertices
+  auto is_border = igl::is_border_vertex(F);
```
jiangzhongshi: This step seems to be wasteful, as the information is better acquired from the existing `FF, FFi`

include/igl/cut_mesh.cpp
```diff
@@ -1,330 +1,142 @@
...
+IGL_INLINE void igl::cut_mesh(
+  Eigen::PlainObjectBase<DerivedV>& V,
+  Eigen::PlainObjectBase<DerivedF>& F,
+  const Eigen::MatrixBase<DerivedC>& C,
+  Eigen::PlainObjectBase<DerivedI>& I
+){
+
+  typedef typename DerivedF::Scalar Index;
+  DerivedF FF, FFi;
+  igl::triangle_triangle_adjacency(F,FF,FFi);
```
jiangzhongshi: Maybe consider another overload that takes `FF, FFi` as input parameter (partially honoring the old interface), so that an external user will have more flexibility when cutting incrementally.

include/igl/cut_mesh.cpp
hankstag: i agree, i'm gonna make the changes

PR #1337, opened 2019-11-02
Bug fix in marching_tets

alecjacobson:
J was not always being filled in and then it was erroneously resized to the wrong length. This most immediately _**broke**_ the WindingNumber tutorial.

### DIFF
--- a/include/igl/marching_tets.cpp
+++ b/include/igl/marching_tets.cpp
@@ -147,14 +147,13 @@ void igl::marching_tets(
 
   for (int f = 0; f < faces.size(); f++)
   {
+    const int ti = faces[f].second;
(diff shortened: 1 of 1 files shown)


include/igl/marching_tets.cpp
```diff
@@ -151,6 +151,10 @@ void igl::marching_tets(
     {
       const int vi = faces[f].first[v];
       const int ti = faces[f].second;
+      assert(ti>=0);
+      assert(ti<TT.rows());
+      // this will overwrite, but should be with the same value
+      J(f) = ti;
```
jdumas: We could also move the `const int ti = faces[f].second;` outside the loop on `v`.

Issue #1343, opened 2019-11-05
Creating point cloud from mesh sampling


ChimMeya:
Hi,
I'm trying to create point clouds from mesh using `random_points_on_mesh(n, V, F, B, FI)` function (see code below). After some iterations, I get this error and sometimes I get it from the first iteration when n is big (about 100000 samples). 
![error_libigl](https://user-images.githubusercontent.com/26687429/68224427-e0c7ca80-ffee-11e9-9021-37d588f6c273.PNG)
Debbuging:
![error2](https://user-images.githubusercontent.com/26687429/68224548-1e2c5800-ffef-11e9-8db0-493c10a9b78c.PNG)

My code:
```
int main() {
	for (int i = 0; i < 100; i++) {
		get_sampled_pointCloud_from_OFF("bathtub_0001.off");
	}
	return 0;
}
Eigen::MatrixXd get_sampled_pointCloud_from_OFF(const std::string fileName) {
	Eigen::MatrixXd V;
	Eigen::MatrixXi F;
	igl::readOFF(fileName, V, F);
        Eigen::VectorXi FI;
	Eigen::MatrixXd B;
	
	igl::random_points_on_mesh(100000, V, F, B, FI);
	Eigen::MatrixXd samples(B.rows(), 3);
	for (int i = 0; i < samples.rows(); i++){
		samples.row(i) =
			B(i, 0)*V.row(F(FI(i), 0)) +
			B(i, 1)*V.row(F(FI(i), 1)) +
			B(i, 2)*V.row(F(FI(i), 2));
	}
	return samples;}
```
This code is inspired from [This comment](https://github.com/libigl/libigl/issues/808#issuecomment-397099198). I changed the number of rows since I want to create point cloud from the n sampled points.
I couldn't figure the problem! I'm new in Libigl.
Thanks.
...
- [x] Windows
- [ ] Mac OS X
- [ ] Linux
- [ ] I have tried the `dev` branch and the problem persists

ChimMeya:
I printed some values of `FI(i)` and I found that some of them are equal to `-1`.
I think that this shouldn't happen, is it? is the problem in `random_points_on_mesh(n, V, F, B, FI)`? why it returns `-1`?

jiangzhongshi:
It seems to be a bug. Could you share the mesh that is creating the problem?

ChimMeya:
It's from ModelNet10, the first bathtub test image.

ChimMeya:
[...]

PR #1345, opened 2019-11-10
fix https://github.com/libigl/libigl/issues/1343

alecjacobson:
floating point error could lead final value in cumsum to be <1. It should be ==1 so that the random values [0,1] do not exceed range spanned by cumsum.

(I couldn't manage to trigger a bad random value, but I confirmed that the old code had values of 0.99999999999993516 instead of 1.0. So I now believe it _could_ have happened.)

https://github.com/libigl/libigl/issues/1343

- [ ] Adds new .cpp file.
- [x] This is a minor change.


include/igl/random_points_on_mesh.cpp
```diff
@@ -35,6 +34,10 @@ IGL_INLINE void igl::random_points_on_mesh(
   A0.bottomRightCorner(A.size(),1) = A;
   // Even faster would be to use the "Alias Table Method"
   cumsum(A0,1,C);
+  const Scalar Cmax = C(C.size()-1);
+  assert(Cmax > 0 && "Total surface area should be positive");
+  // Why is this more accurate than `C /= C(C.size()-1)` ?
+  for(int i = 0;i<C.size();i++) { C(i) = C(i)/Cmax; }
```
jiangzhongshi: I think the reason is: here the last value is explicitly made equal to `Cmax/Cmax` which will more likely be `1`.

alecjacobson: I still don't get it. I expect it to be _exactly_ 1 in both cases. Why would Eigen's broadcaster not give the same result as this loop?

jdumas:
Do you have an example of mesh/nb samples where the old code produced 0.99999999999993516 in the normalized cumsum?

include/igl/random_points_on_mesh.cpp
jdumas: I just tried on the meshes from the tutorial data, and I always get `1` if I replace this line by `C /= C(C.size()-1)`, even for a very large number of samples (10,000,000).

alecjacobson:
I tried on the xyz dragon that has millions of tris

jdumas:
So I tried also on this model, and if I use
```
  cumsum(A0,1,C);
  C /= C(C.size()-1);
  std::cout << std::hexfloat << C(C.size()-1) << std::endl;
```
Then it always prints exactly 1. So I don't really understand your comment _"Why is this more accurate than `C /= C(C.size()-1)` ?"_. To be fair, the old code (where you divide by the total area first `A /= A.array().sum();`) does produce a Cmax != 1, but that's not surprising since the floating point error accumulate in a different way when summing the normalized areas.

PR #1350, opened 2019-11-12
Azure for windows

teseoch:

- [ ] Adds new .cpp file.
- [x] This is a minor change.

This is how it looks:
https://dev.azure.com/Teseo0677/igl/_build/results?buildId=81

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -57,6 +57,11 @@ jobs:
     vmImage: 'macos-10.13'
 
   steps:
+
+  - task: CmdLine@2
+    inputs:
+      script: |
+        brew install gmp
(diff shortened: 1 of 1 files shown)


alecjacobson:
does this only do windows? It seems cleaner to use travis for everything.

jdumas:
Azure does both Windows/macOS/Linux. But the time limits and number of cpu we may have on different services might vary. I feel we can get away with more parallel builds if we have ci on multiple services. Or we could try to have both for all 3 platforms for a while, and evaluate which one works the best for our usage.

teseoch:
The other builds are commented.

There is still the cache to enable and some fixes to do for the other 2 platforms

PR #1361, opened 2019-11-25
dont use floating point values to index eigen types

Janos95:
As discussed in the corresponding issue, I changed the necessary files so there are no Eigen array accesses using floating point types.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
@Janos95 can you rename `cut_mesh_rosy.cpp` back to `cut_mesh.cpp`?

Janos95:
Sorry, I thought this was already merged so I pushed some other work onto my fork.

jdumas:
Thanks! But I think my comment from last time still applies (about casting only the doubles, not the whole expr). I would have merged it earlier but our current CI is a bit wonky (it often times out and I need to restart it manually).

Janos95:
I think I corrected that in 5b996d9 .

jdumas:
5b996d9 only changed one line, but I think the comment applies to other lines as far as I can tell from the PR summary.

PR #1362, opened 2019-11-25
remove LIBIGL_USE_STATIC_LIBRARY in favor of standard BUILD_SHARED_LIBS

jasjuang:
There is actually a standard cmake variable [`BUILD_SHARED_LIBS`](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) that does exactly what the cmake variable `LIBIGL_USE_STATIC_LIBRARY` intends to do. It's better to eliminate `LIBIGL_USE_STATIC_LIBRARY` because what happened was when I first encounter libigl, I instinctively compile it with 

```
cmake -DBUILD_SHARED_LIBS=ON ..
make -j
sudo make install
``` 
and then I realized a simple downstream test project that links to libigl wouldn't compile 

```
/usr/include/igl/any.h:32:12: fatal error: any.cpp: No such file or directory
 #  include "any.cpp"
            ^~~~~~~~~
compilation terminated.
```

after further investigation, I realized it's because `BUILD_SHARED_LIBS` is set to on so it builds a shared library, but the variable `LIBIGL_USE_STATIC_LIBRARY` is by default on hence the confusion. After applying this PR, I can now compile and run my downstream project successfully.

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,11 +15,6 @@ option(LIBIGL_BUILD_TUTORIALS  "Build libigl tutorial"         ${LIBIGL_TOPLEVEL
 option(LIBIGL_BUILD_PYTHON     "Build libigl python bindings"  ${LIBIGL_TOPLEVEL_PROJECT})
 option(LIBIGL_EXPORT_TARGETS   "Export libigl CMake targets"   ${LIBIGL_TOPLEVEL_PROJECT})
 
-# USE_STATIC_LIBRARY speeds up the generation of multiple binaries,
-# at the cost of a longer initial compilation time
-# (by default, static build is off since libigl is a header-only library)
-option(LIBIGL_USE_STATIC_LIBRARY "Use libigl as static library" ON)
-
 # All dependencies that are downloaded as cmake projects and tested on the auto-builds are ON
 # (by default, all build options are off)
 option(LIBIGL_WITH_COMISO            "Use CoMiso"                   ON)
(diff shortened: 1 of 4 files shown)


jdumas:
I don't really understand your concern here. `LIBIGL_USE_STATIC_LIBRARY` has a different meaning than `BUILD_SHARED_LIBS`. libigl can be used in 2 modes: as a header-only library (the default if you include ` cmake/libigl.cmake` in your CMake), and as a "static" library. The second mode requires clients to compile with the preprocessor define `-DIGL_STATIC_LIBRARY`, otherwise implementation files will be included. The term "static" lib is a bit of a misnomer, since you can actually compile libigl as either a static or a shared library, although our CMake is currently not configured this way.

The problem you are encountering is different. You are compiling libigl as a static lib (which is the default when building the repo from the root), and installing it on your system (which I wouldn't recommend). The problem is that I'm guessing you are trying to `#include <>` a libigl header, but do not pass the flag ``-DIGL_STATIC_LIBRARY`. But because libigl is compiled as a static library, only the header files (.h) are copied in your system, and not the .cpp. If you compile/install libigl as a header-only library, both .h and .cpp will be copied over.

In short:
- `LIBIGL_USE_STATIC_LIBRARY` has nothing to do with `BUILD_SHARED_LIBS`
- You probably didn't set up correctly your CMake project to use libigl. Take a look at the [libigl-example-project](https://github.com/libigl/libigl-example-project) for the recommended way to use libigl in your project.

Issue #1109, opened 2019-03-13
Improve label rendering in the viewer


jdumas:
We don't really have a thread to track and discuss issues related to label rendering in the current viewer, so I'm creating one.

Related discussions: #978 and #719

Issues:
- ImGui label rendering would currently crash when rendering too many labels. This is a know issue but has never been truly fixed.
- Labels are added to `ViewerData()` but require the `ImGui` to be displayed. This is kinda awkard, and also starts to be a bit weird with the multiple viewports feature (#978). We should talk about how we want to design this.

Possible solutions:
- Move label rendering into a different plugin so it doesn't require ImGui.
- Font rendering could be handled with a separate minimal third-party library, e.g. [FontStash](https://github.com/memononen/fontstash).
- Some other pointers on [font rendering](https://aras-p.info/blog/2017/02/15/Font-Rendering-is-Getting-Interesting/).

I think this would make a good final project for a student as well, if you catch my drift :D

deGravity:
Labels also current float on top of everything, regardless of z-ordering, so in a scene with many labels it can be very difficult to read them.

samitch:
Labels are too small. Only the closest label is large enough to be legible. How can I make the labels bigger?

PR #1366, opened 2019-11-29
Templates and other minor changes

alecjacobson:
- [ ] Adds new .cpp file.
- [x] This is a minor change.


include/igl/opengl/ViewerData.cpp
```diff
@@ -19,22 +19,23 @@
...
-  show_lines(true),
-  invert_normals(false),
-  show_overlay(true),
-  show_overlay_depth(true),
-  show_vertid(false),
-  show_faceid(false),
-  show_labels(false),
-  show_texture(false),
+  // UINT_MAX sets all bits to 1 so these defaults get applied to all view cores
+  show_faces        (UINT_MAX),
```
jdumas: Using `~unsigned(0)` would be better to carry the meaning that we are setting all bits.

include/igl/opengl/glfw/imgui/ImGuiMenu.cpp
```diff
@@ -335,6 +335,7 @@ IGL_INLINE void ImGuiMenu::draw_labels_window()
 
 IGL_INLINE void ImGuiMenu::draw_labels(const igl::opengl::ViewerData &data)
 {
+  // Alec: How can we get these to respect (optionally) the depth of the scene?
```
jdumas: It should be possible to do with ImGui, albeit a bit difficult. See e.g. [this discussion](https://github.com/ocornut/imgui/issues/1162). There are other important issues with our current setup for drawing labels (see #1109), so we might be better off re-implementing 3D labels outside of the ImGui framework.

PR #1370, opened 2019-11-30
Avoid involving Eigen/SparseExtra in slice.

qnzhou:
Making libigl a little bit more commercial friendly by removing unsupported Eigen features in `slice`, which is used by a bunch of other functions.  All unit tests still pass.  Performance implication is to be determined.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
This PR is a great impetus for us to review `slice` for sparse matrices. This is a highly used and super useful routine. Ideally Eigen just supported this directly.

Does eigen support matlab/numpy style slicing now?

Assuming not... We've been keeping this deprecated DynamicSparsematrix business around because it was actually faster than Eigen's replacement (setfromTriplets) at obsoletion time.

Is that still the case? Can we get rid of dynamicsparsematrix?

This routine is used so often that any performance hit will have ripples all over so I'm weary to switch "just because it's cleaner" if we get even a 0.75x performance hit.

h00shi:
Eigen 3.3.9 has some support for slicing. But it is not released yet.
https://eigen.tuxfamily.org/dox-devel/group__TutorialSlicingIndexing.html

qnzhou:
I conducted some performance testing using this performance test function: https://gist.github.com/qnzhou/d15e978fccdd3239fcad16a56a7af6f9

With Apple LLVM version 9.1.0 and `-O3` optimization:

Slicing with `Eigen::DynamicSparseMatrix`:
```
Time: 0.00335185
```

Slicing with the good old `Eigen::Triplet`:
```
Time: 0.00187511
```

It seems by reverting to `Eigen::Triplet`, libigl actually gained some performance advantage.  Of course, can someone please validate my tests and results.

jdumas:
You should be able to get an even greater performance gain by using .insert() directly instead of the triplet list. The trick is to reserve the exact number of nnz by column (which costs a VectorXi of #cols). Then insert() can be done in O(1) and you avoid the copy + sort/transpose of the nnz incurred by setFromTriplets. The only downside is that it might require to update Eigen 3.3. (This came up a couple of times in previous github issues but I'm on the phone right now so I can't find them).

jdumas:
Oh, and we should be able to remove the x3 duplicated code by replacing `Eigen::MatrixBase` and `Eigen::ArrayBase` by `Eigen::DenseBase` no?

qnzhou:
Feel free to push directly to my branch.  😄 It is strange that my mac is complaining about one set of template specialization is missing but travis is happy.

jdumas:
Yeah I'll give it a shot latter this week.

jdumas:
Ok here is my attempt at `igl::slice` using `insert()`:
https://gist.github.com/jdumas/6d25a6835089df829fcc82d1bf9a2892

Caveat: it only works for *sorted* lists of indices for now. For unsorted inputs, an additional sorting step on the sliced indices would be needed... which mean that it would be the same as @qnzhou's version using `Triplet`. But I believe in most cases the indices are actually already sorted, in which case my version is about x2 as fast as @qnzhou's version.

qnzhou:
I think 
1. We should dump `Eigen::DynamicSparsematrix` because it is deprecated and is actually slower.
2. `igl::slice` should use triplets for sparse matrix assembly instead.
3. We should add `igl::slice_sorted` as a separate method when Eigen3.3 is available and the input indices are sorted.

jdumas:
Agreed. The `slice_sorted` actually works just as fast with Eigen 3.1 and 3.2 though, so it's even better :D

jdumas:
Ok I've updated this PR:
- Cleaned up `igl::slice` and use `DenseBase` as the common type to avoid duplicated code.
- Added `igl::slice_sorted` with unit test.
- Updated Catch2 and use `BENCHMARK` macro to measure the timings of `igl::slice` vs `igl::slice_sorted`

PR #1371, opened 2019-12-01
Phong color mapping for scalar field data, refactor tutorials 

alecjacobson:
We'll need to update the tutorial screenshots to match these better visualizations.

# Check all that apply (change to `[x]`)
- [x] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/opengl/ViewerData.h
```diff
@@ -96,6 +95,26 @@ class ViewerData
...
     const Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic>& A);
+  // Set pseudo-colorable scalar data associated with the mesh.
+  //
+  // Inputs:
+  //   caxis_min  caxis minimum bound
+  //   caxis_max  caxis maximum bound
+  //   D  #V by 3 list of colors
+  //
+  // To-do: support #F by 1 per-face data
+  IGL_INLINE void set_data(
```
jdumas: In `igl::colormap` we pass `minZ` and `maxZ` after the `Z` data. Should we do the same here for consistency?

include/igl/opengl/ViewerData.cpp
```diff
@@ -260,6 +264,44 @@ IGL_INLINE void igl::opengl::ViewerData::set_texture(
...
 
+IGL_INLINE void igl::opengl::ViewerData::set_data(
+  const double caxis_min,
+  const double caxis_max,
+  const Eigen::VectorXd & D)
+{
+  if(!show_texture)
+  {
+    Eigen::MatrixXd CM;
+    igl::parula(Eigen::VectorXd::LinSpaced(21,0,1).eval(),false,CM);
```
jdumas: We should add an option `igl::ColorMapType` to specify the colormap. Also, our colormaps are defined using 256 samples in `igl/colormap.cpp`, so we should use the same number of samples here.

alecjacobson: see other comment. 21 is to create identifiable intervals which helps with scalar field visualization.

include/igl/opengl/ViewerData.h
alecjacobson: agreed.

include/igl/opengl/ViewerData.h
```diff
@@ -96,6 +95,26 @@ class ViewerData
     const Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic>& G,
     const Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic>& B,
     const Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic>& A);
+  // Set pseudo-colorable scalar data associated with the mesh.
+  //
+  // Inputs:
+  //   caxis_min  caxis minimum bound
+  //   caxis_max  caxis maximum bound
+  //   D  #V by 3 list of colors
```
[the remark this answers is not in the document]
alecjacobson: agreed

PR #1372, opened 2019-12-01
Add colormap option to set_data + update jet implementation.

jdumas:
- [x] This is a minor change.


jdumas:
I'm also tempted to simply mark jet as deprecated and add the *turbo* colormap:
https://ai.googleblog.com/2019/08/turbo-improved-rainbow-colormap-for.html

alecjacobson:
Yeah. I like this idea, too.

include/igl/opengl/ViewerData.cpp
```diff
@@ -265,24 +263,25 @@ IGL_INLINE void igl::opengl::ViewerData::set_texture(
...
-    igl::parula(Eigen::VectorXd::LinSpaced(21,0,1).eval(),false,CM);
+    igl::colormap(cmap,Eigen::VectorXd::LinSpaced(256,0,1).eval(),0,1,CM);
     set_colormap(CM);
-  } 
+  }
   set_uv(((D.array()-caxis_min)/(caxis_max-caxis_min)).replicate(1,2));
 }
 
-IGL_INLINE void igl::opengl::ViewerData::set_data( const Eigen::VectorXd & D)
+IGL_INLINE void igl::opengl::ViewerData::set_data( const Eigen::VectorXd & D, igl::ColorMapType cmap)
```
alecjacobson: Let's keep a `set_data(D)` that uses a default colormap. 

We should also use `const` for input params.

alecjacobson: nvm about default. Found the default it in the .h

jdumas:
Ok I switched jet to turbo. Here is a comparison:

Jet:
2019-12-01-165823_1280x800_scrot

Turbo:
2019-12-01-165843_1273x814_scrot

include/igl/opengl/ViewerData.cpp
```diff
@@ -265,24 +263,25 @@ IGL_INLINE void igl::opengl::ViewerData::set_texture(
...
+  const Eigen::VectorXd & D,
+  double caxis_min,
+  double caxis_max,
+  igl::ColorMapType cmap)
 {
   if(!show_texture)
   {
     Eigen::MatrixXd CM;
-    igl::parula(Eigen::VectorXd::LinSpaced(21,0,1).eval(),false,CM);
+    igl::colormap(cmap,Eigen::VectorXd::LinSpaced(256,0,1).eval(),0,1,CM);
```
alecjacobson: this `256` should be reverted `21`.

Intervals are (by default) better for visualizing scalar fields than an effectively continuous map.

jdumas: Ok I see. I've added an argument to set the number of steps/interval then, with default = 21.

jdumas:
Btw I'm also questioning if we should make Parula our default colormap. Matplotlib developed Viridis and co. because Parula is copyrighted by Matworks. Do we really want to make it our default colormap, or should we stick with Viridis?

alecjacobson:
that's fine with me.

>

jdumas:
That's fine meaning ok to switch to viridis, or ok to keep parula?

alecjacobson:
sorry I was unclear. Switching to Viridis is fine. As long as it's close
enough looking to Parula that I don't need to remake all of the new
tutorial figures...

>

alecjacobson:
fwiw, I much more impressed with the colorbrewer colormaps than matlab's or
matplotlib's. Maybe we should support those and use one as a default.

>

jdumas:
You mean [this website](http://colorbrewer2.org)? Their colormaps are defined [here](https://github.com/axismaps/colorbrewer/blob/master/colorbrewer_schemes.js). Should be easy to integrate into libigl.

jdumas:
I've switched to Viridis by default. Should be close enough to Parula to not perturb the tutorials. I'll merge once the build passes. We can add the colorbrew palette in a separate PR. Given that these are discrete colormaps, we can add a function to directly return the colorbar given the number of steps required by the user.

PR #1379, opened 2019-12-06
Templates and other fixes for python bindings

teseoch:

- [ ] Adds new .cpp file.
- [x] This is a minor change.


include/igl/face_occurrences.cpp
```diff
@@ -60,7 +60,7 @@ IGL_INLINE void igl::face_occurrences(
   // Should really just rewrite using Eigen+libigl ...
   std::vector<std::vector<typename DerivedF::Scalar> > vF;
   matrix_to_list(F,vF);
-  std::vector<std::vector<typename DerivedC::Scalar> > vC;
+  std::vector<typename DerivedC::Scalar> vC;
```
teseoch: Can someone verify this?

I checked and it seemed correct, it looks like a major bug...

jdumas: This change looks correct. The function is not explicitly instantiated, so most likely nobody ever used it with this signature.

PR #1380, opened 2019-12-10
Make use of IGL_DEPRECATED

BruegelN:
* Use the macro in `igl::all_edges`, `igl::internal_angles` and `igl::is_border_vertex`.
CTRL+F for "deprecated" to find functions where the comment says they're deprecated.

* Removed #include deprecated.h for `igl::nchoosek` and `igl::readOBJ` because it looks like the functions which have been marked as deprecated are already gone. See [igl::nchoosek](https://github.com/libigl/libigl/commit/471f0c9c37478120dd910bef84bbe87fd96fcaa7#diff-f56795cb8c2b945d68d392cb5d89688e) and [igl::readOBJ](https://github.com/libigl/libigl/commit/dcd8c80fbc71cc85050f4875025b646aeda572c9#diff-e0dd9ddd2a25dde223acf9d1252f437f)

The compiler warning will look sth. like below if you use a deprecated func.
```bash
std::vector<bool> igl::is_border_vertex(const Eigen::MatrixBase<Derived>&, const Eigen::MatrixBase<U>&) [with DerivedV = Eigen::Matrix<double, -1, -1>; DerivedF = Eigen::Matrix<int, -1, -1>]’ is deprecated [-Wdeprecated-declarations]
   15 |   igl::is_border_vertex(V,F);
      |                            ^
In file included from include/igl/is_border_vertex.h:11,
                 from main.cpp:2:
include/igl/is_border_vertex.h:32:32: note: declared here
   32 |   IGL_INLINE std::vector<bool> IGL_DEPRECATED(is_border_vertex(
      |                                ^~~~~~~~~~~~~~
main.cpp:15:28: warning: ‘std::vector<bool> igl::is_border_vertex(const Eigen::MatrixBase<Derived>&, const Eigen::MatrixBase<U>&) [with DerivedV = Eigen::Matrix<double, -1, -1>; DerivedF = Eigen::Matrix<int, -1, -1>]’ is deprecated [-Wdeprecated-declarations]
   15 |   igl::is_border_vertex(V,F);
      |                            ^
In file included from include/igl/is_border_vertex.h:11,
                 from main.cpp:2:
include/igl/is_border_vertex.h:32:32: note: declared here
   32 |   IGL_INLINE std::vector<bool> IGL_DEPRECATED(is_border_vertex(
```

- [ ] Adds new .cpp file.
- [X] This is a minor change.

### DIFF
--- a/include/igl/all_edges.h
+++ b/include/igl/all_edges.h
@@ -8,6 +8,7 @@
 #ifndef IGL_ALL_EDGES_H
 #define IGL_ALL_EDGES_H
 #include "igl_inline.h"
+#include "deprecated.h"
 #include <Eigen/Dense>
 namespace igl
 {
@@ -26,9 +27,9 @@ namespace igl
   // show up once for each direction and non-manifold edges may appear more than
   // once for each direction).
   template <typename DerivedF, typename DerivedE>
-  IGL_INLINE void all_edges(
+  IGL_INLINE void IGL_DEPRECATED(all_edges(
     const Eigen::MatrixBase<DerivedF> & F,
-    Eigen::PlainObjectBase<DerivedE> & E);
+    Eigen::PlainObjectBase<DerivedE> & E));
 }
 
 #ifndef IGL_STATIC_LIBRARY
--- a/include/igl/all_edges.h
+++ b/include/igl/all_edges.h
@@ -27,9 +27,9 @@ namespace igl
   // show up once for each direction and non-manifold edges may appear more than
   // once for each direction).
   template <typename DerivedF, typename DerivedE>
-  IGL_INLINE void IGL_DEPRECATED(all_edges(
+  IGL_DEPRECATED IGL_INLINE void all_edges(
     const Eigen::MatrixBase<DerivedF> & F,
-    Eigen::PlainObjectBase<DerivedE> & E));
+    Eigen::PlainObjectBase<DerivedE> & E);
 }
 
 #ifndef IGL_STATIC_LIBRARY
--- a/include/igl/boundary_loop.cpp
+++ b/include/igl/boundary_loop.cpp
@@ -29,7 +29,7 @@ IGL_INLINE void igl::boundary_loop(
   triangle_triangle_adjacency(F,TT,TTi);
   vertex_triangle_adjacency(Vdummy,F,VF,VFi);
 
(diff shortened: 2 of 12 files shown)


jdumas:
Thanks! I would also vote for marking `igl::remove_duplicates()` as deprecated in favor of `igl::remove_duplicate_vertices()` (or vice-versa, @alecjacobson I never know which one I should use?).

Additionally I think we should revise the `IGL_DEPRECATED` macro. I don't see the need to make it take an argument (since we are mostly deprecating functions). Also, in C++14, we can make use of the official `[[deprecated]]` macro which should be compatible with more compilers. Taking example from [fmt](https://github.com/fmtlib/fmt/blob/744302add085f30120c57f73cddbcb826cc92de9/include/fmt/core.h#L121), I think we can revise the macro as follows:

```cpp
#ifdef __has_cpp_attribute
#  define IGL_HAS_CPP_ATTRIBUTE(x) __has_cpp_attribute(x)
#else
#  define IGL_HAS_CPP_ATTRIBUTE(x) 0
#endif

#ifndef IGL_DEPRECATED
#  if (IGL_HAS_CPP_ATTRIBUTE(deprecated) && __cplusplus >= 201402L) || \
      IGL_MSC_VER >= 1900
#    define IGL_DEPRECATED [[deprecated]]
#  else
#    if defined(__GNUC__) || defined(__clang__)
#      define IGL_DEPRECATED __attribute__((deprecated))
#    elif IGL_MSC_VER
#      define IGL_DEPRECATED __declspec(deprecated)
#    else
#      define IGL_DEPRECATED /* deprecated */
#    endif
#  endif
#endif
```

BruegelN:
Thank you for your feedback!
I like the suggested rework of `IGL_DEPRECATED`. Good point!
As far as I know `igl::remove_duplicates()` is not used by other libigl functions, I only found internal usage of `igl::remove_duplicate_vertices()` on quick searching for the functions names. So i _think_ `igl::remove_duplicates()` is the deprecated one. If @alecjacobson confirms this I'll add a `IGL_DEPRECATED` to `igl::remove_duplicates()`.

jdumas:
Awesome. I've edited my code snippet compared to the msg you might have received by email, just FWI (adding pragma message + macro for MSVC).

jdumas:
I just got the ok from Alec to mark `igl::remove_duplicates` as deprecated @BruegelN, please go ahead.

PR #1383, opened 2019-12-12
Add CMake file to sort targets into folders

jdumas:
Make it easier on the eye for users on Visual Studio/Xcode:

projects

How it works is the user just needs to call `set_igl_targets()` after including libigl in their CMake script. The function is called by default if libigl is a toplevel project, and users will need to call it manually if they want to configure this while including libigl in another project.

I've also added header files to the GLOB pattern so that it shows in Visual Studio.

- [x] This is a minor change.


jdumas:
Making a note for myself: `INTERFACE` targets were never designed by CMake to represent header-only libraries, which is why they don't appear in the IDE (e.g., Visual Studio). See related [CMake issue](https://gitlab.kitware.com/cmake/cmake/issues/19145) which will allow us to set the `FOLDER` property on `INTERFACE` targets in a future CMake version.

Issue #1229, opened 2019-06-15
Why force static linkage of MSVC runtime?


evouga:
#### Describe your issue

When compiling code using the new version of libigl, CMake outputs the following message:
"MSVC -> forcing use of statically-linked runtime."
and as it states, sets the program to link statically against MSVC's runtime libraries (/MT).

Why is this? The old libigl worked fine with /MD, as does the new version, when I manually change the compile flags back myself.

Forcing /MT is not a trivial decision, as it requires propagating this decision to all external libraries linked by the application...

#### Check all that apply (change to `[x]`)
- [X] Windows
- [ ] Mac OS X
- [ ] Linux
- [ ] I have tried the `dev` branch and the problem persists

See https://libigl.github.io/CONTRIBUTING/#bugreport for more tips.

evouga:
Looking more into this: it looks like there's an MSVC_RUNTIME variable that needs to be set in order to prevent libigl from using static linking by default. Note that there's already a USE_MSVC_RUNTIME_LIBRARY_DLL CMake variable that's set by GLFW and exposed as a CMake option, but it does not work because it's overridden by MSVC_RUNTIME.

It would be nice to:
 - restore /MD by default on Windows, unless there is a reason libigl needs /MT: /MD is less likely on average to cause link conflicts with external libraries
 - expose MSVC_RUNTIME as a CMake option() and make it work correctly with GLFW's USE_MSVC_RUNTIME_LIBRARY_DLL

jdumas:
From what I remember, we've added this [LibiglWindows.cmake](https://github.com/libigl/libigl/blob/f6b406427400ed7ddb56cfc2577b6af571827c8c/cmake/LibiglWindows.cmake) (that sets `MSVC_RUNTIME` to `static` by default), because of Embree, and more specifically to be able to compile and link the unit tests. Embree has this [msvc_post.cmake](https://github.com/embree/embree/blob/e766297adcdaf8e5d8dd996c15aed753255230fd/common/cmake/msvc_post.cmake) file that will mess around with `/MT` and `/MD` flags. Now, you should be able to override this `MSVC_RUNTIME` (it's not an option, CMake options can only be ON/OFF) if you set it yourself before including libigl (maybe you have to `FORCE` the setting of the variable, since it is cached).

[...]

PR #1389, opened 2019-12-19
Enable GitHub Actions for continuous integration

jdumas:
This PR enables GitHub Actions on the libigl repository for continuous integration.

### Highlights
- Same config file for Linux/macOS/Windows, easier to mainain.
- Incremental builds with 100% cache hits only take [10min](https://github.com/jdumas/libigl/runs/356969266) in total (huge improvement!).
- Continuous build and PR checks are only testing Release configurations, since build artifacts are much smaller.
- I added a [daily build](https://github.com/jdumas/libigl/commit/ebe96a7dad1a0e969de6bbed2aea366eef1e989f/checks?check_suite_id=366707333), which runs at 4am UTC every day, and test gcc 7, 8, 9, clang 7, 8, 9, macOS and Windows in both Debug+Release, and with and without `LIBIGL_USE_STATIC_LIBRARY`. The total runtime of a daily build is about 1h.

### GitHub Actions vs Azure Pipelines (#1350)
- Disk space limit of 16 GB on GitHub Actions (vs 10 GB for Azure Pipelines).
- Timout is 6h (AppVeyor is 1h).
- Directly integrated into GitHub, no setup required (compared to Azure Pipelines).
- Is likely to gain more traction and be expanded in the future.

### Other Notes
- Tutorials are disabled on Windows in Debug mode, since .pdb information generated by MSVC is gigantic and goes over the space limit.
- Caching is enabled on all 3 platforms, thanks to [sccache](https://github.com/mozilla/sccache) on Windows. CMake configuration was also surpringly easy.
- Caching still has some strict space limits  (400 MB per job, 2 GB total accross all jobs). So it's not ideal but if we limit builds to Release configurations we should stay under this space limit.
- I adapted the yaml script to build with Ninja on Windows, in order to fully exploit build parallelism (even though there's only 2 cores).
- ~Slow unit tests were disabled by default in Debug (but not in Release). To run them in Debug on needs to provide an extra CLI flag~ --> I've updated that and added a `[slow]` tag instead. All tests will be run by default, but it's still easy to skip the slow ones on the command-line.
- I also fixed some compilation issues with static MSVC runtime (#1229)
- Daily build is only triggered on the _default branch_. This means we need to merge the .yaml into `master` and tell it to checkout the `dev` in the .yml file, or we can wait until the next "release" of libigl, where we move our `dev` branch to `master. If we merge it as is, no daily build will be triggered.


teseoch:
cool!

i am only worried about not running some tests in debug, but is is an overall small price to pay.

Can the limitations be lifted by paying/begging for more?

jdumas:
Yeah maybe it's not necessary to hide those tests, since Debug configurations are only tested in the daily builds, and they don't take *that long*. My initial intention was to make running `ctest` with default options fast on the command-line. It's still pretty easy to run hidden tests by calling `./libigl_tests [!hide]`... But maybe it's not needed for now.

alecjacobson:
This looks good to go. So I understand correctly does this _replace_ travis and appveyor?

jdumas:
Yes, this replaces both travis and appveyor. The only thing is that in order to enable nightly builds, we need to merge the .yml into the `master` branch, or (easier) make `dev` the _default_ branch for the time being (at least until the hackaton when we tag a proper release).

alecjacobson:
@danielepanozzo should we combine this merge with the switch to:

 move `master` -> `stable` (from which releases will be tagged)
 move `dev`  -> `master`

?

jdumas:
If you want to do that, we can tag a new release as well. The [changelog](https://libigl.github.io/changelog/) that I maintain on the website is also up to date!

PR #1404, opened 2020-01-13
Bug fix in min_quad_with_fixed.cpp

alecjacobson:
This `&& false` somehow appeared... I don't know/recall the origin of it. Current thinking is that it was debris from debugging that got accidentally committed.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
Isn't that codepath disabled because of the comment above?

jdumas:
Following the link in the comment above, it looks like the bug is fixed in https://gitlab.com/libeigen/eigen/issues/751
I'm ok with merging this but we should update the comment accordingly.

PR #1413, opened 2020-01-30
Speed up of SelfIntersectMesh:

bubnikv:
This pull request speeds up the libigl mesh boolean operation. Namely it speeds up the test of whether two triangles sharing an edge share a plane, and if so, whether they overlap thus intersecting.

The original code used predicates of deep expression trees:
A.supporting_plane() != B.supporting_plane() 
which trigger a multi-precision arithmetic code path for sure if the two planes actually overlap or nearly overlap. The code proposed by this pull request replaces the early tests with significantly cheaper variants in case the two triangles are nearly or exactly parallel. For objects with triangulated planar polygons, the change makes the boolean code run 2x faster easily.

I am sure there is a huge room for optimizing the igl::copyleft::cgal::SelfIntersectMesh self-intersection code by fine crafting specific CGAL like filtered predicates and constructs for all operations that may trigger exact arithmetic, as the exact arithmetic is very expensive compared to counting with intervals of doubles. I co-developed a custom mesh boolean library similar to today's CGAL Polygonal Mesh Processing using the concept of CGAL's filtered predicates at one big corporation some 10 years ago, so I have a feeling for what needs to be done, but also I have a feeling for the amount of work required.

Also the parallelization scales badly due to all the memory locks. I have measured a speed up in range of 10% when running on 8 threads compared to a serial code with the locks removed. It would be much more efficient to split the triangle pair intersection loop into a trivially parallellizable filtering phase to filter out the non-intersecting triangle pairs first, and then to calculate the constructs (intersections), maybe with just a serial code.

- [x] This is a minor change.


alecjacobson:
Very cool. We'll need to run a full test @qnzhou

bubnikv:
I forgot to add. I am not sure whether the modified code would handle degenerate triangles correctly, but frankly I am not convinced that this analysis was done on the original code either. I assume that the SelfIntersectMesh receives non-degenerate triangles only.

PR #1417, opened 2020-02-08
Remove png subdirectory

alecjacobson:
stb is [public domain](https://github.com/nothings/stb) and just a few files. Let's put them in the main directory so that we can read/write pngs (and actually lots of other image formats) anywhere inside libigl.

I moved the files previously in `igl/png` to `igl/` or `igl/opengl` correspondingly. This has the side effect of also solving an annoying legacy issue that some files in `igl/png` were (secretly) dependent on `opengl`.

This will (mildly) break existing code. For example, if you code looked like:

```
#include <igl/png/readPNG.h>
...
igl::png::readPNG(...)
```

It will need to be changed to

```
#include <igl/readPNG.h>
...
igl::readPNG(...)
```

You can simply remove `igl::png` from your project's `CMakeLists.txt`

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Hmm there are potential pitfalls with this approach.
1. stb libraries were not written to be header-only. You need to add `STB_IMAGE_IMPLEMENTATION` into exactly one .cpp file to compile function definitions.
2. stb is a very common library, and there can be conflicts when using libigl with other libraries that depend on a different version of stb. This is the reason I made `igl_stb_image.h` in the first place, so that we can compile our own stb dependency with `-fvisibility=hidden` and prevent it from clashing with other dependencies.
3. Our current solution when embedding third-party libs into libigl (e.g. `raytri.c`, `ply.h` etc.) is to make every function `inline` and wrapper into the `igl` namespace to prevent conflict. We could do that for `stb`, but that means more work when updating it...

Unfortunately I don't think there is an easy answer to this, but I'd be happy to discuss options.

alecjacobson:
These all seem to be problems that also apply to the old igl/png directory. You can understand this pull request as only changing the file-directory tree and namespaces.

jdumas:
Not really, the old png directory was pulling stb as an external dependency through CMake, so one had to actually compile `igl_stb_image.cpp` into its own compilation unit. You will notice that the file `igl_stb_image.h` does *not* include `igl_stb_image.cpp`, and that `igl_stb_image.cpp` contains the following lines:
```cpp
#define STB_IMAGE_IMPLEMENTATION
#include "stb_image.h"

#define STB_IMAGE_WRITE_IMPLEMENTATION
#include "stb_image_write.h"
```

Secondly, the stb dependency is compiled as an independent [CMake target](https://github.com/libigl/libigl-stb/blob/master/CMakeLists.txt), and the line:
```cmake
set_target_properties(${PROJECT_NAME} PROPERTIES CXX_VISIBILITY_PRESET hidden)
```
ensures that the `.a` will be compiled with the flag `-fvisibility=hidden` by gcc, thus hiding symbols from outside interference.

All of this happens transparently via CMake. You may say that this makes us depend more on CMake, and this is true, but since we already need CMake to pull and compile external dependencies for subfolders of libigl, this not a big deal.

alecjacobson:
Those files are still there. Whether they got there by committing them or
running a "transparent" cmake script is not related to the 3 problems
listed above. We _could_ use cmake to grab these files (into the main dir).
I see no reason too since it adds yet another build time surprise and stb
is stable.

jdumas:
Moving the files from `external/libigl-stb` to `include/igl` isn't "just" akin to moving the files in the system directories. You're also changing how the user is expected to compile the files into their own project. Previously: it is expected that the user do it via CMake, which compiles the stb functions in a separate compilation unit, and add the flag `-fvisibility=hidden` to GCC when compiling the library. By isolating the symbols into a separate library (`.a/.so`), the symbols are properly hidden from external interference.

By pulling these files directly into `include/igl`, we need to propagate those expectation, i.e. 1) `igl_stb_image.cpp` needs be its own compilation units (and thus libigl is no longer header-only), and 2) the flag `-fvisibility=hidden` needs to be applies to *all* of libigl, which has other problematic implications (need to explicitly export symbols, etc.). If you do not want to propagate those expectation, then we need to make stb_image.h 1) header-only by marking all functions inline, and 2) protect symbols by adding a `namespace igl` around `stb_image.h`.

While `stb` might be relatively stable, it is still actively maintained, e.g. a new version was released just 6 days ago. So if you *really* want to put stb into `include/igl` and mark all functions inline/wrapped in a namespace, tihs means we cannot easily update it anymore.

alecjacobson:
Thanks for explaining. I'm starting to understand the current setup with stb better. More of a mess than I thought. Perhaps now is a good time to invest in a better-suited image IO library. In the past we used libpng (not sure why we switched to stb, maybe the license). I'd be surprised if there isn't a real C++ header only library that can read pngs.

jdumas:
Hmm I don't think it's a problem of the library itself. libpng is a bad choice because it's complicated to compile on its own. But stb is just a single pair of .cpp/.h (embedded in a single .h). There is [lodepng](https://github.com/lvandeve/lodepng) which is also a single pair of .cpp/.h. But really header-only libraries also have their own drawbacks, such as increased compilation times, difficulty to hide symbols/polluting client code, etc. Honestly I think having a clean pair of .cpp/.h is as clean as it gets.

The problem of duplicated symbols/conflicts with different versions of stb (and any other library we choose) is a real pain though. I mostly blame C++ for the lack of proper package management. We made the choice of using CMake as a build system, and using CMake it is possible to abstract those compilation issues (such as hiding symbols in a separate compilation unit).

Honestly things would be simpler if libigl was *not* header-only, or had its own `LIBIGL_IMPLEMENTATION` trick. In this case we could simply wrap stb and other libraries (such as ply/raytri) in the libigl namespace without having to mark every function inline.

PR #1429, opened 2020-02-23
mesh_to_polyhedron need estimated halfedge num

jiangzhongshi:
In CGAL, `cgal/Polyhedron/include/CGAL/Polyhedron_incremental_builder_3.h:583`,
the number of half edges is estimated to be ` h = (std::size_t)((double)(v + f - 2 + 12) * 2.1);`

For example, when V=296, F=732, `h=2179`. However, the real half edge number should be `3*F=2196`. This will result in CGAL crash. 

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/copyleft/cgal/mesh_to_polyhedron.cpp
+++ b/include/igl/copyleft/cgal/mesh_to_polyhedron.cpp
@@ -21,7 +21,7 @@ IGL_INLINE bool igl::copyleft::cgal::mesh_to_polyhedron(
   typedef typename Polyhedron::HalfedgeDS HalfedgeDS;
   // Postcondition: hds is a valid polyhedral surface.
   CGAL::Polyhedron_incremental_builder_3<HalfedgeDS> B(poly.hds());
-  B.begin_surface(V.rows(),F.rows(),3*F.rows());
+  B.begin_surface(V.rows(),F.rows(),3*F.rows()+1);
   typedef typename HalfedgeDS::Vertex   Vertex;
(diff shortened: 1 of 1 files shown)


jdumas:
Why would *not* specifying an estimate result in a crash? Are we using a data structure that doesn't support insertion?

jiangzhongshi:
I don't quite understand your question. The CGAL default (upper bound) estimate is for arbitrary polyhedron meshes, but since we have triangular meshes, it is easier to have a direct estimate.

jdumas:
Well you're saying "This will result in CGAL crash". I'm not super familiar with that part of the code, but my understanding is that even if your estimate is wrong, it should not *crash*. Reading the CGAL documentation seems to mention that if the container used to store the half-edges doesn't support insertions, then indeed underestimating the number of half-edges will result in a crash. This what I was trying to understand here.

jiangzhongshi:
Yes, what you said makes sense. However, I still don't quite understand why in my case CGAL report precondition error as `1+ size_of_halfedge() < capacity_of_halfedge()`. 
Figuring out why CGAL is not expanding should be a better fix, as the estimate also depends on number of border edges.

jdumas:
Do you know what kind of container is used to store the half edges? Is it an Eigen::Matrix? Or a std::vector?

jiangzhongshi:
This file is a templated function, and there are many options for the `polyhedron` class to use (Yes CGAL). In my case, I used `HalfedgeDS_vector`, which internally use `std::vector`

Issue #1423, opened 2020-02-17
csvread reads only first column (or first element of the file). 


AdithyaPediredla:
#### Describe your issue

csvread command reads only the first column (or first element of each line) of the CSV file. 

I have fixed the problem in my local copy by changing lines 33-34 of readCSV.cpp from 
```
while (iss >> a)
     temp.push_back(a);
```
to 
```
char ch;
while (iss >> a >> ch)
     temp.push_back(a);
temp.push_back(a);
```
I can create a pull request with a unit test if you prefer to. 

#### Check all that apply (change to `[x]`)
- [ ] Windows
- [ ] Mac OS X
- [x] Linux
- [ ] I have tried the `dev` branch and the problem persists

See https://libigl.github.io/CONTRIBUTING/#bugreport for more tips.

jdumas:
That's surprising. Yes please create a PR against the `dev` branch. A unit test would be nice but probably not needed, this looks like a simple enough bug.

alecjacobson:
Seems like this got fixed

PR #1430, opened 2020-02-24
Bug fix in csvreads. (Bug #1423)

AdithyaPediredla:

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/readCSV.cpp
+++ b/include/igl/readCSV.cpp
@@ -1,5 +1,5 @@
 // This file is part of libigl, a simple c++ geometry processing library.
-// 
+//
 // Copyright (C) 2013 Alec Jacobson <alecjacobson@gmail.com>
 // 
 // This Source Code Form is subject to the terms of the Mozilla Public License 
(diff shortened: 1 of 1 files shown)


include/igl/readCSV.cpp
```diff
@@ -30,8 +30,10 @@ IGL_INLINE bool igl::readCSV(
     std::istringstream iss(line);
     vector<Scalar> temp;
     Scalar a;
-    while (iss >> a)
+    char ch;
+    while (iss >> a >> ch)
       temp.push_back(a);
+    temp.push_back(a);
```
jdumas: This will not work if the csv is empty.

AdithyaPediredla: Thank you. I have changed line 36 to
`if(iss >> a)
    temp.push_back(a);
`
and merged into the same commit. 

Do I have to make another pull request?

jdumas: No just update your branch and you'll be fine. But I think your fix still doesn't cover all cases. I would change it to the following:
```cpp
while (iss >> a) {
  temp.push_back(a);
  if (!(iss >> ch)) {
    break;
  }
}
```

AdithyaPediredla: Ya, you are right again. I will fix it.

AdithyaPediredla: I fixed it. You can pull now.

PR #1438, opened 2020-03-04
numeric_limits<double>::min() is the minimal positive number

nepluno:
From the context it seems this line should be numeric_limits::lowest() since numeric_limits::min() refers to the minimal positive number (2.22507e-308) in the double range, instead of the lowest number (-1.79769e+308) in the double range. This bug would cause incorrect computation of the bounding box.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


opeca64:
Raymond,

You could very well be right,  but you should be wrong.  While your code
would make more sense, as you say, double precision is actually quite
lossy. Negative numbers used in this context would create significant loss
of precision, given the range you're implying and how double precision
handles subtraction.  I'm guessing a barrier is in place to prevent that.

I'm not well versed in this specific routine,  but it would have to keep
everything positive to optimize both speed and accuracy... thereby
numeric_limit::lowest is most likely intentionally used here, perhaps in
the past,  but nevertheless.

PR #1476, opened 2020-04-11
Bézier curve evaluation and fit cubic Bézier spline to stream of points

alecjacobson:
This is an implementation of "An Algorithm for Automatically Fitting Digitized Curves" by Philip J. Schneider from "Graphics Gems" for arbitrary dimension input (i.e., not just 2D).

- [x] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/fit_cubic_bezier.cpp
```diff
@@ -22,9 +22,12 @@ IGL_INLINE void igl::fit_cubic_bezier(
   // Don't attempt to fit curve to single point
   if(nPts==1) { return; }
   // Avoid using zero tangent
-  const static auto tangent = [&](const int i,const int dir)->Eigen::RowVectorXd
+  const static auto tangent = [](
```
jdumas, line 25: Why do you even need the lambda to be static? It wouldn't make a difference to the compiler.

alecjacobson:
Both your concerns are very valid and problems with this paper. This can be a reference implementation for now (indeed it's a rewrite from Schneider's gem). It'd be cool to handle those and other issues. Not sure if it's research or there're just better methods out there already.

LINJILIANG:
I wanna know how to use the cubics

PR #1480, opened 2020-04-12
Toggle double sided lighting

alecjacobson:
Add an option to display with double sided lighting. Helpful for poorly oriented models (like the truck in our test data) or for meshes that are the mid-surface of thin shape (rather than boundary of solid region).

truck-double-sided

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/opengl/ViewerData.h
+++ b/include/igl/opengl/ViewerData.h
@@ -250,6 +250,7 @@ public:
   unsigned int show_overlay;
   unsigned int show_overlay_depth;
   unsigned int show_texture;
+  unsigned int double_sided;
   unsigned int show_faces;
   unsigned int show_lines;
   bool show_vertid; // shared across viewports for now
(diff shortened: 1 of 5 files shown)


jdumas:
Can you separate those changes from the Bezier PR (#1476)?

include/igl/opengl/ViewerData.h
```diff
@@ -250,6 +250,7 @@ class ViewerData
   unsigned int show_overlay;
   unsigned int show_overlay_depth;
   unsigned int show_texture;
+  unsigned int double_sided;
```
jdumas, line 253: I'd say it should be a `bool` right below the `face_based` toggle. Or do we want to make this a per-viewport property?

alecjacobson: sure go ahead.

PR #1482, opened 2020-04-13
Matcaps (material captures aka environment maps) in Viewer

alecjacobson:
Texture slot can be used as a material capture (also known as environment map). Swapping in a [matcap](https://github.com/nidorx/matcaps) from an image is very simple.

Here's a minimal example

```
#include <igl/png/readPNG.h>

{
  igl::opengl::glfw::Viewer v;
  Eigen::MatrixXd V;
  Eigen::MatrixXi F;
  igl::read_triangle_mesh(argv[1],V,F);
  Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic> R,G,B,A;
  igl::png::readPNG(argv[2],R,G,B,A);
  v.data().set_mesh(V,F);
  v.data().set_texture(R,G,B,A);
  v.data().use_matcap = true;
  v.data().show_lines = false;
  v.launch();
}
```

For example, then
`./main beetle.off Chrome.png` will produce

Screen Shot 2020-04-12 at 8 30 59 PM

[Chrome.png.zip](https://github.com/libigl/libigl/files/4467466/Chrome.png.zip)

**_Note:_** I've disabled the automatic call to  `grid_texture()` in `ViewerData`. I wonder if this was really being used _as is_ like this. I made the judgement call that putting a default matcap in the texture is more useful (and that leaving the `V_uv` empty is more correct). 

related posts: 
 - https://twitter.com/_AlecJacobson/status/1207728511861088256 
 - http://www.alecjacobson.com/weblog/?p=4827
 - https://github.com/alecjacobson/matcap-viewer

### DIFF
--- a/include/igl/opengl/ViewerData.cpp
+++ b/include/igl/opengl/ViewerData.cpp
@@ -30,6 +30,7 @@ IGL_INLINE igl::opengl::ViewerData::ViewerData()
   show_faceid       (false),
   show_labels       (false),
   show_texture      (false),
+  use_matcap        (false),
   point_size(30),
   line_width(0.5f),
   line_color(0,0,0,1),
@@ -78,7 +79,10 @@ IGL_INLINE void igl::opengl::ViewerData::set_mesh(
       Eigen::Vector3d(GOLD_DIFFUSE[0], GOLD_DIFFUSE[1], GOLD_DIFFUSE[2]),
       Eigen::Vector3d(GOLD_SPECULAR[0], GOLD_SPECULAR[1], GOLD_SPECULAR[2]));
 
-    grid_texture();
+    // Alec: is anyone really using this default? It uses trivial UVs and
+    // produces an anisotropic checkerboard...
+    //grid_texture();
+    normal_matcap();
   }
   else
   {
@@ -131,6 +135,7 @@ IGL_INLINE void igl::opengl::ViewerData::copy_options(const ViewerCore &from, co
   to.set(show_overlay      , from.is_set(show_overlay)      );
   to.set(show_overlay_depth, from.is_set(show_overlay_depth));
   to.set(show_texture      , from.is_set(show_texture)      );
+  to.set(use_matcap        , from.is_set(use_matcap)        );
   to.set(show_faces        , from.is_set(show_faces)        );
   to.set(show_lines        , from.is_set(show_lines)        );
 }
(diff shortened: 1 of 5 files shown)


include/igl/opengl/ViewerData.cpp
```diff
@@ -80,7 +81,10 @@ IGL_INLINE void igl::opengl::ViewerData::set_mesh(
       Eigen::Vector3d(GOLD_DIFFUSE[0], GOLD_DIFFUSE[1], GOLD_DIFFUSE[2]),
       Eigen::Vector3d(GOLD_SPECULAR[0], GOLD_SPECULAR[1], GOLD_SPECULAR[2]));
 
-    grid_texture();
+    // Alec: is anyone really using this default? It uses trivial UVs and
+    // produces an anisotropic checkerboard...
+    //grid_texture();
```
jdumas, line 86: Hmm I think this may be confusing to new users. If I enable textures I'd expect to see a (crappy) checkerboard more than a matcap I think?

alecjacobson:
Maybe. Let's deal with that if it ever comes up.

>

jdumas:
Sounds good. I'd still vote for keeping the `grid_texture()` by default though, otherwise lgtm.

alecjacobson:
Are you actually using the *default* grid texture ? (It is not just setting
the checkboard which also seems to be wrong).

jdumas:
Oh right, I had in mind the grid from tutorial 501. It doesn't look like `grid_texture()` is very useful at all, maybe we should remove the method altogether? In that case I think the default behavior should be to not generate any UV/texture at all. Calling `normal_matcap()` by default would change the behavior of tutorial 501, and show the normal texture instead of the default checkerboard, which I think is a bit weird (if I load a mesh with a UV texture I'd rather see the checkerboard atlas than the mesh normal).

Do you think this matcap tick is worth a new tutorial entry btw?

include/igl/opengl/ViewerData.cpp
alecjacobson: I can't imagine a circumstance where I want to see nonsense UVs + checkboard by default. Maybe we can trigger checkerboard texture generation if `set_UV` is called and `use_matcap` is (still) false?

jdumas: As said in my latest answer, I think we can kill the `grid_texture()` function, and as default behavior maybe do nothing? That way = no surprise.

jdumas:
Do you think it's worth a tutorial entry btw?

alecjacobson:
Sure. That's a great idea.

>

PR #1497, opened 2020-04-28
Fix issue with dijkstra algorithm

kekeblom:
On my system (`gcc --version: g++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0`), dijkstra doesn't work in the case below. It doesn't matter what the mesh is, you can test with your favorite one.

```
#include <igl/dijkstra.h>
#include <igl/adjacency_list.h>
#include <iostream>
#include <Eigen/Core>
int main() {
  Eigen::MatrixXd V;
  Eigen::MatrixXi F;
  igl::read_triangle_mesh("gargoyle_cut.obj", V, F);


  std::vector<std::vector<int>> adj_list;
  igl::adjacency_list(F, adj_list, true);

  std::set<int> targets;
  int j = 200;
  targets.insert(j);
  Eigen::VectorXi min_distance, path;
  int success = igl::dijkstra(100, targets, adj_list, min_distance, path);

  std::cout << "success: " << success << " min_distance: " << min_distance[j] << std::endl;
}
```

`std::numeric_limits<DerivedD::Scalar>::infinity()` returns 0 in this case, which leads to the algorithm not finding the target and returning -1. Infinity probably only works with float values.

Before the change `success` would be `-1`. After the change, it returns the index for the target found and `min_distance` is correctly populated. Should probably add a unit test, but did not have time just now to figure out how to run the corresponding test.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


kekeblom:
Added a test case.

jdumas:
Seems it's missing some explicit template instantiation for the static lib version of libigl.

kekeblom:
That's seems to have done the trick.

jdumas:
Awesome, thanks!

PR #1509, opened 2020-05-08
Remove duplicate implementation of is_edge_manifold

alecjacobson:
Just call the (up to 4x) faster one.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
How come one test took so much longer?

"Build / Windows (Release, OFF) (pull_request) Successful in 53m"

jdumas:
Hmm most likely previous compilation results where rotated out of the cache system (which has very limited space). That's one of the limitation of the current setup...

PR #1510, opened 2020-05-08
exploded view of tetrahedral meshes

alecjacobson:
Adds a function `exploded_view` to convert a tet mesh into an exploded surface triangle mesh suitable for [exploded view](https://en.wikipedia.org/wiki/Exploded-view_drawing) visualization.

Also adds a tutorial `719ExplodedView`, where the parameters are mapped to some mouse UI.

Adjusts `ViewerData::set_color` to default to `face_based?` if the #rows in `V` and `F` match.

octopus-exploded-view

- [x] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Ok looking at the log from GitHub Actions:
```
Cache hit occurred on the primary key Windows-Release-OFF-cache, not saving cache.
```
It seems new builds are not being saved in the cache, thus older builds are still loaded which causes the compilation to be slower (since it may be using an old cache that doesn't correspond to the current master).

I certainly did not expect that, and I'm not sure how to fix this yet ... But I'll make a note and think about it :D

PR #1512, opened 2020-05-09
Update heat_geodesics.cpp

### LINKED ISSUE #1511: Heat Geodesic is 2 times what it is supposed to be.
#### Describe the bug

Starting from `tutorial/data/planexy.off` and upsample 4 times. Calling Heat Geodesic from `v=0` gives the following distance,
```
array([0.        , 3.9252696 , 3.9252696 , ..., 3.08298221, 3.06906022,
        3.12934118])
```

whereas the direct solution (the row-wise norm after subtracting the first) is supposed to be
```
array([0.        , 2.        , 2.        , ..., 1.56281247, 1.55686484,
       1.58761072])
```

The issue seems to stem from https://github.com/libigl/libigl/blob/master/include/igl/heat_geodesics.cpp#L155, where there is a `2.0*div_X` and in the original paper it is `\Delta* \phi = divX`

- [x] Windows
- [x] macOS
- [x] Linux

jiangzhongshi:
Fixes #1511.

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/include/igl/heat_geodesics.cpp
+++ b/include/igl/heat_geodesics.cpp
@@ -152,7 +152,7 @@ IGL_INLINE void igl::heat_geodesics_solve(
   }
   const DerivedD div_X = -data.Div*grad_u;
   const DerivedD Beq = (DerivedD(1,1)<<0).finished();
-  igl::min_quad_with_fixed_solve(data.Poisson,(-2.0*div_X).eval(),DerivedD(),Beq,D);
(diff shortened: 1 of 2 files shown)


jdumas:
Cool! Can we add a unit test for that?

jiangzhongshi:
True, I will try this weekend

jiangzhongshi:
I have added a test

PR #1514, opened 2020-05-10
Update remove_duplicate_vertices.h

jiangzhongshi:
Seems to be a typo in the documentation

- [ ] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
Hi @jiangzhongshi . I agree with the last two, but the first line should still be #SV no?

jiangzhongshi:
Hi @jdumas yes you are right, it is indeed very easy to get confused about this function :)

PR #1518, opened 2020-05-14
Added the implementation of Bridson's blue noise sampling

rin-23:

Adding a new function (.h/.cpp files) for blue noise sampling of the n-dimensional space based on https://www.cs.ubc.ca/~rbridson/docs/bridson-siggraph07-poissondisk.pdf

- [x] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/include/igl/remove_duplicate_vertices.h
+++ b/include/igl/remove_duplicate_vertices.h
@@ -20,8 +20,8 @@ namespace igl
   //     this as a tolerance on L1 distance
   // Outputs:
   //   SV  #SV by dim new list of vertex positions
-  //   SVI #V by 1 list of indices so SV = V(SVI,:) 
-  //   SVJ #SV by 1 list of indices so V = SV(SVJ,:)
+  //   SVI #SV by 1 list of indices so SV = V(SVI,:) 
+  //   SVJ #V by 1 list of indices so V = SV(SVJ,:)
   //
   // Example:
   //   % Mesh in (V,F)


alecjacobson:
Great! Can we rename it simply `blue_noise` ?

I have an implementation of this for surfaces that I'll soon merge, too.

rin-23:
renamed to blue_noise

rin-23:
This is a wrapper around Bridson's code so its usable with libigl/Eigen
https://www.cs.ubc.ca/~rbridson/download/curlnoise.tar.gz

include/igl/blue_noise.cpp
```diff
@@ -0,0 +1,390 @@
...
+#include <iostream>
+#include <cassert>
+#include <stdint.h>
+#include <vector>
+#include <Eigen/Core>
+
+//TODO: reimplement bridsons code below using Eigen
+namespace bridson
+{
+    using std::min;
```
jdumas, line 21: Do not alias std functions, prefer using `std::min` directly.

include/igl/blue_noise.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+    template<class T>
+    inline T sqr(const T &x)
+    { return x*x; }
+
+    // transforms even the sequence 0,1,2,3,... into reasonably good random numbers 
+    // challenge: improve on this in speed and "randomness"!
+    inline unsigned int randhash(unsigned int seed)
+    {
+        unsigned int i=(seed^12345391u)*2654435769u;
```
jdumas, line 33: It would be preferable to use C++11 `<random>` generators directly. E.g. look at `igl::randperm`.

include/igl/blue_noise.cpp
rin-23: fixed

PR #1523, opened 2020-05-20
Specify if the color matrix should be applied to the verticies or faces. 

rin-23:
Right now ViewerData::set_colors automatically determines if the color matrix is applied to vertices or faces by checking the number of rows in the color matrix and matching them against the number of rows in V/F matrices. This is fine for typical triangular meshes where the number of vertices and faces is usually different. But in cases where the triangular mesh is the result of splitting tet mesh into 4 triangles, then the number of vertices and faces will be the same (4*#numtets).   In that case ViewerData::set_colors will just default to assuming the color matrix should be applied to vertices even if you want it to be applied to faces.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


rin-23:
Nvm seems like Alec already fixed similar issue.

PR #1538, opened 2020-06-04
Fix compilation error with Eigen 3.3+

bbrrck:
Fixes the compilation error "incompatible operand types" when libigl is compiled with Eigen 3.3+.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/sort.cpp
+++ b/include/igl/sort.cpp
@@ -242,12 +242,12 @@ IGL_INLINE void igl::sort3(
 
   const auto & inner = [&IX,&Y,&dim,&ascending](const Index & i)
   {
-    YScalar & a = (dim==1 ? Y(0,i) : Y(i,0));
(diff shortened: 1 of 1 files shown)


jdumas:
Thanks! Is the error happening with 3.3.7? Libigl is already being compiled with this version so I'm surprised this isn't caught by our CI.

bbrrck:
> Thanks! Is the error happening with 3.3.7? Libigl is already being compiled with this version so I'm surprised this isn't caught by our CI.

Good point. The error does not happen in 3.3.7, I was using a more recent snapshot from December 2019 (specifically [this tree](https://gitlab.com/libeigen/eigen/-/tree/a7c7d329d89e8484be58df6078a586c44523db37)).

jdumas:
Any reason you're not using the stable version (just curious)?

bbrrck:
Yep, `Eigen::all` :smiley:

PR #1541, opened 2020-06-04
Direct Delta Mush implementation

Dxyk:
Implemented Direct Delta Mesh Skinning (Variant 0) from "Direct Delta Mush Skinning and Variants". Added tutorial example (`408_DirectDeltaMush`).

- [x] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/include/igl/direct_delta_mush.cpp
+++ b/include/igl/direct_delta_mush.cpp
@@ -14,6 +14,8 @@
 // 1. U_precomputed, Psi, Omega should be #V by 10 instead of #V by 16
 // 2. Vectorize Psi computation.
 
+bool use_10_float = true;
+
 template <
   typename DerivedV,
   typename DerivedF,
@@ -48,13 +50,20 @@ IGL_INLINE void igl::direct_delta_mush(
   assert(E.cols() == 2 && "E should contain 2 endpoint indices forming bone edges.");
   assert(E.rows() == T.size() && "E.rows() should equal to T.size()");
   assert(Omega.rows() == V.rows() && "Omega contain the same number of rows as V.");
-  assert(Omega.cols() == T.size() * 16 && "Omega should have #T*10 columns.");
+  if (use_10_float)
+  {
+    assert(Omega.cols() == T.size() * 10 && "Omega should have #T*10 columns.");
+  }
+  else
+  {
+    assert(Omega.cols() == T.size() * 16 && "Omega should have #T*16 columns.");
+  }
 
   int n = V.rows();
   int m = T.size();
 
-  Eigen::MatrixXd U_homogeneous(n, 4);
-  U_homogeneous << V, Eigen::VectorXd::Ones(n);
+  Eigen::MatrixXd V_homogeneous(n, 4);
+  V_homogeneous << V, Eigen::VectorXd::Ones(n);
   U.resize(n, 3);
 
   // R matrix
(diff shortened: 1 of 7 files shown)


include/igl/direct_delta_mush.cpp
```diff
@@ -0,0 +1,309 @@
...
+  SparseMatrix<Scalar> D_L_inv = D_L.asDiagonal().toDenseMatrix().sparseView();
+  SparseMatrix<Scalar> L_bar = L * D_L_inv;
+
+  // Implicitly and iteratively solve for W'
+  // w'_{ij} = \sum_{k=1}^{n}{C_{ki} w_{kj}}      where C = (I + kappa L_bar)^{-p}:
+  // W' = C^T \times W  =>  c^T W_k = W_{k-1}     where c = (I + kappa L_bar)
+  // C positive semi-definite => ldlt solver
+  SimplicialLDLT<SparseMatrix<DerivedW>> ldlt_W_prime;
+  SparseMatrix<Scalar> c(I + kappa * L_bar);
+  SparseMatrix<DerivedW> W_prime(W);
```
alecjacobson, line 191: Even if W starts really sparse, W_prime will be effectively dense after the first iteration. Might as well cast to dense here (and then skip the makeCompressed later).

include/igl/direct_delta_mush.cpp
```diff
@@ -0,0 +1,309 @@
...
+  // The following won't work
+  // Matrix<Scalar, Dynamic, Dynamic> Psi(U_prime);
+  // Matrix<Scalar, Dynamic, Dynamic> b((I + lambda * L_bar).transpose());
+  // for (int iter = 0; iter < p; ++iter)
+  // {
+  //   Psi = b.ldlt().solve(Psi);  // hangs here
+  // }
+  // Convert to sparse matrices and compute
+  Matrix<Scalar, Dynamic, Dynamic> Psi;
+  SparseMatrix<Scalar> Psi_sparse = U_prime.sparseView();
```
alecjacobson, line 233: Same story here. Since the solution Psi is dense, use dense right-hand side during the solves. (It's ok to do `ldlt.solve([dense])`

include/igl/direct_delta_mush.h
```diff
@@ -0,0 +1,87 @@
...
+    typename DerivedV,
+    typename DerivedF,
+    typename DerivedC,
+    typename DerivedE,
+    typename DerivedOmega,
+    typename DerivedU>
+  IGL_INLINE void direct_delta_mush(
+    const Eigen::MatrixBase<DerivedV> & V,
+    const Eigen::MatrixBase<DerivedF> & F,
+    const Eigen::MatrixBase<DerivedC> & C,
```
alecjacobson, line 41: Are `C` and `E` used in the end? If not, then let's remove them from (both) prototypes.

tutorial/408_DirectDeltaMush/main.cpp
```diff
@@ -0,0 +1,169 @@
...
+  igl::directed_edge_parents(BE, P);
+  RotationList rest_pose;
+  igl::directed_edge_orientations(C, BE, rest_pose);
+  poses.resize(4, RotationList(4, Quaterniond::Identity()));
+  const Quaterniond twist(AngleAxisd(igl::PI * 0.8, Vector3d(1, 0, 0)));
+  poses[1][2] = rest_pose[2] * twist * rest_pose[2].conjugate();
+  const Quaterniond bend(AngleAxisd(-igl::PI * 0.7, Vector3d(0, 0, 1)));
+  poses[3][2] = rest_pose[2] * bend * rest_pose[2].conjugate();
+
+  igl::readDMAT(TUTORIAL_SHARED_PATH "/arm-weights.dmat", W);
```
alecjacobson, line 141: For this tutorial, it will be more interesting to compare against piecewise-rigid weights. Here's a hotfix to convert these weights to piecewise-rigid 

```  for(int i = 0; i<W.rows();i++)
  {
    const double Wimax = W.row(i).maxCoeff();
    for(int j = 0; j<W.cols();j++)
    {
      W(i,j) = double( W(i,j) == Wimax);
    }
  }
```
I'm also going to find a more interesting example data that we can read in.

include/igl/direct_delta_mush.cpp
```diff
@@ -0,0 +1,304 @@
...
+  {
+    Matrix<Scalar, 4, 4> u_full = V_homogeneous.row(k).transpose() * V_homogeneous.row(k);
+    U_precomputed.row(k) = extract_upper_triangle(u_full);
+  }
+
+  // Psi: #V by #T*10 of \Psi_{ij}s.
+  // Note: this step deviates from the original paper
+  // - The original paper calculates Psi implicitly and iteratively using
+  //   B = (I + lambda * L_bar)^{-p}, similar to the implicit calculation of W' using C (see above).
+  // - I was not able to successfully vectorize Psi, so here I am explicitly computing Psi using
```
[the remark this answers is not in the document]
Dxyk: I think this is on a previous commit. I was confused about having three terms in the summation in the formula for Psi, so I was having trouble figuring out the `X` in `Psi = Btilde^-1 X`.

In the most recent commit (the commit [this comment](https://github.com/libigl/libigl/pull/1541#discussion_r443568119) was on), I was able to figure out how to compute `X` (U_prime).

include/igl/direct_delta_mush.h
Dxyk: They are not used. I will remove them.

Dxyk:
I wasn't able to debug the [build failure](https://github.com/libigl/libigl/pull/1541/checks?check_run_id=797105742) since the log for the build was not displayed on the webpage. Downloading the log archive does not give any meaningful information either. 

I think this might be a timeout issue of some sort? Is there a way to trigger a rebuild manually?

PR #1549, opened 2020-06-18
Text Rendering Using GLSL in ImGui Menu

gmin7:
Want to open the discussion on fine implementation details for the now-functioning text renderer with depth testing through custom shaders.
- Implements a 2d billboard renderer for text labels  
- Uses a monospace text shader pipeline to render the vertid, faceid, and custom labels available currently through the ImGui Menu (tut 106) with inherent depth test.

june18-4

- Handles orthogonal and perspective projections smoothly
june18-2

Some points of discussion:
- The font atlas is being bound in `MeshGL.cpp` which is not ideal because I have not been able to specify a relative path to the font atlas
- There is a lot of redundant code as I created a separate set of VBOs for each label type. I went with this option of binding each label type (vertid, faceid, and custom) independently because it is faster to refresh the VBOs if they are denoted with separate flags as opposed to merging them all into the same VBO.  Perhaps a solution is extracting all this code into a separate module but I'm not sure how that would integrate
- Convention for dirty flag enumeration
- The font size is hardcoded for both orthographic and projection. Perhaps it would be better to compute a ratio as opposed to fine tuned for more robustness to more use cases
- Should we give the user the option to toggle between the old text rendering and this? 

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


gmin7:
Resolved the issue of not being able to locate the font atlas, though it involved the same procedure as was done for `TUTORIAL_SHARED_PATH` and still unsure whether the texture binding should be done in `MeshGL.cpp`

jdumas:
This is awesome!

Regarding the discussion points you raised, unfortunately we need to embed the default font atlas in the source code, we cannot load it from a path on the disk. While this is fine for tutorial data, the libigl viewer needs to be relocatable, and thus the default loading path of a resource cannot be a hardcoded path provided by CMake.

ImGui solves this problem by using `stb_compress` + base85 encoding to convert a ttf font to an embedded C code. The same should work the same for loading and embedding PNG images. Compression code is [here](https://github.com/ocornut/imgui/blob/master/misc/fonts/binary_to_compressed_c.cpp), and decompression code is [here](https://github.com/ocornut/imgui/blob/1a1dcea1a02c4e1542b6e2619111fe2d20e22383/imgui_draw.cpp#L3423), is you want to take a look.

But in our situation, we don't need all that flexibility, so it may be easier to convert the png file to a C header with ImageMagick:
```
convert verasansmono.png verasansmono.h
```
Then edit the file to make it libigl-friendly.

> Should we give the user the option to toggle between the old text rendering and this?

Nah. We should just replace it.

include/igl/opengl/ViewerData.cpp
```diff
@@ -795,4 +797,101 @@ IGL_INLINE void igl::opengl::ViewerData::updateGL(
...
   }
+
+  if (meshgl.dirty & MeshGL::DIRTY_VID_LABELS)
+  {
+    std::string vertCount = std::to_string(V.rows());
+    int totalVertCharacters = 0;
+    int i=1;
+    while(i<vertCount.length())
+    {
+      totalVertCharacters += (i)*9*pow(10,i-1); i++;
```
jdumas, line 808: Why do you need this?

include/igl/opengl/MeshGL.cpp
```diff
@@ -10,6 +10,11 @@
 #include "bind_vertex_attrib_array.h"
 #include "create_shader_program.h"
 #include "destroy_shader_program.h"
+#include "shaders/text.vert"
```
jdumas, line 13: Please include those directly into the source of `MeshGL::init`, as they are not really "headers".

include/igl/opengl/MeshGL.h
```diff
@@ -40,7 +40,10 @@ class MeshGL
     DIRTY_MESH           = 0x00FF,
     DIRTY_OVERLAY_LINES  = 0x0100,
     DIRTY_OVERLAY_POINTS = 0x0200,
-    DIRTY_ALL            = 0x03FF
+    DIRTY_VID_LABELS     = 0x0300,
```
jdumas, line 43: This should go 1, 2, 4, 8, no 3.

include/igl/opengl/ViewerData.cpp
gmin7: I wanted a more efficient way to determine the number of characters in all vertex labels. 
I was trying to find a faster forloop instead of iterating over every single vertex and summing up the string lengths since we know the contents of all the stringe beforehand. I sketched this out on paper and it seemed to work correctly and it's O(log(v)) instead of O(v).

I condensed it to 
```
    int totalVertCharacters = 0;
    for(int i=1; i<vertCount.length(); i++)
      totalVertCharacters += (i)*9*pow(10,i-1);
    totalVertCharacters += (vertCount.length() * (V.rows() - pow(10,vertCount.length()-1))) + 1;
```
and just did a google search and seems a similar algorithm is out there: https://www.geeksforgeeks.org/count-total-number-digits-1-n/ except we shift everything by 1 (0 to n-1). It's a lot cleaner than mine.

Adapting my code to this form:
```
    int totalVertCharacters = 1; // 0th Index
    for(int i=1; i<V.rows(); i*=10)
      totalVertCharacters += (V.rows() - i);
```

Does this sound good to keep? The final versions would be 
```
// Vertices
    int totalVertCharacters = 1; // 0th Index
    for(int i=1; i<V.rows(); i*=10)
    {
      totalVertCharacters += (V.rows() - i);
    }
// Faces
    int totalFaceCharacters = 1; // 0th Index
    for(int i=1; i<F.rows(); i*=10)
    {
      totalFaceCharacters += (F.rows() - i);
    }

```

include/igl/opengl/MeshGL.h
gmin7: Fixed. Out of curiousity is that the numbering scheme?

jdumas: This is a bitset flag, so you can do bitwise operations like `DIRTY_OVERLAY_LINES & DIRTY_OVERLAY_POINTS` for example.

gmin7: Oh I understand, this makes sense thanks :-)

gmin7:
> But in our situation, we don't need all that flexibility, so it may be easier to convert the png file to a C header with ImageMagick:
> 
> ```
> convert verasansmono.png verasansmono.h
> ```
> 
> Then edit the file to make it libigl-friendly.
> 

So I gave this a shot yesterday, and noticed that when ImageMagick converts .png to a header file, as in that command, it chooses GIF file format which I just couldn't get `glTexImage2D` to read properly. I probably could have stripped the header and such from the byte array but ended up dropping this conversion method.

Instead I parsed the data stream output from `stbi_load()` function in `texture_from_png()` 
```
unsigned char *data = igl::stbi_load(png_file.c_str(), &width, &height, &n, 4);
```
with channels set to 1 instead of 4, and throwing it into a text file with the same byte-array structure as the above `convert` method you suggested, changing it to grayscale, and that made it render correctly. To be clear this is a 256x256x1 array of unsigned chars. 

Will push commits now with the changes above, including the modification to the digit label counter we discuss above^

include/igl/opengl/ViewerData.cpp
jdumas: Hmm that seems like an unnecessary complication. You shouldn't be worried about speed, since you're making the same loop later on anyway (from 1 to #V, then for each char in the vertex label).

For simplicity I would produce suggest producing 1 std::vector<std::string> + 1 matrix of label positions for both the vertex ids and the face ids, and reuse the code for the general "labels" to fill the VBOs for vertex/face labels (--> less code duplication).

include/igl/opengl/ViewerData.cpp
```diff
@@ -795,4 +797,101 @@ IGL_INLINE void igl::opengl::ViewerData::updateGL(
...
+      }
+      meshgl.extra_label_pos_vbo.resize(numCustomChars, 3);
+      meshgl.extra_label_char_vbo.resize(numCustomChars, 1);
+      meshgl.extra_label_offset_vbo.resize(numCustomChars, 1);
+      meshgl.extra_label_indices_vbo.resize(numCustomChars, 1);
+      int charCount=0;
+      assert(labels_strings.size() == labels_positions.rows());
+      for(int s=0; s<labels_strings.size(); s++)
+      {
+        std::string temp = labels_strings.at(s);
```
jdumas, line 885: Do not use `std::string` here, since it will copy the data. You can use `const auto & label = ...` if you want.

include/igl/opengl/ViewerData.cpp
```diff
@@ -795,4 +797,101 @@ IGL_INLINE void igl::opengl::ViewerData::updateGL(
...
+        for(int c=0; c<temp.length(); c++)
```
jdumas, line 886: Since it's a STL container both `s` and `c` should be `size_t` and not `int`, but that's a minor detail.

include/igl/opengl/ViewerData.cpp
gmin7: This is a great idea, will do so now thanks :-)

gmin7: Though this being said, that would mean we would need to initialize those data structures in addition to assigning them to the VBOs. Also, taking a closer look though it's ideal to not have so much duplicated code I'm having a hard time seeing how to avoid having 3 copies of the VBO-populating forloops can be replaced with one. Did you mean this as well? If we were to do so I imagine it being a switch case as to which buffer needs to refreshed then dynamically populating just that one, though that restricts us to updating all three buffers in one instance of `updateGL()`

jdumas: You could encapsulate all the VBO machinery for the labels into a separate struct, e.g. call it `TextGL`. Then MeshGL would have 3 members variables:
```cpp
TextGL vertex_labels;
TextGL face_labels;
TextGL custom_labels;
```

That would simplify a lot of code I think. You also need to be careful about the `bind_font_atlas()` function: `igl::png::texture_from_png` is generating the PNG and generating a new texture on the GPU. Right now you're generating a new texture everytime the label data is updated, so you need to store the handle in the TextGL class, and call `glDeleteTextures()` to destroy it. Although, since it's gonna be a texture shared by all meshes, we should probably make it static and use Meyer's singleton pattern to initialize it on the first access (and thus never destroy it... unless we use some global ref counting which may be a bit overkill).

gmin7: I've encapsulated the labels into structs as well as abstracted the functions they are populated in and it reduced a lot of code so that's good :-)

But I'm definitely noticing the effects of `bind_font_atlas`. It's being corruptive when used in conjunction with the overlay lines and points. I'm looking at how other textures are updated in libigl to try to mimic that...

gmin7: Alright, I resolved the aforementioned issue. 
WRT `glDeleteTextures()`, what about storing the font atlas handle in MeshGL.h, like the other texture handle that's there, initialize the font atlas in `MeshGL::init()`, then just bind it each time we render any of the text labels? I've just pushed code which does this, notably `MeshGL::init_font_atlas()` and `MeshGL::bind_font_atlas()`

include/igl/opengl/MeshGL.h
```diff
@@ -81,6 +85,25 @@ class MeshGL
   RowMatrixXf points_V_vbo;
   RowMatrixXf points_V_colors_vbo;
 
+  // Text Rendering
+  typedef struct
```
jdumas, line 89: Don't need typedef (this isn't C). `struct TextGL { ... };` will work just as well.

include/igl/opengl/MeshGL.cpp
```diff
@@ -137,6 +174,37 @@ IGL_INLINE void igl::opengl::MeshGL::bind_overlay_points()
...
+  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+  glTexImage2D(GL_TEXTURE_2D, 0, GL_RED, 256, 256, 0, GL_RED, GL_UNSIGNED_BYTE, verasansmono_font_atlas);
+
+  // TextGL initialization
+  vertex_labels.dirty_flag = MeshGL::DIRTY_VERTEX_LABELS;
+  face_labels.dirty_flag = MeshGL::DIRTY_FACE_LABELS;
+  custom_labels.dirty_flag = MeshGL::DIRTY_CUSTOM_LABELS;
+}
+
+IGL_INLINE void igl::opengl::MeshGL::bind_labels(igl::opengl::MeshGL::TextGL labels)
```
jdumas, line 193: You can simply type `TextGL labels` here, no need to qualify with `igl::opengl::MeshGL::`

include/igl/opengl/MeshGL.cpp
```diff
@@ -163,6 +231,11 @@ IGL_INLINE void igl::opengl::MeshGL::draw_overlay_points()
   glDrawElements(GL_POINTS, points_F_vbo.rows(), GL_UNSIGNED_INT, 0);
 }
 
+IGL_INLINE void igl::opengl::MeshGL::draw_labels(igl::opengl::MeshGL::TextGL& labels)
```
jdumas, line 234: pass by `const &` ?

include/igl/opengl/MeshGL.cpp
```diff
@@ -137,6 +174,37 @@ IGL_INLINE void igl::opengl::MeshGL::bind_overlay_points()
...
+{
+  bool is_dirty = dirty & labels.dirty_flag;
+  glBindTexture(GL_TEXTURE_2D, font_atlas);
+  glBindVertexArray(labels.vao_labels);
+  glUseProgram(shader_text);
+  bind_vertex_attrib_array(shader_text, "position" , labels.vbo_labels_pos       , labels.label_pos_vbo   , is_dirty);
+  bind_vertex_attrib_array(shader_text, "character", labels.vbo_labels_characters, labels.label_char_vbo  , is_dirty);
+  bind_vertex_attrib_array(shader_text, "offset"   , labels.vbo_labels_offset    , labels.label_offset_vbo, is_dirty);
+  glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, labels.vbo_labels_indices);
+  if (is_dirty)
```
jdumas, line 203: Braces around `if()`

include/igl/opengl/MeshGL.cpp
```diff
@@ -46,6 +48,26 @@ IGL_INLINE void igl::opengl::MeshGL::init_buffers()
   glGenBuffers(1, &vbo_points_V);
   glGenBuffers(1, &vbo_points_V_colors);
 
+  // Text Labels
+  glGenVertexArrays(1, &vertex_labels.vao_labels);
```
jdumas, line 52: The `TextGL` class could have methods `init_buffers()` and `free_buffers()` to avoid repetitions here.

include/igl/opengl/MeshGL.h
```diff
@@ -123,6 +146,11 @@ class MeshGL
   /// Draw the currently buffered point overlay
   IGL_INLINE void draw_overlay_points();
 
+  // Text Binding and Draw functions
+  IGL_INLINE void init_text_rendering();
+  IGL_INLINE void bind_labels(TextGL labels);
```
jdumas, line 151: I'm guessing you want to pass by `const &` here, not by value.

include/igl/opengl/ViewerData.cpp
```diff
@@ -562,6 +569,41 @@ IGL_INLINE void igl::opengl::ViewerData::grid_texture()
   dirty |= MeshGL::DIRTY_TEXTURE;
 }
 
+// Populate VBOs of a particular label stype (Vert, Face, Custom)
+IGL_INLINE void igl::opengl::ViewerData::updateLabels(
```
jdumas, line 573: Method name should use `camel_case`

include/igl/opengl/ViewerData.cpp
jdumas: Yes that looks better indeed. For the `init_font_atlas` I guess we can keep it as is for now, that shouldn't be an issue.

jdumas:
> with channels set to 1 instead of 4, and throwing it into a text file with the same byte-array structure as the above convert method you suggested, changing it to grayscale, and that made it render correctly. To be clear this is a 256x256x1 array of unsigned chars.

Hmm the source file size is now 336K, which starts to be a bit heavy for what it is. I guess before we merge this I'd like to take a stab at reducing the source size a little bit more.

gmin7:
I've managed to compress the file to 47K which includes the compressed string and a decompression function I wrote to unpack it. However it is very heuristic, storing pairs of hex values and their continuous frequency. Ie., it's not a universal structure, but from what I was researching all the compression schemes I found were text to bitstreams which were more characters than the initial one (despite being theoretically smaller in size) given header files are in plain text.
Will try to push in a bit, just cleaning up the decompression code.

jdumas:
@gmin7 I made some minor changes I want to push to your branch before I merge this PR. Can you check if you selected the checkbox to give us write access to your fork/branch when you opened the PR?

gmin7:
> @gmin7 I made some minor changes I want to push to your branch before I merge this PR. Can you check if you selected the checkbox to give us write access to your fork/branch when you opened the PR?

Yes, the  "Allow edits and access to secrets by maintainers" checkbox is set

jdumas:
You're right, problem was on my side (using the wrong git url).

PR #1487, opened 2020-04-18
Connected components

alecjacobson:
We previously had a function that computed connected components based on triangle-triangle adjacency (defined as sharing an edge). This PR generalizes this to a routine to compute connected components from any graph (via adjacency matrix). It also adds a routine to build the facet adjacency matrix explicitly, and to do so provides a more memory efficient edge adjacency arrays.   

#### Check all that apply (change to `[x]`)
- [x] All changes meet [libigl style-guidelines](https://libigl.github.io/style-guidelines/).
- [x] Adds new .cpp file.
- [x] Adds corresponding unit test.
- [ ] This is a minor change.

PR #1554, opened 2020-06-28
Decimate performance optimization

alecjacobson:
`decimate` and it's derivatives/dependences/customizations were built around using a `std::set` and `std::vector` of iterators  to implement an key-updatable min-heap.

This refactor replaces that with a `std::priority_queue` (wrapped in `igl::min_heap`) and maintains a list of "insertion time" so that obsolete entries can be skipped when popping.

During this refactor, I also identified repeated calls to `circulation` that could be removed and incorporate the faster `unique_edge_map` from https://github.com/libigl/libigl/pull/1487. 

The result is a ~3x speed up for large meshes.

_One lingering to-do is to template all these collapse/decimate related functions. I'm leaving this as future work for after this merge (since the previous code was also untemplated.)_

- [x] Adds new .cpp file.


include/igl/always_try_never_care.cpp
```diff
@@ -0,0 +1,36 @@
+#include "always_try_never_care.h"
+
+IGL_INLINE void igl::always_try_never_care(
```
jdumas, line 3: `always_try_never_care`? That's a bit vague no?

PR #1560, opened 2020-07-05
Added vector Dirichlet and curved Hessian functionality

odedstein:
Added functions to compute the vector Dirichlet energy (as described [here](http://www.cs.columbia.edu/cg/a-simple-discretization/)) as well as the curved Hessian energy (as described [here](http://www.cs.columbia.edu/cg/curved-hessian/)).

Modified the smoothing tutorial to include new examples for the curved Hessian energy, and added two new tutorials for the vector Dirichlet energy. There will be a corresponding pull request to libigl.github.io with the tutorial updates.

There are unit tests for some of the new functions included.

- [x] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/cr_vector_laplacian.cpp
```diff
@@ -0,0 +1,128 @@
...
+ Eigen::PlainObjectBase<DerivedE>& E,
+ Eigen::PlainObjectBase<DerivedOE>& oE,
+ Eigen::SparseMatrix<ScalarL>& L)
+{
+  if(E.rows()!=F.rows() || E.cols()!=F.cols() || oE.rows()!=F.rows() ||
+   oE.cols()!=F.cols())
+    orient_halfedges(F, E, oE);
+
+  cr_vector_laplacian(V, F,
+    const_cast<const Eigen::PlainObjectBase<DerivedE>& >(E),
```
jdumas, line 50: `const_cast` is not needed to add constness, only to remove it. You can use the following instead:
```cpp
const Eigen::PlainObjectBase<DerivedE>& E_cst(E);
const Eigen::PlainObjectBase<DerivedOE>& oE_cst(oE);
cr_vector_laplacian(E_cst, oE_cst, L);
```

include/igl/cr_vector_laplacian.cpp
```diff
@@ -0,0 +1,128 @@
...
+ const Eigen::MatrixBase<DerivedOE>& oE,
+ Eigen::SparseMatrix<ScalarL>& L)
+{
+  assert(F.cols()==3 && "Faces have three vertices");
+  assert(E.rows()==F.rows() && E.cols()==F.cols() && oE.rows()==F.rows() &&
+   oE.cols()==F.cols() && "Wrong dimension in edge vectors");
+  assert(l_sq.rows()==F.rows() && l_sq.cols()==3 && "l_sq dimensions wrong");
+  assert(dA.size()==F.rows() && "dA dimensions wrong");
+
+  const int m = F.rows();
```
jdumas, line 91: `Eigen::Index` (64bit) rather than `int` (32bit) ?

include/igl/cr_vector_mass.cpp
```diff
@@ -0,0 +1,112 @@
...
+   oE.cols()==F.cols() && "Wrong dimension in edge vectors");
+
+  const int m = F.rows();
+  const int nE = E.maxCoeff() + 1;
+
+  std::vector<Eigen::Triplet<ScalarM> > tripletList;
+  tripletList.reserve(2*3*m);
+  for(int f=0; f<m; ++f) {
+    for(int e=0; e<3; ++e) {
+      const int v1=F(f,(e+1)%3), v2=F(f,(e+2)%3);
```
jdumas, line 96: Don't assume `int`, type should be `auto` or `DerivedF::Scalar`.

include/igl/cr_vector_curvature_correction.cpp
```diff
@@ -0,0 +1,185 @@
...
+  Derivedtheta angleSum = Derivedtheta::Zero(kappa.rows(), 1);
+  for(int i=0; i<F.rows(); ++i)
+    for(int j=0; j<3; ++j)
+      angleSum(F(i,j)) += theta(i,j);
+  const Eigen::Matrix<typename Derivedkappa::Scalar, Eigen::Dynamic, 1>
+  scaledKappa = kappa.array() / angleSum.array();
+  
+  std::vector<Eigen::Triplet<ScalarK> > tripletList;
+  tripletList.reserve(10*3*m);
+  for(int f=0; f<m; ++f) {
```
jdumas, line 143: Don't assume `int`.

include/igl/cr_vector_curvature_correction.cpp
```diff
@@ -0,0 +1,185 @@
...
+  Eigen::Matrix<typename DerivedL_sq::Scalar,Eigen::Dynamic,1> kappa(n);
+  kappa.setZero();
+  for(int i=0; i<F.rows(); ++i)
+    for(int j=0; j<3; ++j)
+      kappa(F(i,j)) -= theta(i,j);
+  kappa.array() += 2 * PI;
+  std::vector<std::vector<int> > b;
+  boundary_loop(F, b);
+  for(const auto& loop : b)
+    for(auto v : loop)
```
jdumas, line 103: Add braces around code.

include/igl/cr_vector_curvature_correction.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include "boundary_loop.h"
+#include "internal_angles.h"
+
+#include "PI.h"
+
+
+template <typename DerivedV, typename DerivedF, typename DerivedE,
+typename DerivedOE, typename ScalarK>
+IGL_INLINE void
+igl::cr_vector_curvature_correction(
```
jdumas, line 25: So we already have `crouzeix_raviart_massmatrix` and `crouzeix_raviart_cotmatrix`. I guess you want to abbreviate that into `cr_` to keep name shorter, but I just want to make sure we're all ok with introducing an inconsistency here?

include/igl/curved_hessian_energy.cpp
```diff
@@ -0,0 +1,126 @@
...
+  cr_vector_mass_intrinsic(F, l_sq, dA,  E, oE, M);
+  scalar_to_cr_vector_gradient_intrinsic(F, l_sq, dA, E, oE, D);
+  cr_vector_laplacian_intrinsic(F, l_sq, dA, E, oE, L);
+  cr_vector_curvature_correction_intrinsic(F, l_sq, E, oE, K);
+
+  //Invert M
+  std::vector<Eigen::Triplet<ScalarQ> > tripletListMi;
+  for(int k=0; k<M.outerSize(); ++k)
+    for(typename Eigen::SparseMatrix<ScalarQ>::InnerIterator it(M,k);
+      it; ++it)
```
jdumas, line 112: Oh gosh please add braces.

include/igl/curved_hessian_energy.cpp
```diff
@@ -0,0 +1,126 @@
...
+  //Matrices that need to be combined
+  Eigen::SparseMatrix<ScalarQ> M, D, L, K;
+  cr_vector_mass_intrinsic(F, l_sq, dA,  E, oE, M);
+  scalar_to_cr_vector_gradient_intrinsic(F, l_sq, dA, E, oE, D);
+  cr_vector_laplacian_intrinsic(F, l_sq, dA, E, oE, L);
+  cr_vector_curvature_correction_intrinsic(F, l_sq, E, oE, K);
+
+  //Invert M
+  std::vector<Eigen::Triplet<ScalarQ> > tripletListMi;
+  for(int k=0; k<M.outerSize(); ++k)
```
jdumas, line 110: Don't assume `int`

include/igl/orient_halfedges.cpp
```diff
@@ -0,0 +1,48 @@
...
+  const Int m = F.rows();
+  
+  DerivedE allE, EE;
+  oriented_facets(F, allE);
+  Eigen::Matrix<Int, Eigen::Dynamic, 1> IA, IC;
+  unique_simplices(allE, EE, IA, IC);
+  
+  E.resize(m, 3);
+  oE.resize(m, 3);
+  for(int f=0; f<m; ++f) {
```
jdumas, line 34: `int`

include/igl/edge_vectors.h
```diff
@@ -0,0 +1,68 @@
...
+//                                          vecPerpendicular or not.
+//
+// Output:
+// vecParallel: normalized edge vectors for each unique edge in E, according
+//              to order of E.
+// vecPerpendicular: tangent unit perpendicular vector to each edge, according
+//                   to orientation in oE, corresponds to each vector in
+//                   vecParallel rotated by pi/2 around an edge-based normal.
+//
+template<bool computePerpendicular=true,
```
jdumas, line 51: I didn't realize giving default value for the *first* template parameters works (I would expect only the last params can be default-valued).

include/igl/edge_vectors.cpp
```diff
@@ -0,0 +1,89 @@
...
+  if(computePerpendicular) //This should ideally be an if constexpr
+    vecPerpendicular.resize(m, 3);
+  for(int i=0; i<E.rows(); ++i) {
+    for(int j=0; j<3; ++j) {
+      if(oE(i,j)<0)
+        continue;
+      const int e=E(i,j);
+      const int vi=F(i,(j+1)%3), vj=F(i,(j+2)%3);
+      vecParallel.row(e) = (V.row(vj)-V.row(vi)).normalized();
+      if(computePerpendicular) //This should ideally be an if constexpr
```
jdumas, line 77: omg please use braces here.

include/igl/edge_vectors.cpp
```diff
@@ -0,0 +1,89 @@
...
+  using Scalar = typename DerivedvecParallel::Scalar;
+  using MatX = Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic>;
+  
+  assert(E.rows()==F.rows() && "E does not match dimensions of F.");
+  assert(oE.rows()==F.rows() && "oE does not match dimensions of F.");
+  assert(E.cols()==3 && F.cols()==3 && oE.cols()==3 &&
+         "This method is for triangle meshes.");
+  assert(F.maxCoeff()<V.rows() && "V does not seem to belong to F.");
+  
+  const int m = E.maxCoeff()+1;
```
jdumas, line 55: `int`

tests/include/igl/cr_vector_curvature_correction.cpp
```diff
@@ -0,0 +1,84 @@
...
...32,0.37024024484653,0,0,0,0,0.370240244846531,0,0,5.55111512312578e-17,0,0,-0.37024024484653,0,0,0,0,0,0.37024024484653,0,0,0,0,0,5.55111512312578e-17,0,0,0,0,0,-0.37024024484653,0,0,0,0,0,0.37024024484653,2.0943951023932,0,0,0,0,0,0,0,0,-5.55111512312578e-17,0.37024024484653,0,0,0,0,0,0,0,0,0,0.37024024484653,0,5.55111512312578e-17,0,0,0,0,0,0.37024024484653,0,0,0,0,0,0,0,0,2.0943951023932,0.37024024484653,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,-0.37024024484653,0,0.37024024484653,0.370240244846531,-0.370240244846531,0,0,0,0,0,0,0,0,0,0,0,0,0,0.37024024484653,2.0943951023932,0.37024024484653,0.370240244846531,0.370240244846531,0,0,0,5.55111512312578e-17,0,0.37024024484653,0,0,0,0,0,0,0,0,-0.37024024484653,0,0,-5.55111512312578e-17,0,0,0,0,0,-0.37024024484653,0,0,0,0,0,0,0,0,0.37024024484653,2.0943951023932,0,0,0,0,0,0,0,0,0,5.55111512312578e-17,-0.370240244846531,0,0,0,0,-5.55111512312578e-17,-0.370240244846531,0,0,0,0,0,0,0,0,0,0,0,-0.370240244846531,0,0,0,0,0,0.370240244846531,0,2.09439510239319,0,0,0,0,0,0,0,0,0,0,0,0,-0.370240244846531,-5.55111512312578e-17,0,0.370240244846531,5.55111512312578e-17,0,0,0,0,0,0,0,0,0,0,0,0,0,0.370240244846531,0,0,0.370240244846531,0,0,2.09439510239319;
```
jdumas, line 28: FYI, if you need to enter binary data in C++ code, you can dump it with `std::cout << std::hexfloat` first (that would avoid conversion to base 10 and back).

tests/include/igl/cr_vector_curvature_correction.cpp
```diff
@@ -0,0 +1,84 @@
...
...1675,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,-0.00909193558897188,0.0286735231074729,0.207182529766611,-0.0106871349843539,-0.00211782359767894,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,-0.0590751134004252,-0.0593097605652173,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0.0741662702935686,0,-0.015276556558624,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,-0.0232198711528178,0.0226137953763438,0,0,0,0,0,0,0,0,0,0,0,0,0,0,-0.0359122143408624,0,-0.0106871349843539,0.145878066851401,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,-0.0172087138133979,0.0322195728921584,0,0,-0.047667325846014,0.00458298218491675,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0.00109753358996168,0.0392015023680103,0,0,-0.0356172183210761,-0.00211782359767894,0,0.247907389511844};
```
jdumas, line 80: Should this be part of https://github.com/libigl/libigl-tests-data ? Seems rather lengthy.

jdumas (changes requested):
Thanks @odedstein for this contribution! This is really cool =). Would you be able to update the tutorial entries in the website once this is merged? It's a shame that we let PR sit around for so long, sorry about that...

Regarding your PR, the code looks pretty clean. There are a couple of minor issues I want to point out:
- You're using `int` explicitly in a couple of places. The code will likely not work on large meshes using 64 bit face indices.
- Please use braces! You have a habit of putting line-breaks in long one-line statement within a for/if, this almost gave me a heart attack.
- Do not use `const_cast` to add constness to a pointer/reference, this is not needed.

odedstein:
Thanks for the feedback. I'm a bit busy at the moment, but I will look at your requested changes and apply them as soon as I have time. The changes all look sensible and doable.

I have a small question about your suggested fix to avoid the const cast. Can you explain this fix in a bit more detail? Does it avoid allocation / calling a constructor?

I do not know if it makes much sense to add a function
`cr_vector_laplacian(
    const Eigen::MatrixBase<DerivedV>& V,
    const Eigen::MatrixBase<DerivedF>& F,
    Eigen::SparseMatrix<ScalarL>& L);`
that does not also return `E` and `oE`, since the indices of the entries of `L` depend on the edge numbering and edge orientation chosen (which can not be inferred from `V,F` alone), and results of computations with `L` are hard to interpret without knowing which index is which edge.

I have made updates to the libigl website tutorials as well (https://github.com/odedstein/libigl.github.io), although I think the tutorials have been changed since, so this might be out of date.

Thanks again!

jdumas:
> I have a small question about your suggested fix to avoid the const cast. Can you explain this fix in a bit more detail? Does it avoid allocation / calling a constructor?

So, the `const_cast` is not needed to "add" const-ness to a `&` or `*`, since you can use implicit type conversion instead. E.g. calling if you have a function `void foo(const int *p)`, then calling `foo(ptr)` on a `int * ptr` will work out of the box, without any need for const-cast. In this case, due to overload/templating you need to call `foo(x)` on an input with the correct type, otherwise the compiler will not be able to deduce and will complain. So to avoid `const_cast`, you simply need to go through an intermediate temporary variable of the right type. You can use `const auto &x(y)` to convert `y` into a const version of itself. Note that `const_cast` can have other [unintended side-effects](https://stackoverflow.com/questions/20956757/using-const-cast-to-add-const-ness-bad-idea) so it a good habit to avoid it if you can.

If your temporary variable is `&` or `*`, then no copy/allocation is performed (it's just a pointer).

> I do not know if it makes much sense to add a function
cr_vector_laplacian( const Eigen::MatrixBase<DerivedV>& V, const Eigen::MatrixBase<DerivedF>& F, Eigen::SparseMatrix<ScalarL>& L);
that does not also return E and oE, since the indices of the entries of L depend on the edge numbering and edge orientation chosen (which can not be inferred from V,F alone), and results of computations with L are hard to interpret without knowing which index is which edge.

I guess I just found it a bit odd to have an argument marketed as "input" that is "computed if not provided". But now that you explain it it's probably fine. Maybe it should be documented as an "output" of the function in the second signature.

> I have made updates to the libigl website tutorials as well (https://github.com/odedstein/libigl.github.io), although I think the tutorials have been changed since, so this might be out of date.

Cool!

odedstein:
Hi,
I have applied the changes you have suggested.
I would be glad if you could look at the tutorials I changed / added (712, 721, 722). I currently only have access to a Mac machine, so I can't be sure if they look correct on Windows / Linux.

jdumas:
Hmm could you also merge your branch with the current master branch? The PR currently shows about ~200 file changed, so that looks a bit suspicious.

odedstein:
> Hmm could you also merge your branch with the current master branch? The PR currently shows about ~200 file changed, so that looks a bit suspicious.

Wow... I have no idea what just happened. I merged libigl, but I must have merged the wrong thing. Sorry for that.

odedstein:
Now it only says 32 files changed, which looks correct, however now all my noisy commit messages that I made during work on your review are present, sorry...

jdumas:
No worries, we can squash them upon merging. I'll take a look a the tutorials on my linux when I get the chance.

jdumas (approved):
Alright, tested the tutorials on macOS and Linux, they look great. Thanks for your contribution @odedstein ! Let me know when you get the chance to update the website and make a PR accordingly!

PR #1565, opened 2020-07-13
Ear clipping function fix

### LINKED ISSUE #1563: igl::predicates::ear_clipping returns too few triangles
#### Describe the bug

igl::predicates::ear_clipping sometimes returns an invalid result (too few triangles).

```
{
  using namespace std;
  const Eigen::MatrixXd P = (Eigen::MatrixXd(6,2)<<
                   0,    -0.212132034355964,
   0.212132034355964,                 0,
   0.106066017177982, 0.106066017177982,
                   0,                 0,
  -0.106066017177982, 0.106066017177982,
  -0.212132034355964,                 0
      ).finished();
  const Eigen::VectorXi RT = Eigen::VectorXi::Zero(P.rows(),1);
  Eigen::VectorXi I;
  Eigen::MatrixXi eF;
  Eigen::MatrixXd nP;
  igl::predicates::ear_clipping(P,RT,I,eF,nP);
  std::cout<<eF<<std::endl;
}
```
will show:
```
5 0 1
5 1 2
3 4 5
```

But there are 6 vertices in the input so there should be 4 output triangles. (also one of these triangles is not a real ear).

Other problems with the code:

2) it apparently assumes the input is CCW oriented, otherwise it produces null/bogus output.
3) if the input is 3 vertices it doesn't return a single triangle (may be related to 2) ).

_Ear clipping is a really simple algorithm. I might just reimplement rather than try to debug this._

- [ ] Windows
- [x] macOS
- [ ] Linux

hankstag:
Fixes #1563 .

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
I'm hesitating on this because in the meantime I also coded a fix. I'd like to merge them and take the best option.

PR #1568, opened 2020-07-15
ImGuizmo Integration

gmin7:
Summary:
- Added tutorial `109_ImGuizmo` that renders a mesh-slicing plane with ImGuizmo's Maya-like manipulation tools 
- This was originally @jdumas 's port to libigl https://github.com/libigl/libigl/issues/692. I turned it into a tutorial, and edited the view matrix passed to ImGuizmo so that the tools do not scale when the user zooms
- Also fixed a small issue with switching from perspective to orthographic projections by another manipulation of the view matrix though it is currently a "magic number" applied as opposed to a proper transformation. 

Testing:
- All combinations of the transformation tools. I find that they behave well when the tools origin is close to the origin of the mesh. The farther it deviates, ImGuizmo is proned to misalign slightly with the slicing plane. This is resolved by moving the tool back close to the mesh.
- Was not tested with two meshes in one view. 

Views of the different tools and toggling projection matrix:

july15-2-min
july15-3-min

- [x] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Why are you closing this?

gmin7:
> Why are you closing this?

I felt I submitted the PR prematurely, without realizing that I did not properly integrate it properly with CMake, so I thought it was best that I close it, fix it, and then reopen the PR. I added a github workflow to debug as opposed to through the PR.

jdumas:
Hmm it would be best to not copy the source code of ImGuizmo directly into the libigl repository. Even if the upstream repository doesn't have a CMakeLists.txt, we can create a target for it in our `libigl.cmake`.

gmin7:
> Hmm it would be best to not copy the source code of ImGuizmo directly into the libigl repository. Even if the upstream repository doesn't have a CMakeLists.txt, we can create a target for it in our `libigl.cmake`.

I completely agree, I was just having some difficulty deciding where the build should occur. Ie., should the ImGuizmo files only be  downloaded and built when that particular tutorial is set to true? I did not see any tutorial specific conditions in libigl.cmake hence my hesitance.

jdumas:
You should add a new entry around [there](https://github.com/libigl/libigl/blob/45cfc79fede992ea3923ded9de3c21d1c4faced1/cmake/libigl.cmake#L398) to download imguizmo and create a target of the same name. Insteaf of doing `add_subdirectory(...)` you should create the target directly using `add_library()`. I think it's safe to assume that when `LIBIGL_WITH_OPENGL_GLFW_IMGUI` is `ON` we should download both `imgui` and `imguizmo`.

cmake/libigl.cmake
```diff
@@ -397,8 +397,12 @@ if(LIBIGL_WITH_OPENGL_GLFW_IMGUI)
     if(NOT TARGET imgui)
       igl_download_imgui()
       add_subdirectory(${LIBIGL_EXTERNAL}/libigl-imgui imgui)
+      igl_download_imguizmo()
+      add_library(igl_opengl_glfw_imguizmo ${LIBIGL_EXTERNAL}/imguizmo/ImGuizmo.cpp ${LIBIGL_EXTERNAL}/imguizmo/ImGuizmo.h)
+      add_library(igl::opengl_glfw_imguizmo ALIAS igl_opengl_glfw_imguizmo)
     endif()
     target_link_libraries(igl_opengl_glfw_imgui ${IGL_SCOPE} igl_opengl_glfw imgui)
+    target_link_libraries(igl_opengl_glfw_imguizmo ${IGL_SCOPE} igl_opengl_glfw imgui)
```
jdumas: To simplify things, I would suggest changing the cmake files as follows:
```cmake
    compile_igl_module("opengl/glfw/imgui")
    if(NOT TARGET imgui)
      igl_download_imgui()
      add_subdirectory(${LIBIGL_EXTERNAL}/libigl-imgui imgui)
    endif()
    if(NOT TARGET imguizmo)
      igl_download_imguizmo()
      add_library(imguizmo ${LIBIGL_EXTERNAL}/imguizmo/ImGuizmo.cpp ${LIBIGL_EXTERNAL}/imguizmo/ImGuizmo.h)
    endif()
    target_link_libraries(igl_opengl_glfw_imgui ${IGL_SCOPE} igl_opengl_glfw imgui imguizmo)
```

gmin7: Thank you for the advice- though for it to build I had to also link imgui to the imguizmo target through a `PUBLIC` interface, sort of like the imgui change we needed to make before.
Also, I noticed that imguizmo imports `malloc.h` which breaks on macOS, and someone noted this relatively recently on their github page https://github.com/CedricGuillemet/ImGuizmo/issues/113 without acknowledgement...

jdumas: Can we use an older SHA1 commit for now? One that doesn't break on macOS.

include/igl/opengl/glfw/imgui/ImGuizmoMenu.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+////////////////////////////////////////////////////////////////////////////////
+#include <imgui/imgui.h>
+#include <imgui/imgui_internal.h>
+#include <imguizmo/ImGuizmo.h>
+#include <Eigen/Dense>
+#include <vector>
+////////////////////////////////////////////////////////////////////////////////
+
+namespace ImGuizmo {
```
jdumas, line 10: Can you put that in the namespace `igl::opengl::glfw::imgui` like the rest of the folder?

PR #1575, opened 2020-07-24
Disabled building SCAF when triangle is not available

vfonov:
When LIBIGL_WITH_TRIANGLE is disabled , SCAF should not be built. 

Added checks in CMakeLists.txt(s) and moved test case for SCAF into a separate folder 

- [ ] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
@jiangzhongshi @alecjacobson  it looks like `scaf.h` should be moved into `igl/triangle` since it depends on `igl::triangle::triangulate` ... How do you suggest we address the issue? Maybe pass the triangulation function as a lambda, similarly to what we do for `igl::delaunay_triangulation` and `igl::copyleft::cgal::delaunay_triangulation` ?

jiangzhongshi:
I think `igl/triangle/` should be the right place to go. I overlooked the structure when first implementing it.

PR #1578, opened 2020-07-27
Added improved .msh reader/writer + tests + example

vfonov:
Improved version of .msh reader/writer that supports loading triangle surface elements and tetrahedral volume elements, and
also physical tags associated with elements. Also, scalar and vector fields defined per node and per element are supported. 

Attached is a binary file for unit tests, to be placed in tests/data/ , result of FEM simulation of currents distribution inside a sphere. 
[MshTestData.zip](https://github.com/libigl/libigl/files/4984168/MshTestData.zip)

- [x] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Cool! Could you submit a PR to [libigl-test-data](https://github.com/libigl/libigl-tests-data) with a *smaller* binary mesh to try the IO in the unit tests?

@qnzhou this looks like your msh loader/saver from PyMesh, maybe you could take a look at this PR.

vfonov:
I changed the tests to use smaller .msh file ( see attached). 
[MshTestData.zip](https://github.com/libigl/libigl/files/5120831/MshTestData.zip)

vfonov:
I created a pull request for test-data: https://github.com/libigl/libigl-tests-data/pull/6

include/igl/readMSH.cpp
```diff
@@ -1,514 +1,209 @@
...
+
+                TetTag(i_tet) = _loader.get_elements_tags()[USETAG][i];
+
+                for(size_t j=0;j<_loader.get_element_fields().size();++j)
+                    for(size_t k=0;k<_loader.get_element_fields_components()[j];++k)
+                        TetF[j](i_tet,k) = _loader.get_element_fields()[j][_loader.get_element_fields_components()[j]*i+k];
+                
+                ++i_tet;
+            } else {
+                // else: it's unsupported type of the element, ignore for now
```
qnzhou, line 144: Perhaps output a warning?

include/igl/writeMSH.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "MshSaver.h"
+#include "MshLoader.h"
+#include <iostream>
+
+
+#if __cplusplus==201103L
+namespace igl {
+    // helper function, unfortunately impossible to make lambda in c++11
+    template <typename T,typename Derived> 
+    void __append_mat_to_vec(std::vector<T> &vec, const Eigen::PlainObjectBase<Derived> & mat)
```
qnzhou, line 19: I think double underscore names are reserved for compiler's internal usage.  This helper function can be declared in anonymous namespace to avoid symbol leaking.

include/igl/writeMSH.cpp
```diff
@@ -0,0 +1,159 @@
...
+        auto __append_mat_to_vec = [] (auto &vec, const auto & mat)-> void 
+        {
+            size_t st = vec.size();
+            vec.resize(st + mat.size());
+            using Scalar = decltype(mat.minCoeff()); // determine storage type of the matrix
+            Eigen::Map< Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> > 
+                _map_vec( reinterpret_cast<Scalar *>( vec.data() + st ), mat.rows(), mat.cols() );
+            _map_vec = mat;
+        };
+        #endif
```
qnzhou, line 59: Not sure if lambda is more preferable over regular template functions even under c++14.

include/igl/writeMSH.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+            assert(_EF.size() == (TriF[i].size()+TetF[i].size()));
+
+            if( TriF[i].cols() == 1 )
+                msh_saver.save_elem_scalar_field(EFields[i], _EF );
+            else if( TriF[i].cols() == 3 )
+                msh_saver.save_elem_vector_field(EFields[i], _EF );
+            else
+            {
+                assert(false); // unsupported type
```
qnzhou, line 150: Output some warning will be helpful.

qnzhou (approved):
Thanks @vfonov !  Only minor comments.  I think at some point we should just refactor msh IO into a stand alone lib so other repo can just depend on it.

include/igl/readMSH.cpp
vfonov: ok, I added a message on stderr

include/igl/writeMSH.cpp
vfonov: ok, using anonymous namespace.

include/igl/writeMSH.cpp
vfonov: made it throw an error.

include/igl/writeMSH.cpp
jdumas: hmm since libigl is a header-only library, maybe using a `namespace internal` might be preferable here.

qnzhou: Good point, perhaps `internal` namespace is more preferable in this case.  Sorry.

vfonov: so, 
```
namespace igl {
 namespace {
   template <typename T,typename Derived> 
    void __append_mat_to_vec(...
...
```
?

jdumas:
```cpp
namespace igl {
namespace internal {

template<...>
void append_mat_to_vec() {
...
}

}
}
```

jdumas:
Hi! So I was about to (finally) merge this, but as I tried out the tutorial entry, I was wondering if the shading is supposed to be like this?

Screen Shot 2020-09-20 at 11 50 03 PM

Wouldn't it be better with flat shading? Also, could you add a `.msh` file to [libigl/libigl-tutorial-data](https://github.com/libigl/libigl-tutorial-data) (different from the test data) so that the tutorial entry doesn't require command-line argument (and we're sure to have the correct data for the tutorial).

vfonov:
yes, probably flat shading will be better. How do I enable that?

jdumas:
Use `ViewerData::set_face_based(true)`

vfonov:
Screenshot of the new version of tutorial code for reading .msh file:
image

jdumas (approved):
Ok I tried it and it looks good. I'll merge this PR once I get write access to merge https://github.com/libigl/libigl-tutorial-data/pull/1

PR #1581, opened 2020-08-04
Fixed writing PLY file, when coordinates are stored as floats

vfonov:
Fixed an error in writePLY when vertex coordinates are stored in single precision.

- [ ] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
@vfonov I tried to push a fix to make the CI pass, but it seems you didn't give us write access to your branch when you opened this PR.

PR #1590, opened 2020-08-18
move scaf to triangle folder

### LINKED ISSUE #1589: "Target not found" error when building tutorial files
#### Describe your question

When trying to build the tutorial files via

```
cd tutorial
mkdir build
cd build
cmake .. -DLIBIGL_WITH_OPENGL=ON -DLIBIGL_WITH_OPENGL_GLFW=ON
```
I get the error

```
CMake Error at 710_SCAF/CMakeLists.txt:4 (add_executable):
  Target "710_SCAF_bin" links to target "igl::triangle" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

...

- [ ] Windows
- [ ] macOS
- [x] Linux

jiangzhongshi:
Fixes #1589 #1575.

The changes move the files for SCAF, and modify the corresponding paths in the tutorial and test file, adding CMake guards for the compilation.

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/triangle/scaf.cpp
```diff
@@ -7,41 +7,41 @@
...
-#include <igl/polar_svd.h>
-#include <igl/slice.h>
-#include <igl/slice_into.h>
-#include <igl/slim.h>
-#include <igl/triangle/triangulate.h>
-#include "mapping_energy_with_jacobians.h"
-
-#include <iostream>
+#include "../PI.h"
+#include "../Timer.h"
```
jdumas, line 18: Hmm you don't have to change those include, `<igl/cat.h>` should still be valid.

jiangzhongshi: Yes, but I looked at other files and thought that this is part of the style.

jiangzhongshi: Do you think I should revert this?

jdumas: It probably doesn't matter much in this case, but using full path (`<igl/Timer.h>` is preferable for public headers, while relative paths should be used to point to private dependencies.

Issue #1492, opened 2020-04-24
_bin suffix really needed?


alecjacobson:
Do we really need the _bin suffix on all of the tutorial executables? similarly in https://github.com/libigl/libigl-example-project

I vaguely remember this was a hack because of name collisions on Windows with cmake. Is it still an issue? Is there a way to fix the name collision in the other way (rename what ever the binary would have collided with)?

This is a minor thing obviously, but if it's also easy to fix it'd make the mac/linux executables more normal looking.

#### Check all that apply (change to `[x]`)
- [x] Windows
- [ ] macOS
- [ ] Linux

jdumas:
I think there is some kind of conflict when an executable target has the same name as a folder or CMake project. I can try it out again on a Windows machine this week-end, but I'm all for killing the `_bin` suffix.

PR #1594, opened 2020-08-22
rm _bin from names and use bin dir instead

alecjacobson:
Fixes https://github.com/libigl/libigl/issues/1492

**_before_** when you did:

```
mkdir build
cd build
cmake ../
make 
```

The tutorial executables had the `_bin` suffix and were mixed in with the CMake flotsam and jetsam, e.g.:

```
build/
  tutorial/
    101_FileIO/
      CMakeFiles
      Makefile
      cmake_install.cmake
    101_FileIO_bin
    ...
```

This change creates `bin/` directory and puts all the tutorial executables in there without a funny suffix:

```
build/
  bin/
    101_FileIO
    102_DrawMesh
    103_Events
    ...
```

No clue, if this affects windows, but I guess it shouldn't (let's see!)


jdumas:
Yes I think it's fine to merge.

PR #1596, opened 2020-08-23
rm strange function, not in use

alecjacobson:
This  barycentric_to_global function does not appear to be in use. It takes as input an usual packing of triangle ids (indices) and barycentric coordinates (floats) into the same matrix.

We could replace it with a prototype such as `barycentric_interpolation(V,F,I,BC)`, for now just delete because it's not in use.


ryanlukejohns:
I just noticed this was removed when trying to build using the most recent release.  This function is useful in combination with the output of `igl::embree::line_mesh_intersection` , for example:

```Eigen::MatrixXd bc = igl::embree::line_mesh_intersection(VP, NP, V, F); // Output: #Vx3 matrix of barycentric coordinates:  id b1 b2.
Eigen::MatrixXd XP = igl::barycentric_to_global(V, F, bc); //convert bc to 3d```

Would this same result be possible with the new `barycentric_interpolation`?  It has some extra inputs I'm not sure I understand...

PR #1597, opened 2020-08-23
Blue noise sampling on triangle meshes

alecjacobson:
Implements Robert Bridson's "Fast Poisson Disk Sampling in Arbitrary Dimensions" but samples are restricted to be on input triangle mesh. (Euclidean distance is still used).

Also adds a `barycentric_interpolation` functions (cf. https://github.com/libigl/libigl/pull/1596) and corresponding test

- [x] Adds new .cpp file.
- [ ] This is a minor change.


alecjacobson:
image
Left: white noise, Right: blue noise 
From tutorial entry

include/igl/blue_noise.cpp
```diff
@@ -0,0 +1,375 @@
...
+        if(Siter !=S.end() && Siter->second < 0)
+        {
+          assert(M.find(nk) != M.end());
+          N.emplace_back(nk);
+        }
+      }
+    }
+        //printf("  --------\n");
+    // randomize order: this might be a little paranoid...
+    std::random_shuffle(std::begin(N), std::end(N));
```
jdumas, line 204: You should use `std::shuffle` rather than `std::random_shuffle`, as the latter rely on `std::rand()` and is unsafe/not deterministic. Unfortunately `igl::random_points_on_mesh()` itself is already not deterministic (relying on Eigen's random function), so it would be hard to get reproducible results for both functions.

include/igl/blue_noise.cpp
```diff
@@ -0,0 +1,375 @@
...
+  // Map 3D subscripts (x,y,z) to unique index (return value)
+  //
+  // Inputs:
+  //   w  side length of w×w×w integer cube lattice
+  //   x  subscript along x direction
+  //   y  subscript along y direction
+  //   z  subscript along z direction
+  // Returns index value
+  //
+  inline BlueNoiseKeyType key(
```
jdumas, line 40: Maybe those functions could be prefixed by `blue_noise_`? E.g. `blue_noise_key`, blue_noise_far_enough`, etc.

include/igl/blue_noise.cpp
```diff
@@ -0,0 +1,375 @@
...
+  typename DerivedP>
+IGL_INLINE void igl::blue_noise(
+    const Eigen::MatrixBase<DerivedV> & V,
+    const Eigen::MatrixBase<DerivedF> & F,
+    const typename DerivedV::Scalar r,
+    Eigen::PlainObjectBase<DerivedB> & B,
+    Eigen::PlainObjectBase<DerivedFI> & FI,
+    Eigen::PlainObjectBase<DerivedP> & P)
+{
+  const auto & tictoc = []() -> double
```
jdumas, line 245: Why not use `igl::Timer` ? Anyhow the `printf` are commented out, so maybe the code can be scrapped out.

include/igl/blue_noise.cpp
```diff
@@ -0,0 +1,375 @@
...
+    std::unordered_map<BlueNoiseKeyType,std::vector<int> > & M,
+    std::unordered_map<BlueNoiseKeyType,int> & S,
+    std::vector<int> & active,
+    std::vector<int> & collected
+    )
+  {
+    //considered.clear();
+    if(active.size() == 0) return false;
+    // random entry
+    const int e = rand() % active.size();
```
jdumas, line 175: I would recommend not using `rand()`, due to reproducibility concerns.

include/igl/blue_noise.cpp
alecjacobson: postponing as per https://github.com/libigl/libigl/issues/1613

include/igl/blue_noise.cpp
alecjacobson: fixed

include/igl/blue_noise.cpp
alecjacobson: postponing as per https://github.com/libigl/libigl/issues/1613

PR #1598, opened 2020-08-24
Fixed behaviour of writePLY to properly treat force_ascii flag

vfonov:
`writePLY(X,V,force_ascii)` was not treating `force_ascii` as expected. 
This commit fixes it. 

- [ ] Adds new .cpp file.
- [x] This is a minor change.


jdumas:
And this is why we shouldn't use booleans in function APIs... @alecjacobson @qnzhou I think it would be beneficial to use an enum in this case, e.g.:

```cpp
enum class FileType {
  Binary,
  Ascii
};

void writePLY(..., FileType::Ascii);
```

jdumas:
Alright I've made a new PR that changes our API to use enums instead of bools to select file encoding: #1619. Closing this one.

PR #1601, opened 2020-09-03
Use dense matrix to improve performance of MIQ with constraints.

marston2yu:
Fixes performance problem of MIQ algorithm with constraints.

Change constraint matrix from sparse matrix to dense matrix to gain performance advantage. The constraint matrix can be large when constraints increase. In a test of fandisk model with 14454 facets and 682 edge constraints, the sparse constraint matrix's size is about 10 million, which leads to 1.5 hours to loop over the matrix (test on MBP 2016 with  2.6GHz quad-core Intel Core i7). As a contrast, the dense one need only about 16 seconds. 

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/copyleft/comiso/miq.cpp
+++ b/include/igl/copyleft/comiso/miq.cpp
@@ -194,6 +194,7 @@ namespace comiso {
     // Matrices
     Eigen::SparseMatrix<double> Lhs;
     Eigen::SparseMatrix<double> Constraints;
+    Eigen::MatrixXd T;
     Eigen::VectorXd rhs;
     Eigen::VectorXd constraints_rhs;
     ///vector of unknowns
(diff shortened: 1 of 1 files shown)


danielepanozzo:
We are planning to remove the MIQ functionality from libigl at the next major release, as it is becoming older/less maintained than the new version that @avaxman is working on for lib directional. @avaxman do you think this PR could be integrated in directional instead?

avaxman:
Yes. Directional does not implement MIQ for fields (there are newer alternatives with the same functionality such as power fields), but there is MIQ integration.

danielepanozzo:
@avaxman Could you please add a link here to the MIQ integration in libdirectional? We can then close this issue and mark MIQ in libigl as deprecated.

avaxman:
The following is currently operational (Master branch) and implements classical MIQ iterative rounding integration:
https://avaxman.github.io/Directional/tutorial/#502-mixed-integer-parametrization

A version that guarantees local-injectivity is working and is underway to replace the above soon:
https://osf.io/s5ack/

Issue #1556, opened 2020-06-29
cmake fails when looking for Eigen3 and libigl


jjcasmar:
I have compiled libigl and installed it in my system. Now I want to use it in a project and that project also uses Eigen3. 
If I do

```
find_package(Eigen3 REQUIRED)
find_package(libigl REQUIRED)
```
I get the following issue

```CMake Error at /usr/local/share/libigl/cmake/libigl-export.cmake:37 (message):
  Some (but not all) targets in this export set were already defined.

  Targets Defined: Eigen3::Eigen

  Targets not yet defined: igl::core;igl::common
```

If I change the order, looking first for libigl and then for Eigen3, works fine, but then compilation doesn't work because the include directories arent apprently well defined. 

```
find_package(libigl REQUIRED)
find_package(Eigen3 REQUIRED)
find_package(OpenMP REQUIRED)
...
target_link_libraries(${PROJECT_NAME} PUBLIC igl::common)
target_link_libraries(${PROJECT_NAME} PUBLIC Eigen3::Eigen)
target_link_libraries(${PROJECT_NAME} PUBLIC OpenMP::OpenMP_CXX)
```

with the following compilation line

```
FAILED: CMakeFiles/PhySim.dir/source/PhySim/Geometry/Meshes/Rod.cpp.o
/usr/bin/c++   -I../include -fPIC   -fopenmp -std=gnu++14 -MD -MT CMakeFiles/PhySim.dir/source/PhySim/Geometry/Meshes/Rod.cpp.o -MF CMakeFiles/PhySim.dir/source/PhySim/Geometry/Meshes/Rod.cpp.o.d -o CMakeFiles/PhySim.dir/source/PhySim/Geometry/Meshes/Rod.cpp.o -c ../source/PhySim/Geometry/Meshes/Rod.cpp
```

I am not sure why this is happening, but clearly is a big issue. Any idea of how to fix it or workaround it?

teseoch:
you can try to remove `find_package(Eigen3 REQUIRED)` and `target_link_libraries(${PROJECT_NAME} PUBLIC Eigen3::Eigen)` since `igl::xxx` carries Eigen3 as dependency.

jdumas:
Do not link against `igl::common`, use `igl::core` instead. Also, if you have any `install()` run in place, try to remove those from your CMake.

jjcasmar:
[...]

PR #1606, opened 2020-09-04
Update libigl.cmake

ShuangLiu1992:
https://github.com/libigl/libigl/issues/1556

give a chance for cmake to pick up user provided Eigen3 through defining -DEigen3_DIR=${usr_path} when configuring the libigl.

Fixes # .

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/cmake/libigl.cmake
+++ b/cmake/libigl.cmake
@@ -103,8 +103,8 @@ endif()
 
 if(HUNTER_ENABLED)
   hunter_add_package(Eigen)
-  find_package(Eigen3 CONFIG REQUIRED)
 endif()
+find_package(Eigen3 CONFIG QUIET)
 
 # Eigen
 if(NOT TARGET Eigen3::Eigen)


jdumas:
If you want to provide your own version of Eigen, you should call `find_package(Eigen3)` yourself **before** including libigl in your project. The call to find_package() should create a target `Eigen3::Eigen`, which will be picked up by libigl. You can even define the target `Eigen3::Eigen` yourself, to prevent libigl from downloading its own.

ShuangLiu1992:
The use case here is this:
1. Cloning the project, compile libigl as a static library , so that later it could be used in another project via find_package(libigl).
2. In step 1 libigl assume there is no Eigen3 since it doesn't have a place for user to provide it. Libigl proceed to download Eigen with Hunter or igl_donwload_eigen()
3. make install libigl, because of step1 Eigen is exported as one of libigl's install target.
4. User creates a project, somewhere before they find_package(libigl), they call find_package(Eigen) or some library (opencv or whatever) they are using needs Eigen and will call find_package(Eigen). Eigen3::Eigen would have been defined by that point which conflicts with the exported libigl's Eigen3 target

jdumas:
I agree that our support for installation is a bit lacking at the moment. However the solution that you propose forces everyone to pick a system-wide installation of Eigen if available, and this is undesirable in general (you might end up using a different version than your colleague for example). I'm not sure what's the best way to deal with this. At the very least we could add a CMake option `LIBIGL_PREFER_SYSTEM_EIGEN` that defaults of `OFF`. But in general I find this is more trouble than it's worth.

jdumas:
Config mode doesn't mean it will not do a system-wide search. It means it will look for a file `Eigen3Config.cmake` (provided by Eigen), rather than rely on an external FindEigen3.cmake that someone else wrote with their feet. The search procedure is described [here](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure).

ShuangLiu1992:
I think these will stop it from looking at the default places.
[NO_DEFAULT_PATH]
[NO_PACKAGE_ROOT_PATH]
[NO_CMAKE_PATH]
[NO_CMAKE_ENVIRONMENT_PATH]
[NO_SYSTEM_ENVIRONMENT_PATH]
[NO_CMAKE_SYSTEM_PATH]

I guess the solution could be either adding something like LIBIGL_PREFER_SYSTEM_EIGEN or disabling searching default places. Although I doubt LIBIGL_PREFER_SYSTEM_EIGEN is what people want, because they probably want to compile libigl with their own version of Eigen but not export it as one of the libigl install target to have conflicts)

Maybe have an option to not export Eigen as a installation target....?
LIBIGL_EXPORT_EIGEN=ON/OFF

jdumas:
I think for every third-party library you need to ask yourself:
1. Do I want to use the system-wide library available on my system?
2. Do I want to let libigl download and use its own version?

This question is only relevant for situations where you want to *install* libigl. Build from source situation are easier to resolve since you can always provide your own `Eigen3::Eigen` target. The problem is that looking for system-wide libs, even if they are available, is **not** a desirable default behavior.

We could have a global CMake option `LIBIGL_PREFER_SYSTEM_LIBS` that seeks system-wide libraries by default seems to me the most reasonable approach to provide a configurable solution to end-users.

ShuangLiu1992:
re point 1.

I believe very few people would want to use system-wide installed library (in most cases).
In practice, when installing Eigen, libigl or opencv whatever, people would change the CMAKE_INSTALL_PREFIX to somewhere else to create some sort of "virtual/sandbox environment", and when a project wants to consume Eigen,libigl, they would add that path to CMAKE_PREFIX_PATH to tell the project to look for these dependencices there instead of going to /usr/local.

PR #1617, opened 2020-09-18
blkdiag function

alecjacobson:
Surprised we didn't already have this.

- [x] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/include/igl/blkdiag.cpp
+++ b/include/igl/blkdiag.cpp
@@ -1,0 +1,77 @@
+// This file is part of libigl, a simple c++ geometry processing library.
+// 
+// Copyright (C) 2020 Alec Jacobson <alecjacobson@gmail.com>
+// 
+// This Source Code Form is subject to the terms of the Mozilla Public License 
+// v. 2.0. If a copy of the MPL was not distributed with this file, You can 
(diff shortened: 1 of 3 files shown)


jdumas:
We're turning libigl into MATLibberish with those function names :sweat_smile:

include/igl/blkdiag.cpp
```diff
@@ -0,0 +1,77 @@
...
+    nr += A.rows();
+    nc += A.cols();
+  }
+  Y.resize(nr,nc);
+  Eigen::VectorXi per_col = Eigen::VectorXi::Zero(nc);
+  for(int j = 0;j<L.size();j++)
+  {
+    per_col(j) = L[j].cols();
+  }
+  Y.reserve(per_col);
```
jdumas, line 29: Hmm, this is wrong. You're reserving a number of nonzero equal to the #cols in each block matrix, which is overkill, but also the indices don't match (`L.size()` is < `nc`).

You should count the number of nnz using `per_col(c + offset) = L[j].innerVector(c).nonZeros();`.

alecjacobson: right. this was wrong. deleted for now.

PR #1619, opened 2020-09-27
Make file encoding option an enum.

jdumas:
Following up on #1598, this updates our public API to use enums rather than bools to chose between binary/ascii file encoding. It's not clear whether `true` means "use ascii" or "use binary" (and we are not consistent in our usage), so hopefully this change will helps smooth things over.


jdumas:
@alecjacobson let me know if you have any opinion on this one.

teseoch (approved):

Issue #1605, opened 2020-09-04
Shader warning


alecjacobson:
```WARNING: Output of vertex shader 'vPosition' not read by geometry shader```

not sure when this started appearing. Are we really using a geometry shader in the viewer now? Why?

#### Check all that apply (change to `[x]`)
- [ ] Windows
- [x] macOS
- [ ] Linux

jdumas:
We're using it to draw 3d text labels (#1549).

PR #1627, opened 2020-10-16
Blindly fixing shader warning

alecjacobson:
Fixes https://github.com/libigl/libigl/issues/1605

I did not test whether text labels still work (assume they do since this just deleted an used variable).


jdumas (approved):
Looks good with vertex labels on my side.

Issue #1437, opened 2020-03-04
Cmake Download of Eigen references deprecated mirror


graphitical:
#### Describe your issue

`LibiglDownloadExternal.cmake` attempts to download Eigen from the [deprecated github mirror](https://github.com/eigenteam/eigen-git-mirror.git) instead of [new gitlab repo](https://gitlab.com/libeigen/eigen) referenced on mirror site.

...

#### Check all that apply (change to `[x]`)
- [x] Windows
- [x] Mac OS X
- [x] Linux
- [ ] I have tried the `dev` branch and the problem persists

See https://libigl.github.io/CONTRIBUTING/#bugreport for more tips.

lxwlzr:
is there a new cmake file out there or should I try to copy-paste it?

jdumas:
Fixed by #1631

PR #1631, opened 2020-10-18
Use new gitlab remote to clone Eigen.

BruegelN:
As [Eigen is now on GitLab.com](http://eigen.tuxfamily.org/index.php?title=News:Eigen_is_now_on_GitLab.com) change path to Eigen's git repository in cmake.
The old github repo is only kept for convenience but does not receive further updates.
Using the repo on gitlab.com it's possible to change to the latest Eigen version 3.3.8 ([ChangeLog](http://eigen.tuxfamily.org/index.php?title=ChangeLog#Eigen_3.3.8) )
Whether updating to Eigen 3.3.8 is a good idea/worth it or not is another story.

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/cmake/LibiglDownloadExternal.cmake
+++ b/cmake/LibiglDownloadExternal.cmake
@@ -67,7 +67,7 @@ endfunction()
 set(LIBIGL_EIGEN_VERSION 3.3.7 CACHE STRING "Default version of Eigen used by libigl.")
 function(igl_download_eigen)
 	igl_download_project(eigen
-		GIT_REPOSITORY https://github.com/eigenteam/eigen-git-mirror.git
+		GIT_REPOSITORY https://gitlab.com/libeigen/eigen.git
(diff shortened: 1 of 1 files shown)


jdumas (approved):
Thanks! I think we should stay compatible with Eigen 3.3.7 for a while longer, but at least it should be easy for the user to use a newer version if they want.

BruegelN:
Agree, @jdumas!
The ChangeLog for Eigen 3.3.8 mentions quite a lot bugfixes but I fully understand the risks of upgrading such an essential component!
 
BTW: fixes #1437

jdumas:
Yeah it's more a question that we should not use newer Eigen features right of the bat, as it might be problematic to integrate into codebases that haven't updated their Eigen version yet. Of course if you're a product using libigl/Eigen you should update if possible, but that part is not up to us.

PR #1638, opened 2020-11-02
Fix bug in readSTL

n-taka:
reset the position of the current character in the input

Fixes # .

Commit 1e905cba3d5e7d82487a6826d0d3fa7cd47de114 introduces bug in readSTL.
- when `is_stl_binary(...)` is called, the position of the current character for `std::istream input` is accidentally updated because of `input.read(...)`. So, I just add `input.seekg(...)` to reset the position.

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/readSTL.cpp
+++ b/include/igl/readSTL.cpp
@@ -48,6 +48,7 @@ IGL_INLINE bool is_stl_binary(std::istream &input) {
   constexpr size_t HEADER_SIZE = 80;
   char header[HEADER_SIZE];
   input.read(header, HEADER_SIZE);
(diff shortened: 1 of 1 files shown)


jdumas (changes requested):
Good catch! But the `seekg` should only be inside the two `if` statements (L53 and L56). In case of binary file you need to continue reading the next 4 bytes for the number of triangles.

n-taka:
@jdumas Thank you for pointing out!
I quickly update the position of `seekg` function call.

jdumas (approved):

PR #1650, opened 2020-11-17
Argument type mistaken in facet_components

jiangzhongshi:
The argument type has an extra `std::vector`

- [ ] Adds new .cpp file.
- [x] This is a minor change.


jdumas (approved):

alecjacobson:
This is _not a mistake_. If you'd like an overload for manifold meshes, then we could add one. The current implementation has this ridiculous triple vector to support non-manifold meshes (faces have edges have lists of face neightbors). See the improved triangle_triangle_adjacency for serialized arrays that are faster.

jiangzhongshi:
Aha, thanks for the explanation, I just took manifoldness for granted.

Issue #1458, opened 2020-03-25
Rendering without a main window


aschier:
#### Describe your issue

Can libigl be used to render a scene without opening a window to render into a PNG file?

I created a simple program to render mesh previews, but currently it still opens a window for a second during rendering.
It would be useful when I could use the Viewer without opening a window at all.

**Minimal example code:**

```cpp
Eigen::MatrixXd V;
Eigen::MatrixXi F;

igl::readOBJ("test.obj", V, F);

igl::opengl::glfw::Viewer viewer;
viewer.core().is_animating = true;
viewer.data().set_mesh(V, F);

Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic> R(1024, 1024);
Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic> G(1024, 1024);
Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic> B(1024, 1024);
Eigen::Matrix<unsigned char,Eigen::Dynamic,Eigen::Dynamic> A(1024, 1024);

viewer.callback_post_draw = [&](igl::opengl::glfw::Viewer &viewer) {
	viewer.core().draw_buffer(viewer.data(), false, R, G, B, A);
	igl::png::writePNG(R, G, B, A, output_filename);
	return true;
};

viewer.core().camera_zoom *= 1.4;

viewer.launch_init();
viewer.launch_rendering(false);
```

Is there a better way to render a scene in the background?

#### Check all that apply (change to `[x]`)
- [ ] Windows
- [ ] Mac OS X
- [x] Linux
- [x] I have tried the `dev` branch and the problem persists

alecjacobson:
Take a look at igl::opengl::glfw::background_window

>

aschier:
I tried

```cpp
igl::opengl::glfw::background_window(viewer.window);
glfwMakeContextCurrent(viewer.window);
for(int i=0;i<viewer.core_list.size(); i++)
{
  viewer.core_list[i].align_camera_center(viewer.data().V,viewer.data().F);
}
viewer.init();
viewer.launch_rendering(false);
```

trying to mimic `launch_init`, but it only renders a transparent image.
Do you have any hint what I may be missing?

alecjacobson:
I can't take a look immediately, but maybe you need to trigger a draw call
or something to do with reshaping...

>

jdumas:
[...]

Issue #601, opened 2017-09-09
Screen capture without showing GUI


Morpheus3000:
Hi,

I was looking at the examples and seems that the only way to capture screenshots is to have the capturing routine as a callback. But I was wondering if it was possible to use libigl to render the mesh in memory and save a capture based on a specific view rotation, lighting etc. without the need to start up the launcher. I ask this because while libigl is a wonderful library that allows me to do almost everything that I can imagine myself needing to, I need to process a large number of models and would ideally run it on a headless server. I could use MATLAB to do the bit, but I would want a purely C++ software without MATLAB as a dependency. Would this be possible?

Thanks!

P.S.: I didn't know where to reach out, so used this instead.

jiangzhongshi:
Just a comment for your reference. 

Basically, you would like to capture the `openGL` content created by `igl::viewer`. 
So in order for it to be "in memory", you need to render `openGL` in background, which can be done through off-screen rendering by [Mesa3D](https://www.mesa3d.org/osmesa.html). An example that I can find is [MeshQuickLookPlugin](https://github.com/alecjacobson/MeshQuickLookPlugin/blob/master/src/render_to_buffer.cpp#L474)

I tried this, but ended up using Blender for rendering in the end.

PR #1652, opened 2020-11-19
Offscreen meshes rendering using embree

vfonov:
Added simple raytracing renderer which allows off-screen rendering of meshes
Using embree (already included in libigl). Addresses #1458, #601
Does not depend on OpenGL.
Tutorial example renders `fertility` example with average gaussian curvature field, in 0.1 sec.
Based on `include/igl/embree/EmbreeIntersector.h` and examples from embree tutorials. 
fertility_curvature

- [X] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/embree/EmbreeRenderer.h
```diff
@@ -0,0 +1,256 @@
...
+      // Inputs:
+      //   origin     3d origin point of ray
+      //   direction  3d (not necessarily normalized) direction vector of ray
+      //   tnear      start of ray segment
+      //   tfar       end of ray segment
+      //   mask      a 32 bit mask to identify active geometries.
+      // Output:
+      //   hit        information about hit
+      // Returns true if and only if there was a hit
+      IGL_INLINE bool intersectRay(
```
jdumas, line 157: rename `interset_ray` for consistency

include/igl/embree/EmbreeRenderer.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+  orthographic = false;
+  
+
+  uC << 1,0,0;
+}
+
+IGL_INLINE igl::embree::EmbreeRenderer::EmbreeRenderer()
+  :
+  //scene(NULL),
```
jdumas, line 77: remove commented code

include/igl/embree/EmbreeRenderer.cpp
```diff
@@ -0,0 +1,457 @@
...
+  geomID(0),
+  initialized(false)
+{
+  init_view();
+}
+
+IGL_INLINE igl::embree::EmbreeRenderer::EmbreeRenderer(
+  const EmbreeRenderer &)
+  :// To make -Weffc++ happy
+  //scene(NULL),
```
jdumas, line 87: remove commented code

include/igl/embree/EmbreeRenderer.cpp
```diff
@@ -0,0 +1,457 @@
...
+  if(C.rows()==V.rows()) // per vertex color
+  {
+    face_based = false;
+    this->C = C.cast<float>();
+    this->uniform_color=false;
+  } else if (C.rows()==F.rows()) {
+    face_based = true;
+    this->C = C.cast<float>();
+    this->uniform_color=false;
+  } else {
```
jdumas, line 391: maybe you could check if there is only 1 row (uniform color), then use that?

include/igl/embree/EmbreeRenderer.cpp
```diff
@@ -0,0 +1,457 @@
...
+// would depend on the template and then we might end up with initializing
+// embree twice. If only there was a way to ask embree if it's already
+// initialized...
+namespace igl
+{
+  namespace embree
+  {
+    // Keeps track of whether the **Global** Embree intersector has been
+    // initialized. This should never been done at the global scope.
+    static RTCDevice g_device = nullptr;
```
jdumas, line 25: We should use a singleton pattern to allocate/deallocate a single device shared by libigl code. For example something like that:

```cpp
class EmrbeeContext
{
public:
    static RTCDevice& device() {
        static EmrbeeContext context;
        return context.m_device;
    }

private:
    EmrbeeContext() {
        m_device = rtcNewDevice();
        if (rtcGetDeviceError (g_device) != RTC_ERROR_NONE) {
            std::cerr << "Embree: An error occurred while initializing embree core!" << std::endl;
        #ifdef IGL_VERBOSE
        } else {
            std::cerr << "Embree: core initialized." << std::endl;
        #endif
        }
    }

    ~RTCDeviceManager() {
        rtcReleaseDevice(m_device);
    }

    RTCDevice m_device;
};
```

jdumas (changes requested):
Hi. This is nice, thanks!

Is there a way to factor out the shared code between `EmbreeIntersector` and `EmbreeRenderer`, maybe through a common base class for example?

include/igl/embree/EmbreeRenderer.h
vfonov: done in new commit

include/igl/embree/EmbreeRenderer.cpp
vfonov: done in new commit

include/igl/embree/EmbreeRenderer.cpp
vfonov: I created EmbreeDevice class with static functions to manage RTCDevice instance

include/igl/embree/EmbreeRenderer.cpp
vfonov: done in new commit

include/igl/embree/EmbreeRenderer.cpp
vfonov: done in new commit

include/igl/embree/EmbreeDevice.h
```diff
@@ -0,0 +1,61 @@
...
+namespace igl
+{
+  namespace embree
+  {
+     // keep track of embree device
+     class EmbreeDevice
+     {
+       private:
+        static RTCDevice g_device;
+        static int       g_cntr;
```
jdumas, line 26: What's the purpose of this counter? All libigl function can share the same device for all we know.

include/igl/embree/EmbreeDevice.cpp
```diff
@@ -0,0 +1,9 @@
+#include "EmbreeDevice.h"
+// Implementation
+
+namespace igl {
+    namespace embree {
+     RTCDevice EmbreeDevice::g_device = nullptr;
```
jdumas, line 6: Using such globals in a libigl library is problematic, since the .cpp can be included in multiple compilation units. I would recommend following the Meyer singleton approach that I suggested last time. Meyer's singleton pattern has the advantage of being safe to initialize from multiple threads.

jdumas (changes requested):
Thanks! But I think there are still some problems with the new EmbreeDevice implementation that you proposed. I've commented on those. Lmk what you think.

include/igl/embree/EmbreeDevice.h
vfonov: to be able to release RTCDevice and destroy completely, when nobody is using it. For example, it is possible to initialize it with different parameters (for whatever reasons).

jdumas: I'm not sure what are the advantages of being able to release the device concretely. In any case, due to the header-only nature of libigl those global variables should be encapsulated in a singleton "DeviceManager" class for linkage reasons.

include/igl/embree/EmbreeDevice.cpp
vfonov: ok, i fixed that

include/igl/embree/EmbreeDevice.h
```diff
@@ -0,0 +1,90 @@
...
+
+namespace igl
+{
+  namespace embree
+  {
+     // keep track of embree device
+     struct EmbreeDevice
+     {
+        RTCDevice embree_device;
+        int       embree_device_cntr;
```
jdumas, line 25: So apparently, `RTCDevide` is already reference-counted (you can use `rtcRetainDevice` to increase its internal counter). However I do not see a way to access the internal ref counter, so I guess there's no way around this approach.

include/igl/embree/EmbreeDevice.cpp
```diff
@@ -0,0 +1,3 @@
+#include "EmbreeDevice.h"
+// Implementation
```
jdumas, line 2: If this file is empty we might as well remove it.

include/igl/embree/EmbreeIntersector.h
```diff
@@ -176,6 +169,8 @@ namespace igl
       Triangle* triangles;
       bool initialized;
 
+      RTCDevice g_device;
```
jdumas, line 172: It is not a global variable, so I'd remove the `g_` prefix. (It'd be more correct to call it `device_handle` but that might be too verbose?).

jdumas (commented):
Hi, sorry about the late review. I think the new singleton class looks good. Too bad we can't use the internal refcounter of the embree device? Anyhow I just have minor comments this time, then we should be good to merge.

jdumas (approved):

jdumas:
Merged, thanks! Let me know if you can add a new tutorial entry for the website.

vfonov:
I will try ...

PR #1666, opened 2020-12-04
copyleft::marching_cubes -> marching_cubes

### LINKED ISSUE #1663: Move (via alternative implementation) marching_cubes out of copyleft
It's a bit frustrating that marching_cubes is in copyleft. We happened to grab the IsoEX LGPL implementation and have been building off of it. That implementation itself likely traces itself to http://paulbourke.net/geometry/polygonise/marchingsource.cpp which is public domain. Skimming through it and ours, I don't see that ours should be especially faster (or otherwise better).

We should use / adapt this public domain version (or find another implementation) and migrate marching_cubes (and the fun other functions that depend on it in copyleft) to the main lib.

alecjacobson:
Fixes #1663 

Reimplementation of marching cubes (or rather a translation from public domain code). Pleasantly, it's also a bit faster for massive grids.

In the meantime, I also implemented marching tets for regular grids, so I also included some functions for regular tetrahedral grids.

714_MarchingTets used to depend on tetgen, now it just uses these regular grids and compares marching_tets to marching_cubes (need to update image in tutorial doc).

And as a consequence, offset_surface and swept_volume have moved out of copyleft/

The copyleft::marching_cubes remains because it has a sparse grid version and a root-finding function-handle version (should be trivial to port to new core version; just did do it yet).

- [x] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Nice! I'll do a review this week-end. Why do this PR include changes to the ImGuizmo plugin, do you need that to update the tutorial?

alecjacobson:
No. That's a mistake. Will fix.

include/igl/marching_cubes.h
```diff
@@ -0,0 +1,52 @@
...
+  //
+  // performs marching cubes reconstruction on a grid defined by values, and
+  // points, and generates a mesh defined by vertices and faces
+  //
+  // Input:
+  //   S   nx*ny*nz list of values at each grid corner
+  //       i.e. S(x + y*xres + z*xres*yres) for corner (x,y,z)
+  //   GV  nx*ny*nz by 3 array of corresponding grid corner vertex locations
+  //   nx  resolutions of the grid in x dimension
+  //   nx  resolutions of the grid in y dimension
```
daviesthomas, line 25: nx, ny, nz.

include/igl/marching_cubes.cpp
```diff
@@ -0,0 +1,451 @@
...
+      v = it->second;
+    }
+    return v;
+  };
+  //fGetOffset finds the approximate point of intersection of the surface
+  // between two points with the values fValue1 and fValue2
+  const auto inv_lerp = [&isovalue](const Scalar & a, const Scalar & b)
+  {
+    const Scalar delta = b-a;
+    if(delta == 0) { return 0.5; }
```
daviesthomas, line 361: return type should be casted to Scalar. My compiler is very unhappy with changing lambda return types 

`if (delta == 0) { return static_cast<Scalar>(0.5); )`

daviesthomas: line below (362) should be intrinsically cast by /delta (which is of type scalar)

include/igl/tetrahedralized_grid.cpp
```diff
@@ -0,0 +1,112 @@
...
+  typename DerivedGV,
+  typename Derivedside,
+  typename DerivedGT>
+IGL_INLINE void igl::tetrahedralized_grid(
+  const Eigen::MatrixBase<DerivedGV> & GV,
+  const Eigen::MatrixBase<Derivedside> & side,
+  const TetrahedralizedGripType type,
+  Eigen::PlainObjectBase<DerivedGT> & GT)
+{
+  const int nx = side(0);
```
jdumas, line 30: We should use `using Index = typedef Derivedside::Scalar` rather than `int`

jdumas (approved):
Ok I think PR looks good (just a minor comment about types).

PR #1673, opened 2020-12-13
writePLY fix

vfonov:
Fixed a bug in writePLY, when face properties were doubled and edge properties were dropped

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


jdumas (approved):
Thanks for the fix!

PR #1674, opened 2020-12-13
Fix nightly build + move EmbreeIntersector impl to .cpp

jdumas:


jdumas:
@alecjacobson you broke the build when you pushed 7250ea7 directly to the master branch ...

PR #1675, opened 2020-12-13
fix mosek to work with 9.2; fix mosek dynamic link via cmake

alecjacobson:
I think mosek is not in the CI tests. So I guess this will just test whether I mucked up the cmakes (don't think I did).

In the long run we should decide about keeping the mosek wrappers in the library. For now, they might as well work with the current mosek version.


tests/CMakeLists.txt
```diff
@@ -52,6 +52,11 @@ if(LIBIGL_WITH_MOSEK)
   target_sources(libigl_tests PRIVATE ${TEST_SRC_FILES} ${TEST_INC_FILES})
 
   target_link_libraries(libigl_tests PUBLIC igl::mosek)
+  # fix mosek
+  IF(APPLE)
+    INCLUDE(../cmake/OSXFixDylibReferences.cmake)
+    OSX_FIX_DYLIB_REFERENCES(libigl_tests "${MOSEK_LIBRARIES}")
```
jdumas, line 58: eeek. this looks like a horrible hack. Do we really need this?

jdumas (changes requested):
Honestly I would vote for dropping wrappers for commercial packages (mosek/matlab) from libigl, since we cannot test them in github actions. I have no idea why you need the horrible hack of `OSXFixDylibReferences.cmake`, and since I don't have access to mosek myself, it'll be difficult to test. In the meantime I'm not sure what we should do with this one.

tests/CMakeLists.txt
alecjacobson: yes.

Issue #1683, opened 2020-12-20
Limit Number of Threads Used


cheng-chi:
#### Describe your feature request
We are currently using libigl on a shared dual-socket research server. Since other people are using the server at the same time, I'm looking for a way to not use all cores available. However, it seems like many algorithms depend on igl/parallel_for.h, which will automatically use all threads available. Is there an easy way to implement something similar to OpenMP's OMP_NUM_THREADS environment variable? I will be glad to help if needed.

#### Check all that apply (change to `[x]`)
- [x] Windows
- [x] macOS
- [x] Linux

jdumas:
Well you can always define IGL_PARALLEL_FOR_FORCE_SERIAL to turn the parallel for into a serial loop:
https://github.com/libigl/libigl/blob/fbfb2d56ad7bab2844433b27f32025314edc60f8/include/igl/parallel_for.h#L130

We'd be happy to review any PR to customize the number of threads. Ideally we should provide ways to let the client override the parallel for implementation with their own (e.g. use OpenMP or TBB rather than the default implementation).

cchi-nuro:
Thanks @jdumas, unfortunately we are mainly using the python binding of libigl, and IGL_PARALLEL_FOR_FORCE_SERIAL is a compile time option. I will try to implement a simple version with get_env (which does not make system call when being used, and hopefully will not cause performance problems.)

jdumas:
Fixed by #1684

PR #1684, opened 2020-12-21
Control number of threads with IGL_PARALLEL_FOR_NUM_THREADS

cheng-chi:
Added environment variable IGL_PARALLEL_FOR_NUM_THREADS to control the number of threads used at runtime. The code is a static initializer and will only run once per process, therefore should have almost zero effect on performance. 
#1683 

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/default_num_threads.cpp
+++ b/include/igl/default_num_threads.cpp
@@ -1,0 +1,54 @@
+// This file is part of libigl, a simple c++ geometry processing library.
+//
+// Copyright (C) 2021 Jérémie Dumas <jeremie.dumas@ens-lyon.org>
+//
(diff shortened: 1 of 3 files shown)


jdumas:
@cheng-chi Sorry for the late review. I've modified your PR to make `igl::default_num_threads()` into a standalone function, and allow for manual user-override without environment variables (if needed). The env variable to determine the number of threads is now `IGL_NUM_THREADS`, and will be used if present in the system.

jdumas (approved):

cheng-chi:
@jdumas Thanks!

PR #1693, opened 2021-01-05
Separated png module from opengl, added unit test for png

vfonov:
Separated png module from opengl, by splitting it into png and png_opengl, so that it's possible to have PNG reading and writing without opengl headers.
Added a unit test for PNG reading and writing.

- [x] Adds new .cpp file.
- [ ] This is a minor change.


cmake/libigl.cmake
```diff
@@ -414,13 +414,16 @@ endif()
...
+  target_link_libraries(igl_png ${IGL_SCOPE} igl_stb_image )
+  
   if(TARGET igl_opengl)
-    if(NOT TARGET stb_image)
-      igl_download_stb()
-      add_subdirectory(${LIBIGL_EXTERNAL}/stb stb_image)
-    endif()
-    compile_igl_module("png" "")
-    target_link_libraries(igl_png ${IGL_SCOPE} igl_stb_image igl_opengl)
+    compile_igl_module("png_opengl" "")
```
jdumas, line 425: Maybe we should have a CMake option `LIBIGL_WITH_PNG_OPENGL` for consistency with other modules.

jdumas (approved):
Nice! I think this is a cleaner solution than #1417.

jdumas:
Maybe we should update the module namespaces though. Following the model of glfw/imgui it _should_ be `igl::png::opengl` (with nested folders). But honestly I'd rather avoid deeply nested namespace. I would prefer something simple like `igl::pngx`, `igl::png_opengl` or similar, but I think @alecjacobson has pretty strong opinions against it.

cmake/libigl.cmake
vfonov: right now if you enable WITH_PNG and WITH_OPENGL you also get png_opengl , so no need to have an extra switch?

alecjacobson:
Thank you for the pull request. (This is a bit orthogonal to https://github.com/libigl/libigl/pull/1417 which moved stb-dependent code into the main lib; and as a side effect freed up the opengl+stb double dependency.)

So we can set aside the debate about nested namespaces / directory layout and merge this PR faster, let's use the current style guide. We can have a separate discussion about whether to use `_`s.

So in this case `include/igl/png_opengl/` should become `include/igl/png/opengl/` and the files inside that folder should have nested namespace `igl::png::opengl` (or `include/igl/opengl/png/` and `igl::opengl::png`, respectively.)

I do wonder whether `png` should be `stb` everywhere. Since this PR will break code, I wonder if we should use it as an opportunity to name this namespace/folder/cmake macro more accurately. The name was originally png when we linked to libpng (IIRC). Most of our functions in `igl::png` secretly work with whatever formats stb understands (jpg, bmp, etc.)

jdumas:
I think a `igl::image` namespace would be nice indeed.

alecjacobson:
image or stb ? I don't think I have a strong preference. They're both
better than png

stb seems good because it refers to the dependency explicitly (like the
others)
image seems good because it's more informative about what functions are in
there

>

jdumas:
Well, maybe one day we'll add tinyexr to support .exr format, or maybe we'll replace stb with something else. So `igl::image` is both more informative and future-proof. I've never been a big fan of having the namespace reflect the third-party dependencies -- we have CMake targets for that.

alecjacobson:
The rationale for namespaces/folders as dependencies is clear. In a user's
C++ code (not buried in the cmake), you know what the dependencies are. So,
while igl::image depending on tinyexr _and_ stb makes sense from a
structure point of view and makes it possible to write a unified
read_image_from_whatever format, the cost is that the C++ code written by
the user of libigl doesn't reveal the dependencies.

It's also more deterministic and explicit. What would we call igl::opengl?
igl::rasterization ?

"more informative"  unfortunately cuts both ways. If we did have
igl::image, then it's awkward if we have code in the core igl:: library
that does image-related things but isn't in igl::image.

igl::image reflects a very different layout of code (i.e., organizing by
topic or functionality) than we currently have (i.e., organized by
3rd-party dependency).

So, maybe the question is (even assuming we'd want to switch to that, which
I'm not sold on) whether making the transition piecemeal is more confusing
than doing it all at once with a major breaking and style-changing reorg
release down the road. If the answer is that it's more confusing to have a
mixture of layouts, then this particular PR can be swiftly merged as
igl::stb and we can discuss major reorg for the future.

>

jdumas:
We could rename modules as follows:
- `igl::opengl` --> `igl::graphics`
- `igl::glfw` --> `igl::viewer` or `igl::application`
- `igl::imgui` --> `igl::ui`, `igl::menu` or `igl::widget`.

I don't think mixing dependencies and functionalities is that confusing when the terms are already pretty widely used (e.g. we could keep `igl::opengl` and `igl::imgui`, while using `igl::viewer`).

Problems with the folder/nested namespace organization is that:
1. It's not clear what happens for shared dependencies (e.g. the viewer depends on the png, glfw and opengl modules)
2. I think a well-documented readme (if you don't like CMake) is just as informative regarding exactly what you are pulling into your own code. The current structure doesn't help with undocumented dependencies anyway (the png module is an example of that).
3. Deeply nested namespaces are very unfriendly to type/use (hello `igl::opengl::glfw::imgui::ImGuiMenu`)
4. I don't think the user needs to be reminded of the dependencies at each include/namespace usage (we have CMake to take care of the admin stuff for us).

If we want to do a major folder restructure, we should probably tag a stable version of libigl beforehand, then progressively move modules in separate commits. If someone finds it confusing, they can stick to the stable version for the duration of the transition.

vfonov:
I would say that if you want to rename png into something that reflects functionality, then it should be something like raster_io or image_io, since all it does is reading/writing images in disk. Then it would be logical to add support for exr and other file formats. Maybe an interface for HDF5 will go there too, one day.

alecjacobson:
Points 1-4 are well taken, there are some finer points to discuss, but the
immediate result is that this conversation has convinced me that in general
this should be a consistent library wide overhaul rather than fixing things
one by one.

So, for now, let's go with igl::stb and keep it in line with the rest of
the org-by-dependency structure.

Since tinyexr and hdf5 are still hypotheticals,  maybe we'll have solved
the reorg by the time they show up. Or maybe they will help create some
initiative to do the reorg.

vfonov:
I renamed png module to stb

jdumas:
Hmm I'm not seeing any filename change as well. Did you forget to commit that?

vfonov:
so, rename read/writePNG into read/writeImage ?

alecjacobson:
I think so. Or maybe write_image ... ahhh this is getting into another
funny legacy inconsistency in how libigl names things. Most of the file IO
things are read[EXTENSION] and write[EXTENSION]. We do have
read_triangle_mesh and write_triangle_mesh which determine the
functionality based on the path. Maybe this one should be write_image /
read_image for the same reason.

>

vfonov:
should I move `igl::png_opengl` into `igl::opengl::image` too ?

jdumas:
If we want to go with the "legacy" style it should be:
- `igl::stb` namespace in `include/igl/stb` folder
- `igl::stb::opengl` namespace in `include/igl/stb/opengl` folder

But honestly if we're gonna rename it again in another future refactoring, maybe we should keep `igl::png` and `igl::png::opengl` for now?

alecjacobson:
let's go with consistency with "legacy" style.

The namespace/directory changes are already code breaking, so at least this
is breaking it toward a consistent style.

>

vfonov:
why not `igl::opengl::stb` ?

alecjacobson:
that's fine to me.
-Alec

jdumas:
fine with me too.

vfonov:
and `read_image` / `write_image` ? We can also keep `readPNG`/`writePNG` for compatibility.

alecjacobson:
I think since anyone using readPNG / writePNG will have to change their
code to get the right includes (+ maybe namespace), we can use write_image
/ read_image

PR #1716, opened 2021-02-08
Dual contouring

alecjacobson:
Implements "Dual Contouring of Hermite Data" by Ju et al. on a dense or sparse regular grid (no octree yet). Optionally can guarantee that the dual vertices lie strictly inside their primary cells (reducing self-intersections), which in turn requires the newly added very fast box-constraint convex quadratic program solver for tiny dense problems.

The 715_MeshImplicitFunction has been updated to a more interesting shape which highlights dual contouring's better crease handling over marching cubes.

dual-contouring-comparison

Dual contouring outputs a pure-quad mesh. To render this nicely, a per_corner_normals overload has been added for polygon meshes.

Includes unit tests for various functions.

- [x] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/min_quad_with_fixed.cpp
```diff
@@ -582,9 +583,330 @@ IGL_INLINE bool igl::min_quad_with_fixed(
...
+    // Huu x(u) = -(fu + H(u,k) bc(k))
+    // x(u) = (-fu + ∑ -Huj bcj)/Huu
+    Eigen::Matrix<Scalar,n,1> x = bc;
+    x(u) = -f(u);
+    for(int i=0;i<k.size();i++){ if(i!=u){ x(u)-=bc(i)*H(i,u); } } 
+    x(u) /= H(u,u);
+    //std::cout<<igl::matlab_format(x,"x")<<std::endl;
+    return x;
+  }
+  // Is there a smart template way to do this?
```
jdumas, line 713: I guess you could do a templated for-loop starting from 16, and dispatching to the appropriate templated function when the argument matches (with a fallback to the dynamic version). Cf this example:
https://gist.github.com/disconnect3d/13c2d035bb31b244df14

include/igl/min_quad_with_fixed.cpp
```diff
@@ -582,9 +583,330 @@ IGL_INLINE bool igl::min_quad_with_fixed(
...
+  typedef typename 
+    std::conditional<Hpd,
+      Eigen::LLT<MatrixSuu>,
+      // LDLT should be faster for indefinite problems but already found some
+      // cases where it was too inaccurate when called via quadprog_primal.
+      // Ideally this function takes LLT,LDLT, or
+      // CompleteOrthogonalDecomposition as a template parameter.  "template
+      // template" parameters did work because LLT,LDLT have different number of
+      // template parameters from CompleteOrthogonalDecomposition.  Perhaps
+      // there's a way to take advantage of LLT and LDLT's default template
```
jdumas, line 874: Why not take an abstract `Solver` class as template argument directly?

include/igl/min_quad_with_fixed.cpp
```diff
@@ -582,9 +583,330 @@ IGL_INLINE bool igl::min_quad_with_fixed(
...
+  typedef Eigen::Matrix<Scalar,ucount,ucount> MatrixSuu;
+  typedef Eigen::Matrix<Scalar,ucount,kcount> MatrixSuk;
+  typedef Eigen::Matrix<Scalar,n,1>      VectorSn;
+  typedef Eigen::Matrix<Scalar,ucount,1> VectorSu;
+  typedef Eigen::Matrix<Scalar,kcount,1> VectorSk;
+  const auto dyn_n = n==Eigen::Dynamic ? H.rows() : n;
+  const auto dyn_kcount = kcount==Eigen::Dynamic ? k.count() : kcount;
+  const auto dyn_ucount = ucount==Eigen::Dynamic ? dyn_n- dyn_kcount : ucount;
+  // For ucount==2 or kcount==2 this calls the coefficient initiliazer rather
+  // than the size initilizer, but I guess that's ok.
```
jdumas, line 829: As long as you're sure that you are going through each entries in the next for loops (seems to be the case, but who knows...).

include/igl/min_quad_with_fixed.cpp
```diff
@@ -582,9 +583,330 @@ IGL_INLINE bool igl::min_quad_with_fixed(
...
+        x(i) = bck(ki);
+        ki++;
+      }else
+      {
+        x(i) = xu(ui);
+        ui++;
+      }
+    }
+  }
+  //std::cout<<igl::matlab_format(x,"x")<<std::endl;
```
jdumas, line 896: Remove debug code.

include/igl/min_quad_with_fixed.cpp
```diff
@@ -582,9 +583,330 @@ IGL_INLINE bool igl::min_quad_with_fixed(
...
...Eigen::StorageOptions)0))), -1, -1> const&, Eigen::Matrix<double, -1, 1, ((Eigen::StorageOptions)0) | ((((-1) == (1)) && ((1) != (1))) ? ((Eigen::StorageOptions)1) : ((((1) == (1)) && ((-1) != (1))) ? ((Eigen::StorageOptions)0) : ((Eigen::StorageOptions)0))), -1, 1> const&, Eigen::Array<bool, -1, 1, ((Eigen::StorageOptions)0) | ((((-1) == (1)) && ((1) != (1))) ? ((Eigen::StorageOptions)1) : ((((1) == (1)) && ((-1) != (1))) ? ((Eigen::StorageOptions)0) : ((Eigen::StorageOptions)0))), -1, 1> const&, Eigen::Matrix<double, -1, 1, ((Eigen::StorageOptions)0) | ((((-1) == (1)) && ((1) != (1))) ? ((Eigen::StorageOptions)1) : ((((1) == (1)) && ((-1) != (1))) ? ((Eigen::StorageOptions)0) : ((Eigen::StorageOptions)0))), -1, 1> const&, Eigen::Matrix<double, -1, -1, ((Eigen::StorageOptions)0) | ((((-1) == (1)) && ((-1) != (1))) ? ((Eigen::StorageOptions)1) : ((((-1) == (1)) && ((-1) != (1))) ? ((Eigen::StorageOptions)0) : ((Eigen::StorageOptions)0))), -1, -1> const&, Eigen::Matrix<double, -1, 1, ((Eigen::StorageOptions)0) | ((((-1) == (1)) && ((1) != (1))) ? ((Eigen::StorageOptions)1) : ((((1) == (1)) && ((-1) != (1))) ? ((Eigen::StorageOptions)0) : ((Eigen::StorageOptions)0))), -1, 1> const&);
```
jdumas, line 903: Oh boy those template instantiations...

include/igl/per_corner_normals.cpp
```diff
@@ -101,9 +101,102 @@ IGL_INLINE void igl::per_corner_normals(
     }
   }
 }
+
+#include "doublearea.h"
```
jdumas, line 105: Move include to top of the file?

include/igl/dual_contouring.cpp
```diff
@@ -0,0 +1,559 @@
...
+        result = (result << 16) + uint16_t(std::get<2>(key));
+        return result;
+      };
+  };
+  template <typename Scalar>
+  class DualContouring
+  {
+    // Types
+    public:
+      typedef Eigen::Matrix<Scalar,1,3> RowVector3S;
```
jdumas, line 36: Small comment, but I find the syntax `using Foo = Bar;` more clear for typedefs (plus it also works with templated typedefs).

include/igl/polygon_corners.h
```diff
@@ -32,6 +32,20 @@ namespace igl
     const std::vector<std::vector<PType> > & P,
     Eigen::PlainObjectBase<DerivedI> & I,
     Eigen::PlainObjectBase<DerivedC> & C);
+  // Convert a pure k-gon list of polygon mesh indices to list of polygon corners
+  // and sizes
+  //
+  // Inputs:
+  //   Q  #Q by k list of k-gon indices (Q(i,j) = -1 means that face i is a
+  //     j-gon)
```
jdumas, line 40: Technically a l-gon with l < j.

include/igl/quadprog.h
```diff
@@ -0,0 +1,87 @@
+#ifndef IGL_QUADPROG_H
+#define IGL_QUADPROG_H
+#include "igl_inline.h"
+#include <Eigen/Core>
+namespace igl
+{
+  // Solve a convex quadratic program. Optimized for small dense problems.
+  // Still works for Eigen::Dynamic (and then everything needs to be Dynamic).
```
jdumas, line 8: So what are the differences between this one and `igl::copyleft::quadprog()`? Can we drop the latter now? How does it compare to [qpmad](https://github.com/asherikov/qpmad)?

include/igl/quadprog.cpp
```diff
@@ -0,0 +1,117 @@
...
+  Arraybn k = Arraybn::Constant(dyn_n,1,false);
+  Eigen::Index iter;
+  // n³ is probably way too conservative. 
+  for(iter = 0;iter<dyn_n*dyn_n*dyn_n;iter++)
+  {
+    // For dual contouring 99% of the time the active set is empty.
+    // Optimize for this common case.
+    // Windows needs template arguments spelled out
+    x = min_quad_with_fixed<Scalar,n,m>(H,f,k,bc,A,b);
+    //std::cout<<igl::matlab_format(x,"x")<<std::endl;
```
jdumas, line 30: Remove debug code.

jdumas (changes requested):
Thanks @alecjacobson, this looks great !

I sprinkled a couple of comments and questions on this PR. The main thing would be to cleanup commented debugging code before we merge. I would also advise to squash this PR when we merge it, there is no point in keeping commits that do not compile on all OSes in the history.

tests/include/igl/quadprog.cpp
```diff
@@ -0,0 +1,137 @@
...
+        -1,
+        -1
+        ).finished();
+    Eigen::Vector3d ub = (Eigen::Vector3d(3,1)<<
+        1,
+        1,
+        1
+        ).finished(); 
+	// Windows needs template args spelled out
+    Eigen::Vector3d x = igl::quadprog<double,3>(H,f,lb,ub);
```
jdumas, line 30: So this particular issue seems similar to this [stack overflow issue](https://stackoverflow.com/questions/58520935/avoiding-ambiguous-parameters-in-templated-eigen-functions-with-msvs). The suggestion  would be to add more template arguments to match the `Eigen::Matrix` type. Looks like a MSVC bug for sure...

include/igl/decimate.cpp
```diff
@@ -185,7 +185,13 @@ IGL_INLINE bool igl::decimate(
       cost_and_placement(e,V,F,E,EMAP,EF,EI,cost,p);
       C.row(e) = p;
       costs(e) = cost;
-    },10000);
+    },
+#ifndef NDEBUG
+    SIZE_MAX
```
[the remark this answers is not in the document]
alecjacobson: This should be superseded by the new way of setting the number of parallel for threads. I wonder if that method should default to serial in debug mode? (obviously also settable via the env variable, but might be nice to set defaults to the common case if we indeed think debugging in serial is what's generally desired).

include/igl/slice_mask.cpp
```diff
@@ -66,7 +66,11 @@ IGL_INLINE void igl::slice_mask(
         {
           if(R(i))
           {
-            Y.row(yi++) = X.row(i);
+            for(int j = 0;j<X.cols();j++)
+            {
+              Y(yi,j) = X(i,j);
```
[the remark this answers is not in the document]
alecjacobson: I believe this is getting around an annoying Eigen static assert. If `Y` is `Eigen::Matrix<double,Eigen::Dynamic,3>` then it will complain if `X` is `Eigen::Matrix<double,Eigen::Dynamic,Eigen::Dynamic>`. An alternative is `Y.row(yi++).head(X.cols())`.

tests/test_common.h
```diff
@@ -24,6 +24,28 @@
...
+struct Catch::StringMaker<std::tuple<int,int,int> > 
+{
+  static std::string convert(std::tuple<int,int,int> const& t)
+  {
+    return 
+      STR("("<<std::get<0>(t)<<","<<std::get<1>(t)<<","<<std::get<2>(t)<<")");
+  }
+};
+template<>
+struct Catch::StringMaker<std::tuple<int,int,double> >
```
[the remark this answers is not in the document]
alecjacobson: I was surprised, too, but I don't know Catch2 very well. Maybe it exists, but I couldn't find it.

include/igl/quadprog.h
alecjacobson: `igl::quadprog` implements a _very_ simple primal active set method. The new `igl::min_quad_with_fixed` is very fast for small dense problems so the iterations of `igl::quadprog` become very fast. Even _if_ it ends up doing many more iterations than `igl::copyleft::quadprog` it would be much faster (in reality it doesn't do _that_ many more iterations). It's a health 10-100x faster than `igl::copyleft::quadprog` for specific cases of QPs.

Unfortunately, that set is limited. `igl::quadprog` is really good at tiny box-constrained QPs with a positive definite objective (like the kind that show up in dual contouring). `igl::copyleft::quadprog` handles more general problems (and also starts to beat `igl::quadprog` when the number of variables gets over ~20). I tried extending `igl::quadprog` so that we could use it for `igl::copyleft::progressive_hulls` and drop `igl::copyleft::quadprog` but it was trickier than I thought. Something like `qpmad` or the non GPL version of `quadrog++` would be good future PR.

tests/include/igl/quadprog.cpp
alecjacobson: This was such a nightmare to debug. At least I figured out how to compile libigl on windows in the process.

include/igl/slice_mask.cpp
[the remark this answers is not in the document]
jdumas: Ah yes. I'm not sure the `Y.row(yi++).head(X.cols())` workaround would work if Y has a fixed column size and X does not, so might as well write the loop.

tests/test_common.h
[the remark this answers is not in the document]
jdumas: Could you try to update the version of Catch2 that we're using in libigl? It's been a while since we've touched.

include/igl/quadprog.h
jdumas: Ok sounds good. It would be good to put your message in comments to the code so that it can serve as future reference.

include/igl/decimate.cpp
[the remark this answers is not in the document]
jdumas: I don't think debugging in serial should be the default, as it would be tricky to debug race conditions that occur due to parallel for loops. But yes if the goal of this ifdef was to run parallel code in serial, then the new env var method should be preferred to set the number of threads.

tests/test_common.h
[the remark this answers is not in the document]
alecjacobson: yes, but let's do it as a separate PR please. I have some more code depending on this PR that I'd like to merge soon.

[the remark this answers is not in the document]
jdumas: Sure.

include/igl/dual_contouring.cpp
```diff
@@ -0,0 +1,559 @@
...
+        const int & x, 
+        const int & y, 
+        const int & z, 
+        const int & o,
+        const RowVector3S & e0,
+        const Scalar & f0)
+      {
+        //e1 computed here needs to precisely agree with e0 when called with
+        //correspond x,y,z. So, don't do this:
+        //Eigen::RowVector3d e1 = e0;
```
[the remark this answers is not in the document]
alecjacobson: the commented code belongs to the comment as a reminder not to do that in case it's revisited later to increase performance etc.

include/igl/dual_contouring.cpp
```diff
@@ -0,0 +1,559 @@
...
+        if(root_finding)
+        {
+          Scalar tl = 0;
+          bool gl = f0>0;
+          Scalar tu = 1;
+          bool gu = f1>0;
+          assert(gu ^ gl);
+          int riter = 0;
+          const int max_riter = 7;
+          while(true)
```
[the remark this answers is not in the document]
alecjacobson: So it's easier to read the loop. Are you thinking to optimize for unrolling?

include/igl/dual_contouring.cpp
```diff
@@ -0,0 +1,559 @@
...
+            vcount[v]++;
+            vH[v] += H;
+            face(k++) = v;
+          }
+        }
+        {
+          const std::lock_guard<std::mutex> lock(Qmut);
+          if(triangles)
+          {
+            if(m+4 >= Q.rows()){ Q.conservativeResize(2*m+4,Q.cols()); }
```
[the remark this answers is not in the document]
alecjacobson: re:conservativeResize This is just amortizing the array allocation as it grows in size. The `2*m` ensures that this will only get called a few times (exponential growth of array --> log(size) number of calls). 

re:mutex, I agree. Using a map reduce would be better here. igl::parallel_for has the hooks to do it (and using tbb will be even easier/more elegant/less overhead; but that's a larger discussion). I didn't do any analysis yet on how much time is being wasted on the mutexs. I'd _guess_ that it's not much the root finding computation/memory access is the bottleneck. Worth looking into in the future. (If the atomic is a quick fix, then let's do it. I'm not confident on using it so I'll leave it for the future).

include/igl/dual_contouring.cpp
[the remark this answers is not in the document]
jdumas: As a rule of thumb I avoid writing any `write(true)` in my code, as it's very easy to run into infinite loop. Explicit for loops are less error-prone.

include/igl/dual_contouring.cpp
```diff
@@ -0,0 +1,559 @@
...
+    // helper function
+    public:
+    // Working variables
+    public: 
+      // f(x) returns >0 outside, <0 inside, and =0 on the surface
+      std::function<Scalar(const RowVector3S &)> f;
+      // f_grad(x) returns (df/dx)/‖df/dx‖ (normalization only important when
+      // f(x) = 0).
+      std::function<RowVector3S(const RowVector3S &)> f_grad;
+      bool constrained;
```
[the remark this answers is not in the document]
alecjacobson: I didn't give it much thought since this is not meant to be exposed to library user's anyway.

include/igl/dual_contouring.cpp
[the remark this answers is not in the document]
jdumas: > The 2*m ensures that this will only get called a few times (exponential growth of array --> log(size) number of calls).

Right. I though that because m grows linearly, the call to resize also had a linear amortized cost. My bad.

include/igl/min_quad_with_fixed.cpp
alecjacobson: whoa. This is cool. The nightmare with windows and the templated code make not want to go near this for a while. I will add your note as a comment for the future.

include/igl/min_quad_with_fixed.cpp
alecjacobson: Augh... I tried but failed to do this. Maybe I wasn't thinking clearly at the time, but I kept running into issues. If you see and obvious way to make it work then maybe we could do it as a follow up PR.

include/igl/min_quad_with_fixed.cpp
alecjacobson: you know, when -1 != 1 ...

include/igl/slice_mask.cpp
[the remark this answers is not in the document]
alecjacobson: oops, it'd be the other way around `Y.row(yi++) = X.row(i).head(Y.cols())`

jdumas (approved):
I think you've addressed most of the comments. We should make sure to keep notes for the items we plan to address in separate PRs if you want to merge it now.

PR #1745, opened 2021-03-08
10-100x speedup on per corner normals + caching adjacency

alecjacobson:
Fixes a few flaws in `igl::per_corner_normals`. Replaces vector<vector<vector>>> style adjacency with serialized adjacency, provides overload with cached adjacency, new function to build this cached adjacency (e.g., at mesh load) so --- for example ---  sharp creases/corners are maintained during smooth deformation (compared to rerunning vanilla igl::per_corner_normals which could cause crease-popping). This alone gave good speedups, adding parallel_for moves it into 100x speedup range on my machine.


include/igl/per_corner_normals.cpp
```diff
@@ -1,108 +1,168 @@
...
+  igl::parallel_for(m*n,[&](const Index ci)
+  {
+    for(int k = CC(ci); k<CC(ci+1); k++)
+    {
+      // add to running sum
+      const auto cfk = CI(k);
+      CN.row(ci) += FN.row(cfk);
+    }
+    // normalize to take average
+    CN.row(ci).normalize();
```
jdumas, line 145: I'd recommend using `stableNormalize()` instead.

include/igl/per_corner_normals.cpp
```diff
@@ -1,108 +1,168 @@
...
 {
-  using namespace Eigen;
-  using namespace std;
+  typedef typename DerivedV::Scalar Scalar;
+  typedef Eigen::Index Index;
+  // unit normals
+  Eigen::Matrix<Scalar,Eigen::Dynamic,3,Eigen::RowMajor> FN(F.rows(),3);
+  // face areas
+  Eigen::Matrix<Scalar,Eigen::Dynamic,1> FA(F.rows());
+  //for(Index f = 0;f<F.rows();f++)
```
jdumas, line 52: Let's remove the commented code. We can turn parallel_for into sequential for loops using the env variable `IGL_NUM_THREADS`.

include/igl/per_corner_normals.cpp
```diff
@@ -1,108 +1,168 @@
...
+template <typename DerivedNV, typename DerivedNF, typename DerivedCN>
+IGL_INLINE void igl::per_corner_normals(
+  const Eigen::MatrixBase<DerivedNV> & NV,
+  const Eigen::MatrixBase<DerivedNF> & NF,
+  Eigen::PlainObjectBase<DerivedCN> & CN)
+{
+  const auto m = NF.rows();
+  const auto nc = NF.cols();
+  CN.resize(m*nc,3);
+  for(Eigen::Index i = 0;i<m;i++)
```
jdumas, line 158: This one can also be made parallel no?

include/igl/smooth_corner_adjacency.cpp
```diff
@@ -0,0 +1,149 @@
...
+  Eigen::VectorXi & CI,
+  Eigen::VectorXi & CC)
+{
+  typedef double Scalar;
+  typedef Eigen::Index Index;
+  Eigen::Matrix<Index,Eigen::Dynamic,1> VF,NI;
+  igl::vertex_triangle_adjacency(F,V.rows(),VF,NI);
+  // unit normals
+  Eigen::Matrix<Scalar,Eigen::Dynamic,3,Eigen::RowMajor> FN(F.rows(),3);
+  //for(Index f = 0;f<F.rows();f++)
```
jdumas, line 21: Remove commented code.

include/igl/smooth_corner_adjacency.cpp
```diff
@@ -0,0 +1,149 @@
...
+  igl::parallel_for(F.rows(),[&](const Index f)
+  {
+    const Eigen::Matrix<Scalar,1,3> v10 = V.row(F(f,1))-V.row(F(f,0));
+    const Eigen::Matrix<Scalar,1,3> v20 = V.row(F(f,2))-V.row(F(f,0));
+    const Eigen::Matrix<Scalar,1,3> n = v10.cross(v20);
+    const Scalar a = n.norm();
+    FN.row(f) = n/a;
```
jdumas, line 28: Let's use `stableNormalized()` again here.

include/igl/smooth_corner_adjacency.cpp
```diff
@@ -0,0 +1,149 @@
...
+          append_CI(nf);
+        }
+      }
+    }
+  }
+  CI.conservativeResize(ncc);
+}
+
+
+void igl::smooth_corner_adjacency(
```
jdumas, line 85: Why not call the other function with a threshold of 0 ? (and using a `>=` comparison).

jdumas (commented):
Nice!
My main recommendation for this PR would be to use `stableNormalize()` to normalize vectors, otherwise this may introduce NaNs for degenerate facets.

I'd be curious to compare speed with the [compute_normal](https://github.com/adobe/lagrange/blob/f56e7d5b27dbfe4ef6f208f6b3ad458dda928a8a/modules/core/include/lagrange/compute_normal.h#L39) function we have in Lagrange -- it should be pretty optimized.

include/igl/smooth_corner_adjacency.cpp
alecjacobson: I don't understand this comment.

jdumas: Well you've basically implemented the function `smooth_corner_adjacency` twice. But the variant without the angle threshold can simply call the first function with `corner_threshold_radians == -M_PI`, provided you change the condition `if(dp > cos_thresh)` to a `if(dp >= cos_thresh * fnhat.norm() * ifn.norm())`. Well due to floating point error that might not be entirely correct, but you could use a lambda to implement the generic function instead.

alecjacobson: Sorry for the confusion, Jeremie. These two functions are not related that way.

I'll fix the header to document the second overload. It is purely combinatorial based on indexed normals (e.g., coming from a .obj).

jdumas: Oh ok, that makes sense.

alecjacobson:
https://github.com/libigl/libigl/issues/1785 is causing macos to fail. It was passing before and passes locally. I'm moving on.

PR #1746, opened 2021-03-08
Split min_quad_with_fixed implementation into separate cpp.

jdumas:
Nightly job fails because MSVC runs out of heap space and chokes on `min_quad_with_fixed.cpp`. Hopefully by splitting the implementation into separate .cpp this should alleviate the problem.

Also, our label Static/HeaderOnly was apparently flipped...


jdumas:
> Maybe we should have a special indicator in the filename for the .cpp that just contain template instantiations.

I'm using `.impl.h` for this one, so it can be grepped by the same regex as other .h in the project. We could consider other file extensions such as `.tpp` or `.ipp`, but imho those tend to interfere with syntax highlighting in code editors..

jdumas:
@alecjacobson I'm gonna merge this one soon, unless you have a strong opinion on the file extension.

PR #1767, opened 2021-03-28
Fix the order of principal curvature

lchaooo:
principle_curvature returns PD1, PV1 as minimal curvature direction and value, PD2, PV2 as maximum curvature direction and value.
The order is not the same as described in the docs(https://libigl.github.io/libigl-python-bindings/igl_docs/#principal_curvature) and the comments in the header file:
https://github.com/libigl/libigl/blob/36e8435a2f724e83e14f79d128102d06b514a4f4/include/igl/principal_curvature.h#L34-L38

Here's the code (using python bindings) and screenshot to demonstrate this bug:
```python
v, f = igl.cylinder(20, 20)
# pd1, pv1: maximal
# pd2, pv2: minimal
pd1, pd2, pv1, pv2 = igl.principal_curvature(v, f)
print(np.all(pv1<=pv2)) # printing true means pv2 is actually maximal curvature value

p = meshplot.plot(v, f, shading={"wireframe": False}, return_plot=True)
avg = igl.avg_edge_length(v, f) / 2
# red segments should be parallel to the maximal curvature direction
p.add_lines(v + pd1 * avg, v - pd1 * avg, shading={"line_color": "red"})
# blue segments should be parallel to the minimal curvature direction
p.add_lines(v + pd2 * avg, v - pd2 * avg, shading={"line_color": "blue"})
```
Screen Shot 2021-03-28 at 2 14 55 PM

The red lines are actually parallel to the minimal curvature direction.

I wrote unit test to simply make sure that max curvature is no less than min curvature.

I also fix the corresponding tutorial.

#### Checklist

- [x] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/principal_curvature.cpp
+++ b/include/igl/principal_curvature.cpp
@@ -40,8 +40,8 @@ class CurvatureCalculator
 {
 public:
   /* Row number i represents the i-th vertex, whose columns are:
-   curv[i][0] : K2
-   curv[i][1] : K1
+   curv[i][0] : K1
+   curv[i][1] : K2
    curvDir[i][0] : PD1
    curvDir[i][1] : PD2
    */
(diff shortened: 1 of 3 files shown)


jdumas:
@danielepanozzo can you confirm this was a bug?

danielepanozzo (approved):
Yes, this was indeed a bug, thank you for the PR.

PR #1778, opened 2021-04-04
added uniform sampler estimation multiplier parameter

gorilux:
Fixes # .

I've added a new parameter to change the uniform distribuition multiplier that was hardcoded to 30.0. I've maintained the existing function signature so it doesn't break the api and functionallity. I calls the new function with the multiplier parameter set to 30.
Unit tests passed.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/copyleft/cgal/SelfIntersectMesh.h
+++ b/include/igl/copyleft/cgal/SelfIntersectMesh.h
@@ -635,32 +635,14 @@ inline bool igl::copyleft::cgal::SelfIntersectMesh<
   const std::vector<std::pair<Index,Index> > shared)
 {
   using namespace std;
-  
-  auto opposite_vertex = [](const Index a0, const Index a1) {
-    // get opposite index of A
-    int a2=-1;
-	for(int c=0;c<3;++c)
-      if(c!=a0 && c!=a1) {
-        a2 = c;
-        break;
-      }
-      assert(a2 != -1);
-      return a2;
-  };
 
   // must be co-planar
-  // must be co-planar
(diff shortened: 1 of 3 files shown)


include/igl/copyleft/cgal/SelfIntersectMesh.h
```diff
@@ -635,14 +635,32 @@ inline bool igl::copyleft::cgal::SelfIntersectMesh<
   const std::vector<std::pair<Index,Index> > shared)
 {
   using namespace std;
+  
+  auto opposite_vertex = [](const Index a0, const Index a1) {
```
jdumas, line 639: Please do not mix up contributions in this PR. Could you keep the contrib to separate branches so we can review them separately? Thanks!

gorilux: It wasn't supposed to be included. I pushed this change after I created the pull request.

gorilux: I've reverted the commit.

alecjacobson: looks cool though. maybe we should test and merge this as a separate PR?

gorilux: Ok, I'll cancel this one and create a new one.

jdumas: @alecjacobson this was taken from #1413. I've asked @qnzhou to review and approve the PR, but it's been sitting in limbo since >1y ...

PR #1790, opened 2021-04-23
Remove serialize / deserialize overloads

tomlankhorst:
These overloads effectively replaced compile-time errors with run-time errors. 

If a user calls `serialize` on an object that does not implement this, the compilation should fail and the user should either implement the serialization or remove the serialization call.

#### Checklist
- [ ] Adds new .cpp file.
- [X] This is a minor change.

(This PR makes #1789 irrelevant)

### DIFF
--- a/include/igl/serialize.h
+++ b/include/igl/serialize.h
@@ -1241,19 +1241,6 @@ namespace igl
  
     }
  
-    // functions to overload for non-intrusive serialization
-    template <typename T>
-    inline void serialize(const T& obj,std::vector<char>& buffer)
-    {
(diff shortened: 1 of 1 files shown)


alecjacobson:
agreed. I wonder if there was a fear that callers wouldn't know what to do or that the compiler error is too cryptic? In that case, I guess another option would be to put static_asserts here.

alecjacobson:
I don't use the serialize functions so my heads not really in this implementation. Is this PR close to being ready to merge? If not, what's the remaining work if any?

alecjacobson:

PR #1800, opened 2021-05-02
Instead of calling `exit(1)`, teach triangle to throw exception

alecjacobson:
Switch triangle version to [one where it's possible to customize the function `triexit`](https://github.com/libigl/triangle/commit/9cceb2f0c9d364ec229fd9182b4f45e3c5bed63a) which is called when triangle fails.

Meanwhile, in libigl provide this custom implementation to throw a `std::runtime_exception`.

(Note, this change is unrelated to switching to https://github.com/qnzhou/trianglelite . That may be a good idea in the future but doesn't immediately solve the problem of triangle calling exit)

### DIFF
--- a/cmake/LibiglDownloadExternal.cmake
+++ b/cmake/LibiglDownloadExternal.cmake
@@ -157,7 +157,7 @@ endfunction()
 function(igl_download_triangle)
 	igl_download_project(triangle
 		GIT_REPOSITORY https://github.com/libigl/triangle.git
-		GIT_TAG        5a70326574b34d6a51d9eaf6a9f78813657ee108
+		GIT_TAG        4df461c0083e0d768fe42ab41a617070b5acc5ef
 	)
 endfunction()
 
--- a/include/igl/triangle/triangulate.cpp
+++ b/include/igl/triangle/triangulate.cpp
@@ -22,10 +22,7 @@
 #define REAL double
 #define VOID int
 
-extern "C"
-{
 #include <triangle.h>
-}
 
 #undef ANSI_DECLARATORS
 #ifdef IGL_PREVIOUSLY_DEFINED_ANSI_DECLARATORS
@@ -85,9 +82,9 @@ IGL_INLINE void igl::triangle::triangulate(
   using namespace std;
   using namespace Eigen;
 
-  assert( (VM.size() == 0 || V.rows() == VM.size()) && 
+  assert( (VM.size() == 0 || V.rows() == VM.size()) &&
     "Vertex markers must be empty or same size as V");
-  assert( (EM.size() == 0 || E.rows() == EM.size()) && 
+  assert( (EM.size() == 0 || E.rows() == EM.size()) &&
     "Segment markers must be empty or same size as E");
   assert(V.cols() == 2);
   assert(E.size() == 0 || E.cols() == 2);


jdumas:
Hmm this is a really murky modification. At this point we might as well modify our version of triangle to compile with .cpp and throw an exception directly. Providing an incomplete "triangle" lib where `triexit` is undefined can cause other problems down the line. E.g. what if another project have two components A and B, but A depends on libigl, and B only depends on triangle? Suddenly B is linked against `triangle` but doesn't provide implementation for `triexit` (which is in libigl). To make things more complicated libigl is currently a "header-only" library, but the implementation currently resides in a .cpp in the `include/` folder, which is problematic.

alecjacobson:
> Hmm this is a really murky modification. At this point we might as well modify our version of triangle to compile with .cpp and throw an exception directly. Providing an incomplete "triangle" lib where `triexit` is undefined can cause other problems down the line. E.g. what if another project have two components A and B, but A depends on libigl, and B only depends on triangle? Suddenly B is linked against `triangle` but doesn't provide implementation for `triexit` (which is in libigl). 
Happy to accept a PR that does that change instead. This was meant to be a minimal change to libigl/triangle. I don't know what converting the whole thing to .cpp would entail (maybe that's what qnzhou/trianglelite; but it still just downloads the original triangle with this exit problem).

I'm not sure I 100% understand the A/B scenario. In this scenario, some component of the project is using triangle via libigl but not linking to libigl? Is this a scenario worth worrying about?

> To make things more complicated libigl is currently a "header-only" library, but the implementation currently resides in a .cpp in the `include/` folder, which is problematic.
This aspect should be fixed now.

jdumas:
> Happy to accept a PR that does that change instead. This was meant to be a minimal change to libigl/triangle. I don't know what converting the whole thing to .cpp would entail (maybe that's what qnzhou/trianglelite; but it still just downloads the original triangle with this exit problem).

Actually you don't need to convert the whole file to .cpp. Simply add a file `triexit.cpp` in the [libigl/triangle](https://github.com/libigl/triangle) repo instead of adding your custom triexit function into libigl. Bonus point: you get to keep the `TRIANGLE_CUSTOM_TRIEXIT` define private to the `triangle` library, and it doesn't leak into client headers.

> I'm not sure I 100% understand the A/B scenario. In this scenario, some component of the project is using triangle via libigl but not linking to libigl? Is this a scenario worth worrying about?

Yes, this happens all the time. If we put aside the `triangle` licensing issues for the sake of the argument, in a large product codebase you can have plenty of components using triangle directly, with other parts of the code using triangle through libigl (e.g. anyone who wants to use polyfem). Putting `triexit()` directly in `libigl` instead of `triangle` make the `triangle` not self-contained anymore, and make integration more problematic.

It is a small change to add the file `triexit.cpp` directly to our `libigl/triangle` repo instead, so I would prefer to do that for now.

include/igl/triangle/triexit.cpp
```diff
@@ -0,0 +1,9 @@
+#include <string>
+#include <stdexcept>
+extern "C"
+{
+  void triexit(int status)
+  {
+    throw std::runtime_error("triexit("+std::to_string(status)+")");
```
jdumas, line 7: Btw I am not sure what happens when you put `extern "C"` around a code block and then throw a C++ exception inside...

jdumas: This seems to be undefined behavior (albeit it probably works in practice):
https://stackoverflow.com/questions/15845681/can-c-functions-marked-as-extern-c-throw

jdumas:
Oh and yes `qnzhou/trianglelite` wouldn't solve this issue either, as it also depends on vanilla `triangle`. `qnzhou/trianglelite` is more like an alternative to the whole `igl/triangle` wrapper.

jdumas:
Btw I just tried renaming `triangle.c` into `triangle.cpp` it compiles perfectly fine out of the box, so to avoid the undefined behavior of throwing exceptions across C functions, I would rather switch triangle to .cpp directly.

alecjacobson:
Let's do it! That seems better all around.

jdumas (approved):

jdumas:
Alright PR updated -- changes are much smaller now. Feel free to merge.

PR #1802, opened 2021-05-04
Introduce continuous fuzzing

AdamKorcz:
Hello, this is [Adam](https://twitter.com/AdamKorcz4) of [Ada Logics](https://adalogics.com/). I have been working on setting up continuous fuzzing of libligl through OSS-fuzz. 
I have now come up with the first fuzz test. An integration PR with OSS-fuzz has been set up [here](https://github.com/google/oss-fuzz/pull/5708), and for the fuzzer to run continuously, we need to steer it away from runtime errors in the target code. For that a macro has been introduced.

For those unfamiliar: Fuzzing is a way of testing software, whereby pseudo-random data is passed to a target application with the goal of finding bugs and vulnerabilities. 

### Changes made:

1. Add fuzz test
2. Add macro to prevent target code from terminating during fuzz run

#### Checklist

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,10 @@ if(LIBIGL_BUILD_TESTS)
 	add_subdirectory(tests)
 endif()
 
+# OSS-fuzz build
+# This is only used by OSS-fuzz
+option(OSS_FUZZ "Build for OSS-fuzz" OFF)
+
 if(LIBIGL_TOPLEVEL_PROJECT)
 	# Set folders for Visual Studio/Xcode
 	igl_set_folders()
--- a/include/igl/MshLoader.cpp
+++ b/include/igl/MshLoader.cpp
@@ -16,6 +16,12 @@
 
 #include <string.h>
 
+#ifdef OSS_FUZZ
+#define NO_ABORT_ON_THROW 1;
+#else
+#define NO_ABORT_ON_THROW 0;
+#endif
+
 namespace igl {
     // helper function
     void inline _msh_eat_white_space(std::ifstream& fin) {
@@ -28,11 +34,15 @@ namespace igl {
 }
 
 IGL_INLINE igl::MshLoader::MshLoader(const std::string &filename) {
+    // Used to prevent fuzzer from terminating on std::runtime_error;
+    int no_abort = NO_ABORT_ON_THROW;
+    
     std::ifstream fin(filename, std::ios::in | std::ios::binary);
 
     if (!fin.is_open()) {
         std::stringstream err_msg;
         err_msg << "failed to open file \"" << filename << "\"";
+        if(no_abort){ return; };
         throw std::ios_base::failure(err_msg.str());
     }
     // Parse header
@@ -40,7 +50,10 @@ IGL_INLINE igl::MshLoader::MshLoader(const std::string &filename) {
     double version;
     int type;
     fin >> buf;
-    if (buf != "$MeshFormat") { throw std::runtime_error("Unexpected .msh format"); }
+    if (buf != "$MeshFormat") {
+        if(no_abort){ return; };
+        throw std::runtime_error("Unexpected .msh format");
+    }
 
     fin >> version >> type >> m_data_size;
     m_binary = (type == 1);
@@ -49,6 +62,7 @@ IGL_INLINE igl::MshLoader::MshLoader(const std::string &filename) {
         // probably unsupported version
         std::stringstream err_msg;
         err_msg << "Error: Unsupported file version:" << version << std::endl;
+        if(no_abort){ return; };
         throw std::runtime_error(err_msg.str());
 
     }
@@ -56,11 +70,13 @@ IGL_INLINE igl::MshLoader::MshLoader(const std::string &filename) {
     if (m_data_size != 8) {
         std::stringstream err_msg;
         err_msg << "Error: data size must be 8 bytes." << std::endl;
+        if(no_abort){ return; };
         throw std::runtime_error(err_msg.str());
     }
     if (sizeof(int) != 4) {
         std::stringstream err_msg;
         err_msg << "Error: code must be compiled with int size 4 bytes." << std::endl;
+        if(no_abort){ return; };
         throw std::runtime_error(err_msg.str());
     }
 
@@ -74,6 +90,7 @@ IGL_INLINE igl::MshLoader::MshLoader(const std::string &filename) {
                 err_msg << "Binary msh file " << filename
                 << " is saved with different endianness than this machine."
                 << std::endl;
+            if(no_abort){ return; };
             throw std::runtime_error(err_msg.str());
         }
     }
@@ -83,6 +100,7 @@ IGL_INLINE igl::MshLoader::MshLoader(const std::string &filename) {
     { 
         std::stringstream err_msg;
         err_msg << "Unexpected contents in the file header." << std::endl;
+        if(no_abort){ return; };
         throw std::runtime_error(err_msg.str());
     }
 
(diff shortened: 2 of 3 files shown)


include/igl/MshLoader.cpp
```diff
@@ -16,6 +16,12 @@
 
 #include <string.h>
 
+#ifdef OSS_FUZZ
+#define NO_ABORT_ON_THROW 1;
```
jdumas, line 20: I do not get why you need to define a special behavior for the fuzz test. Most libigl code is written in C++, and throws exceptions upon error. We cannot go around and start replacing all our throw with a simple return.

CMakeLists.txt
```diff
@@ -68,6 +68,10 @@ if(LIBIGL_BUILD_TESTS)
 	add_subdirectory(tests)
 endif()
 
+# OSS-fuzz build
+# This is only used by OSS-fuzz
+option(OSS_FUZZ "Build for OSS-fuzz" OFF)
```
jdumas, line 73: Any new CMake/macro introduced should be prefixed by `LIBIGL_`. As it stands, I am not sure this one is needed at all (see below).

tests/fuzzing/igl_fuzzer.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+
+#include <stdint.h>
+#include <string.h>
+#include <stdlib.h>
+#include <igl/MshLoader.h>
+#include <iostream>
+#include <fstream>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
```
jdumas, line 17: Why is this linked as `extern "C"`?

jdumas (changes requested):
Thanks. I appreciate the contribution, but I have a couple of questions because I'd be willing to merge this.

First of all, why does this need to be integrated into the upstream repo OSS-Fuzz? Can't we just add fuzz testing directly in libigl through Github Actions? Having a dangling code `igl_fuzzer.cpp` that is not compiled in any way by our CI is problematic for maintenance. I would rather have a proper unit test integrated with Catch2 for example.

I am not very familiar with fuzzing and how OSS-Fuzz operates, so please be patient with me =)

AdamKorcz:
> First of all, why does this need to be integrated into the upstream repo OSS-Fuzz?

Fuzzing has by far the greatest benefit when done continuously. I have written a small blog post about why [here](https://adalogics.com/blog/the-importance-of-continuity-in-fuzzing-cve-2020-28362).  In short: There are many practical examples of bugs being found only after running the fuzzers for days, months and even several cpu years. Fuzzing for 10-30 minutes is not enough (I assume anything more than that would be a nuisance during CI).

> Can't we just add fuzz testing directly in libigl through Github Actions?

I would definitely recommend setting up [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) which is offered for all projects integrated with OSS-fuzz. I would be happy to set that up once Libigl is integrated with OSS-fuzz.

include/igl/MshLoader.cpp
AdamKorcz: This is a suggested solution to the following problem: runtime errors, `abort`, `exit` are fuzz blockers. 
Another solution could be to use a custom libigl_exit(exit_type) and then consider if we are fuzzing inside it. This is also merely a suggestion.
Nonetheless, I do have some bandwidth for this if it could help improve the fuzzability of Libigl.

tests/fuzzing/igl_fuzzer.cpp
AdamKorcz: It is a requirement to expose it to Libfuzzer when we are dealing with C++

include/igl/MshLoader.cpp
jdumas: Can't you simply wrap the libigl function call with a try/catch in the `igl_fuzzer.cpp` test file? If exceptions are a blocker for fuzzing, you can't even call STL functions anymore, since they can raise standard exceptions. I would like to find a solution that is less intrusive to the libigl codebase.

jdumas:
> Fuzzing has by far the greatest benefit when done continuously. I have written a small blog post about why here. In short: There are many practical examples of bugs being found only after running the fuzzers for days, months and even several cpu years. Fuzzing for 10-30 minutes is not enough (I assume anything more than that would be a nuisance during CI).

Right. I do have some environmental concerns about running CPU resources continuously like that though (even if the purpose is to find bugs rather than mine bitcoins).

Regarding this PR though, the concern I have is maintenance. We need to ensure that `igl_fuzzer.cpp` at least compiles and link correctly, otherwise any change we make can break the fuzz test code. If we do not compile `igl_fuzzer.cpp` in our CI, it sounds like it should go into the OSS-Fuzz project instead.

> I would definitely recommend setting up CIFuzz which is offered for all projects integrated with OSS-fuzz. I would be happy to set that up once Libigl is integrated with OSS-fuzz.

Hmm yes, that would be nice actually. I am currently working on a revamp of our CMake system, so I would suggest wait until this is merged to work on CIFuzz integration. I also need to read a bit more about OSS-Fuzz and how it works in the meantime.

AdamKorcz:
> Right. I do have some environmental concerns about running CPU resources continuously like that though (even if the purpose is to find bugs rather than mine bitcoins).

OSS-fuzz is a project that runs fuzzers of critical open source projects continuously for free. The igl_fuzzer.cpp would be run by OSS-fuzz continuously, and if bugs were found, a report would be sent to maintainers.

> We need to ensure that igl_fuzzer.cpp at least compiles and link correctly

A way could be to simply integrate it into the build system. From experience it seems to make checking easy enough to notice any breakages. It is something I would be happy to do, however I would strongly suggest doing that in the next step to not solve too many problems in this PR.

jdumas:
> It is something I would be happy to do, however I would strongly suggest doing that in the next step to not solve too many problems in this PR.

I don't see why this needs to be done in two separate PRs. It doesn't really make sense to merge something without ensuring it compiles on all 3 platforms.

AdamKorcz:
> I don't see why this needs to be done in two separate PRs. It doesn't really make sense to merge something without ensuring it compiles on all 3 platforms.

From my perspective: Fuzzing Libigl needs be done continuously. Otherwise we are doing work on writing the fuzzers that will not be executed properly and the benefit of the fuzzers will be low. Therefore, I would like a confirmation that Libigl would like to integrate into OSS-fuzz. 
For that I suggest an incremental approach to first integrate with OSS-fuzz with a working prototype, and then proceed with an improved build system.
We can absolutely start with having the fuzzer and the modified project files hosted on the OSS-fuzz side.

jdumas:
Thanks. I am definitely interested in fuzz testing integration, with the following caveats:
1. We cannot modify our code to prevent exceptions from being thrown. We should use try/catch around the testing code to prevent leaking if that is an issue.
2. If a file is not compiled by our CMake system, it should not live in the libigl repository, since we cannot maintain compatibility. I would prefer the fuzzer to be hosted on the OSS-fuzz side until we can compile it with our CMake system.

I also want to be clear that fuzz testing is not a high priority item for us as the moment, so while I appreciate the contribution, we need to be careful of the maintenance cost/benefits incurred by theses changes.

AdamKorcz:
I have made the necessary changes on the OSS-fuzz side which means:

1. For now all files are hosted on OSS-fuzz.
2. No changes are made to Libgil.

If it looks fine, could you respond with an email address for the bug reports on the OSS-fuzz PR?

jdumas:
Closing as work has been integrated in oss-fuzz without any change to libigl itself.

Issue #1783, opened 2021-04-09
multiplication out of "unsigned int" range in function "point_in_poly"


xarthurx:
#### Describe the bug
OK, I have to say this bug costs me a lot of time to dig and find...

When deciding if a point is inside a given closed polygon, the function uses an approach as stated in:
http://www.visibone.com/inpoly/

However, there is a step that makes the internal value going out of the bound if its type, and thus make the function not working as expected.

This line and the line below:
https://github.com/libigl/libigl/blob/e2877039121023dd870d1172d4b209aae160f9a4/include/igl/point_in_poly.cpp#L43

Here, the function converts `unsigned int` to `long` first and conduct the multiplication. However, on my machine:
```
max limit of unsigned int: 4294967295
max limit of long:         2147483647
```

This makes the test go out of range if the given point is large (it doesn't have to be too large even, `400000` is already enough to make it buggy).

The original function is designed for images (the number is usually not too big), but when applying to general geometry, it is quite limited.

I'm not sure about the range of different types in C as the original code, but here, it seems the bug will affect a lot.
Perhaps the "sum of angle" approach should be used to re-write the function?

#### Platform
- [x] Windows
- [ ] macOS
- [ ] Linux

alecjacobson:
Sum of angle is what winding number does. I always use that and had
forgotten we have this in poly function.

xarthurx:
I've checked the WindingNumberTree` Class, may I ask how can I use it for 2D polyline boundary?
Does it seem that was designed for closed meshes?

alecjacobson:
call igl::winding_number directly. The tree you're referring to is just for
3D. In 2D, the computation is not accelerated (i.e., it's O( #F #P) )

>

xarthurx:
@alecjacobson 
Does the `igl::winding_number` only accept `V` in `double` format?\

Considering the example below:

```
  MatrixXi testV(4, 2);
  testV << 0, 0, 2, 0, 2, 2, 0, 2;
  MatrixXi testF(4, 2);
  testF << 0, 1, 1, 2, 2, 3, 3, 0;
  Vector2i pt(1, 1);

  auto wn = igl::winding_number(testV, testF, pt);
  cout << "winding number: " << wn << endl;
```
gives `0`, but 

```
  MatrixXd testV(4, 2);
  testV << 0, 0, 2, 0, 2, 2, 0, 2;
  MatrixXi testF(4, 2);
[...]

PR #1810, opened 2021-05-21
delete obsolete func based on: https://github.com/libigl/libigl/issue…

xarthurx:
…s/1783#issuecomment-845248820

Fixes https://github.com/libigl/libigl/issues/1783

https://github.com/libigl/libigl/issues/1783#issuecomment-845248820

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
Wow. I didn't even realize this old function was defined for integer coordinates. I'm guess it wasn't getting used much. For posterity, if anyone was using this function before. Perhaps like this:

```cpp
std::vector<std::vector<unsigned int > >poly;
unsigned int xt,yt;
bool res = igl::point_in_poly(poly,xt,yt);
```

You could switch to calling `igl::winding_number`: 
```
Eigen::MatrixXd V;
igl::list_to_matrix(poly,V);
Eigen::MatrixXi E(V.rows(),2);
for(int v = 0;v<V.rows();v++){ E.row(v) << v,(v+1)%V.rows(); }
bool res = std::abs(igl::winding_number(V,E,Eigen::RowVector2d(xt,yt))>0.5;
```

For most users, you'll already have your polygon in the `V` format. `igl::winding_number` can also process many query points (in parallel). In the future, it'd be nice to implement the fast variants of wn for point-in-poly queries.

In my experience, `igl::winding_number` will hold up pretty well w.r.t. floating point robustness, with the most obvious pitfall of the singularity for points directly on the polygon (or "too close" for good floating point math). For a robust query, probably CGAL is needed.

PR #1811, opened 2021-05-23
CGAL cdt with triangle's api

alecjacobson:
Future work is to implement more of the triangle api and to overload a version that doesn't use the triangle string flags (and keep this version to call it for easy switching).

#### Checklist

- [x] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/copyleft/cgal/triangulate.cpp
```diff
@@ -0,0 +1,143 @@
...
+    cdt.insert_constraint(points[E(e,0)],points[E(e,1)]);
+  }
+
+  // https://doc.cgal.org/latest/Triangulation_2/index.html#title30
+  //
+  // "remove" connected component of face start by marking faces as
+  // in_domain = false. start should not yet be visited.
+  const auto remove_cc = [&](Face_handle start)
+  {
+    std::list<Face_handle> queue;
```
jdumas, line 68: I wouldn't use a std::list for that. std::queue under the hood uses a std::dequeue, which should offer better performances.

jdumas (commented):
I'm not sure it's a good idea to introduce a new function using the `std::string` flags, if we know we want to move away from that in the future. [trianglelite](https://github.com/qnzhou/trianglelite/blob/main/include/trianglelite/Config.h) uses a simple config struct, which would be less error prone and easier to use. We could have a utility function `Config::from_string()` that converts a triangle cli switch to a struct filled with the appropriate content to facilitate code update...

alecjacobson:
Agreed. That's what I'm referring to as future work. For now, it's good to have the same API.

jdumas:
"future work" just sound like an easy way for the task to be added to the backlog and never be completed. Why not do it as part of this PR?

alecjacobson:
If you're offering, please push it onto this PR.

>

jdumas:
Sure, I can add it to my own backlog. But let's hold on merging this in the meantime. The last thing I want is to introduce more [broken windows](https://blog.codinghorror.com/the-broken-window-theory/) for the sake of expediency.

include/igl/copyleft/cgal/triangulate.cpp
```diff
@@ -0,0 +1,148 @@
...
+  for(Eigen::Index i = 0;i<V.rows();i++)
+  {
+    Point_2 p(V(i,0),V(i,1));
+    cdt.insert(p);
+    points.emplace_back(p);
+  }
+  // Insert all edges 
+  for(Eigen::Index e = 0;e<E.rows();e++)
+  {
+    cdt.insert_constraint(points[E(e,0)],points[E(e,1)]);
```
sloriot, line 61: You may want to use [this function](https://doc.cgal.org/latest/Triangulation_2/classCGAL_1_1Constrained__triangulation__plus__2.html#a247b1f12c580958700ba7618c59b96c2) to get better performances. As far as I see, you can use [`boost::transform_iterator`](https://www.boost.org/doc/libs/1_78_0/libs/iterator/doc/html/iterator/specialized/transform.html) together with a lambda to create pairs on the fly.

PR #1813, opened 2021-05-24
Bumps the triangle version

alecjacobson:
The latest libigl triangle now has a conservative bail out for an [infinite loop bug](https://github.com/qnzhou/trianglelite) in the original triangle.


jdumas (approved):

jdumas:
Honestly feels like we are doing Shewchuk's maintenance for him... Which would be fine if the license on triangle wasn't such a grey area.

PR #1815, opened 2021-05-26
bump tetgen commit hash

alecjacobson:
tetgen commit hash was pointing to the wrong repo. Meanwhile, tetgen has released a new version, now mirrored in https://github.com/libigl/tetgen

### DIFF
--- a/cmake/LibiglDownloadExternal.cmake
+++ b/cmake/LibiglDownloadExternal.cmake
@@ -141,7 +141,7 @@ endfunction()
 function(igl_download_tetgen)
 	igl_download_project(tetgen
 		GIT_REPOSITORY https://github.com/libigl/tetgen.git
-		GIT_TAG        6407f3962d4e24426a81955664391da74e2e2311
+		GIT_TAG        4f3bfba3997f20aa1f96cfaff604313a8c2c85b6
 	)
 endfunction()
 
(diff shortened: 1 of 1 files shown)


cmake/LibiglDownloadExternal.cmake
```diff
@@ -140,8 +140,8 @@ endfunction()
 ## TetGen
 function(igl_download_tetgen)
 	igl_download_project(tetgen
-		GIT_REPOSITORY https://github.com/jdumas/tetgen.git
-		GIT_TAG        c63e7a6434652b8a2065c835bd9d6d298db1a0bc
+		GIT_REPOSITORY https://github.com/libigl/tetgen.git
+		GIT_TAG        6407f3962d4e24426a81955664391da74e2e2311
```
jdumas, line 144: Where did you get this SHA1 from? I don't see it in our tetgen mirror.

jdumas:
I already fixed this as part of #1805. I would rather avoid creating additional conflicts with that PR, and focus on reviewing/merging #1805 instead...

cmake/LibiglDownloadExternal.cmake
alecjacobson: it's in the v16 branch. The commit hash should be the same when that's merged.

alecjacobson:
> I already fixed this as part of #1805. I would rather avoid creating file conflicts with that PR, and focus on reviewing/merging #1805 instead...

I think you fixed the wrong repo thing but it's still pointing to the out-of-date version of tetgen.

jdumas:
Oh I didn't see TetGen had a new version 1.6 released in 2020... What a time to live it. Let's merge your 1.6 branch onto the main tetgen branch first then. Then we should update the SHA1 in #1805 directly.

PR #1819, opened 2021-05-30
Rework sparse repmat

paul0noah:
Fixes # .

Repmat not working for colum major sparse matrices.
Improves performance of repmat

#### Checklist

- [x] Adds new .cpp file.
- [ ] This is a minor change.


alecjacobson:
Thanks!

Does https://eigen.tuxfamily.org/dox/classEigen_1_1Replicate.html in Eigen work for sparse matrices? If so, we could deprecate this.

If not, we should keep `repmat` and merge this helpful fix. Unless I'm misremembering some functions, I think every where we use `Eigen::SparseMatrix` in libigl we just expose the numeric type (e.g., `Eigen::SparseMatrix<T>` where `T` could be `double` or `float`). This means we assume/force the storage ordering to be the default (which I believe is column-major). If we merge this, it will be a bit of an outlier in that this function will template on the order, but all(?) other libigl functions will expect default ordering.

paul0noah:
Sorry for the late reply. 

As far as i know, `replicate()` is not supported for sparse matrices (at least it does not appear in the sparse matrix manipulations tutorial https://eigen.tuxfamily.org/dox/group__TutorialSparse.html and my experiments with it did also fail)

I understand that it might be nice to not expose the major type to the outside but it does limit the functionality quite a lot (you can only use `RowMajor` or `ColMajor` format). I pledge to expose the major type to the outside to increase the functionality.

Issue #1707, opened 2021-01-26
c++17 imcompatible code in the blue_noise


xarthurx:
#### Describe your question

https://stackoverflow.com/questions/45013977/random-shuffle-is-not-a-member-of-std-error

I encounter the exact error when compiling the `blue_noise.cpp` file.

based on **https://en.cppreference.com/w/cpp/algorithm/random_shuffle**
The `std::random_shuffle` is removed in `c++17` and is replaced by `std::shuffle`.
Perhaps we should change it?

P.S. the first link above provides a setting in `Visual Studio` to suppress the issue, but I guess it is anyway good to point it out here.

#### Check all that apply (change to `[x]`)
- [x] Windows
- [ ] macOS
- [ ] Linux

xarthurx:
A `PR` could be submitted if needed.

z-hasher:
Just encountered this today when tuning clang, seems like I'm not alone.😁

PR #1824, opened 2021-06-15
make libigl cpp17 complient

AndreyMlashkin:
Fixes # .

make libigl cpp17 complient

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/blue_noise.cpp
+++ b/include/igl/blue_noise.cpp
@@ -14,6 +14,7 @@
 #include "get_seconds.h"
 #include <unordered_map>
 #include <algorithm>
+#include <random>
 #include <vector>
 
 namespace igl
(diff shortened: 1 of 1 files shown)


AndreyMlashkin:
fixes https://github.com/libigl/libigl/issues/1707

include/igl/blue_noise.cpp
```diff
@@ -209,7 +210,9 @@ namespace igl
     }
         //printf("  --------\n");
     // randomize order: this might be a little paranoid...
-    std::random_shuffle(std::begin(N), std::end(N));
+    std::random_device rd;
```
jdumas, line 213: Don't initialize with random device, this makes the function nondeterministic. Either use the default seed, or pass the seed as a && argument to the function.

jdumas (changes requested):

include/igl/blue_noise.cpp
AndreyMlashkin: Isn't it the point of shuffling to make it nondeterministic?

jdumas: No. This shuffle seems to be here to avoid biasing the sampling due to cell ordering, and it might not even be strictly necessary at that. A random function shouldn't be nondeterministic, those are different concepts.

jdumas:
Fixed by #1949

PR #1851, opened 2021-07-09
Add a helper function to get the current size of the viewer.

xarthurx:
Based on the discussion here:
https://github.com/libigl/libigl/discussions/1850

Add a helper function to provide the current size of the viewer.
This is meant the help relocate the menu to different borders of the viewer after manual resizing.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/readMESH.cpp
+++ b/include/igl/readMESH.cpp
@@ -48,7 +48,7 @@ IGL_INLINE bool igl::readMESH(
     while(still_comments)
     {
       has_line = fgets(line,LINE_MAX,mesh_file) != NULL;
-      still_comments = (line[0] == '#' || line[0] == '\n');
+      still_comments = (line[0] == '#' || line[0] == '\n' || line[0] == '\r');
     }
     return has_line;
   };


include/igl/opengl/glfw/Viewer.cpp
```diff
@@ -938,6 +938,13 @@ namespace glfw
     }
   }
 
+  IGL_INLINE void Viewer::get_size(int &w, int &h)
+  {
+    if (window) {
+    glfwGetWindowSize(window, &w, &h);
```
jdumas, line 944: Please indent correctly.

include/igl/opengl/glfw/Viewer.h
```diff
@@ -72,7 +72,8 @@ namespace glfw
     IGL_INLINE bool save_scene(std::string fname);
     // Draw everything
     IGL_INLINE void draw();
-    // OpenGL context resize
+    // OpenGL context get size / resize
+    IGL_INLINE void get_size(int &w, int &h); // helper func for getting the size of current window
```
jdumas, line 76: The function should return a `std::pair<int, int>` rather than use output args.

jdumas (changes requested):

include/igl/opengl/glfw/Viewer.h
xarthurx: is `libigl` c++17 compatible?

xarthurx: so that `return {w, h}` is acceptable for `std::pair<int, int>`?

xarthurx: code updated. Please let me know if it needs additional edit.

jdumas: Returning `{w, h}` is just C++11, building a std::pair from an initializer list. If you were to retrieve those values using a structured binding `auto [w, h]` then it would be C++17.

xarthurx: Yes, I use that in my code. OK, then all good. Thank you for the explanation.

xarthurx:
BTW, the Action for Windows virtual machine seems down?

PR #1854, opened 2021-07-17
Naming cleanup for fast_winding_number

wtholliday:
This renames a couple functions so it's more apparent they are pre-computations.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
I think it's worth determining a clearer style policy on this. Currently 90+% of functions have the (exact) same filename. This has a lot of advantages. One big exception to this is "precompute/solve" type functions which are mixed. Some are overloaded so `func(..., out & precompute_data)` and then `func(const in & precompute_data, out & solve_output` and others (notably `min_quad_with_fixed*`) use suffixes  (so, `func_precompute(..., out & precompute_data)` and `func_solve(const in & precompute_data, out & solve_output`).

I'm inclined to argue for being strict about function name matching file name. That means that there are two consistent options: we overload the function name to do both precomputation and solve (which sometimes really could be ambiguous if types are the same) _or_ move the `func_precompute` to the file `func_precompute.h` and `func_solve` to `func_solve.h`.

So in this case, we'd move `fast_winding_number_cloud_precompute` to `fast_winding_number_cloud_precompute.h` etc.

PR #1853, opened 2021-07-13
lscm: fix sign for the area term

kenshi84:
I'm pretty sure this fix is correct -- the area term should be subtracted, not added.
To confirm the correctness, I wrote the following code

```c++
#include "grad.h"
#include "doublearea.h"
#include "per_face_normals.h"

double compute_lscm_energy(
  const Eigen::MatrixXd& V,
  const Eigen::MatrixXi& F,
  const Eigen::VectorXd& W_flat)
{
  const int nV = V.rows();
  const int nF = F.rows();

  // Compute gradient
  Eigen::SparseMatrix<double> G;
  igl::grad(V, F, G);

  Eigen::VectorXd nablaU_flat = G * W_flat.head(nV);
  Eigen::VectorXd nablaV_flat = G * W_flat.tail(nV);

  Eigen::MatrixXd nablaU(nF, 3);
  Eigen::MatrixXd nablaV(nF, 3);
  nablaU << nablaU_flat.segment(0, nF), nablaU_flat.segment(nF, nF), nablaU_flat.segment(2 * nF, nF);
  nablaV << nablaV_flat.segment(0, nF), nablaV_flat.segment(nF, nF), nablaV_flat.segment(2 * nF, nF);

  // Compute face normal
  Eigen::MatrixXd N;
  igl::per_face_normals(V, F, N);

  // Compute area
  Eigen::VectorXd dblA;
  igl::doublearea(V, F, dblA);

  // Sum up
  double sum = 0;

  for (int i = 0; i < nF; ++i) {
    Eigen::Vector3d N_i = N.row(i);
    Eigen::Vector3d nablaU_i = nablaU.row(i);
    Eigen::Vector3d nablaV_i = nablaV.row(i);

    // Rotate nablaU_i about N_i by 90 degrees ccw
    nablaU_i = N_i.cross(nablaU_i);

    sum += 0.5 * dblA[i] * (nablaU_i - nablaV_i).squaredNorm();
  }

  return 0.5 * sum;
}
```

and put the following code

```c++
  std::cout << std::setprecision(20) << "LSCM energy 1: " << compute_lscm_energy(V, F, W_flat) << std::endl;
  std::cout << std::setprecision(20) << "LSCM energy 2: " << (0.5 * W_flat.transpose() * Q * W_flat) << std::endl;
```

after the call to `min_quad_with_fixed_solve`.

Before the fix:
```
LSCM energy 1: 1.5662298333641699877
LSCM energy 2: 0.0041413289004784102632

```
[...]

PR #1863, opened 2021-07-28
lscm: fix order of UV after area term sign fix

factoryofthesun:
The fix in #1853 causes the solution UVs to be flipped, so this is a simple fix to that.

See attached a parameterization of an example surface. Apologies in advance for no reproducible example code in C -- I've been working with the python bindings and whipped this up on short notice. 

#### Checklist

screenshot_000000
igl_lscm_param2

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/lscm.cpp
+++ b/include/igl/lscm.cpp
@@ -63,7 +63,7 @@ IGL_INLINE bool igl::lscm(
   V_uv.resize(V.rows(),2);
   for (unsigned i=0;i<V_uv.cols();++i)
   {
-    V_uv.col(V_uv.cols()-i-1) = W_flat.block(V_uv.rows()*i,0,V_uv.rows(),1);
+    V_uv.col(i) = W_flat.block(V_uv.rows()*i,0,V_uv.rows(),1);
   }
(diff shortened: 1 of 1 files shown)


kenshi84 (approved):

alecjacobson:
lol, so the git blame shows that this line has been toggled before https://github.com/libigl/libigl/commit/e4876bd595df855ba37ba0dc443854d8560cf0d2

What's the real issue here? Isn't LSCM orientation determined by the boundary conditions? If so, why should this method flip the output? If my understanding is correct then I'm inclined to merge this PR (and revert back to the original code!), but it'd be nice to be sure that the intent.

kenshi84:
> Isn't LSCM orientation determined by the boundary conditions? 

I don't think fixing two boundary vertices will determine the orientation of triangles.

My guessed logical explanation is that by using the wrong sign (+ instead of -) of the area term, the wrong LSCM energy wanted to make all the triangles flipped (i.e. to make their areas negative).

alecjacobson (approved):

PR #1867, opened 2021-08-03
Fixes a bug in writePLY 

vfonov:
A bug was introduced in dfda6de04ee733a7b00b1edc57051d1fee78b219
Added a test to make sure it works as expected

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
lgtm. Can't imagine the builds won't succeed, but let's merge once they do. thanks.

PR #1895, opened 2021-09-20
Bump CGAL → Boolean + remesh_*intersections performance boost

alecjacobson:
1. Bumps the CGAL to v5.4
2. adds a bunch of `parallel_for`s to CGAL routines
3. relaxes floating-point rounding in `assign_scalar`
4. removes some (but not all) of the `vector<vector<vector...` types used in edge-adjacency navigation

Before getting too excited, there are a couple good reasons why we should wait to merge this:
 - the CGAL version is not an official release yet. They're pretty regular and this change looks stable. So, we might just wait for when it's in a release.
 - @qnzhou and I would like to rerun the Thingi10k gauntlet on the Boolean code to be sure these changes don't break anything
 - using this branch for a while in some other projects will flesh out missing templates

#### 1.
The downside to this bump is that the auto-generated template prototypes for CGAL::Epeck related things seems to have changed which means this PR has _deleted all the old /cgal template instantiations_. This PR should contain the (non Windows specific) template instantiations needed by the tutorial and tests.

#### 2.
https://github.com/CGAL/cgal/pull/5402 recently fixed the thread safety issues with the CGAL::Epeck::FT type. This immediately enables a lot of opportunities for parallelism in the `mesh_boolean` and `remesh_intersection` routines. Previously, due to reference counting, even read operations were not threadsafe, so we had some extreme `std::mutex` usage and barely any benefit of parallelism.

For example, on my 56-core macpro I'm seeing:
- ~14x speed up for `remesh_self_intersections`
- ~7x speed up for `mesh_boolean`

#### 3.
While profiling and improving the performance of these routines, I've fixed other bottlenecks. Most notably, the `assign` related conversions to and from CGAL's exact type are quite costly. We previously used a rounding from float→rational, which could be best thought of as "round to nearest float" in the absolute sense. After some tests (not exhaustive yet; see above), I'm coming to the conclusion this is overkill and using `CGAL::to_double( .exact() )` is a safe and faster alternative (vaguely think of as "arbitrarily round down/up"). (see also https://github.com/CGAL/cgal/discussions/6000) 

#### 4.
Unrelated to CGAL, the `mesh_boolean` code uses many `vector<vector<vector<...` type lists of lists. We've been previously trying to encourage use of the `uEC` and `uEE` serialized arrays for edge-to-half-edge equivalence. This makes that change in relevant places related to `mesh_boolean`ing. There are still a number of `vector<vector<vector<...`, notably in `triangle_triangle_adjacency` that are an eyesore (thought nowhere near the bottleneck of `mesh_boolean`.

#### other.
Various other clean up, documentation, function splittings, templates, etc. made along the way.


jdumas:
Cool! I suspect there will be some incompatibilities with #1413 (which has been under review since forever), and #1805 (for the cgal::assign file). Would be great to get those sorted first.

alecjacobson:
Great points, Jeremie! I believe this should be compatible with #1413 , when @qnzhou and I do the test, we should start with that one, then re-run this PR.

Re our favorite #1805 , Daniele and I discussed a strategy for moving ahead this morning. We'll be in touch shortly. In terms of compatibility, I hope this PR is just bumping the commit hash, so easy merge but should be tested of course.

alecjacobson:
**Update:**

CGAL v5.4 has been officially released.

I've re-run the PWN subset of Thingi10k "self-union" stress test for the booleans. So far, 8595 out of 8621 have finished (a few take 10+ hours 🤯 ). All looks good in the sense of exactly matching rational output of the previous code. I'm content declaring success here (I'll try to squeeze out the last few models).

I'd also like to test the success rate of the float conversion. This was never guaranteed and the heuristic we used in the paper was never implemented fully in C++ anyway. In any case, I'd like to get a sense of whether we significantly more often get poor snap rounding results with the new code.

alecjacobson:
I'm nearly ready to declare success on this. I've run self-unions all 8620 valid PWN models from thingi10k. The new code produces _exact_ (as in `CGAL::Epeck`) results that are the same (w.r.t. exactly equivalent triangles).

The new code also introduces new, faster rounding. Of these 8620 _exact_ outputs, I first detect any self-intersections in the _exact_ representations. After the #2019 fix, the new and old code agree there are no intersections. Then I convert to double (via `assign`) and check for intersections. Here we expect that some models have intersections. We could be worried that the new code has significantly more intersections after rounding to `double`. This is happily not the case. There are 109 out of 8620 models which produce intersections-after-rounding in either the new or old code. Of those, 29% have no intersections in the old code but ≥1 intersections in the new code.  Conversely, 18% have no intersections in the new code but ≥1 intersections in the old code. In this sense, the old code is _better_, but I think it's roughly random. For 50% of the 109, the new code has more intersections than the old code. Putting in the larger context of the 8620 models, this means for   only 0.006% would the new code output more intersections-after-rounding than the old code. And similarly, the old code would only 0.006% of the time output more intersections-after-rounding than the new code. My conclusion is that it's a wash and we should take the newer, simpler, faster rounding (the old code is still there behind a flag).

qnzhou (approved):
Looks good to me!

PR #1915, opened 2021-11-09
Close files opened by read_triangle_mesh

martinhath:
I think the title and diff speaks for themselves :) 

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


martinhath:
I initially thought this was a problem in `read_triangle_mesh`, but it turns out that most of the `readXXX` functions that take a `FILE *` also closes them. This change makes `readPLY` be consistent with the rest of the mesh reading functions. 

I think it'd be better to remove all of the `fclose` calls in all of the other files and have the rule that if `libigl` `fopen`s a file it will also `fclose` it, and not otherwise.

alecjacobson:
Sure that makes sense to me.

PR #1922, opened 2021-11-20
Modified arap interface to accept Eigen::Map

Poupine:
This Pull Request implements changes discussed here: https://github.com/libigl/libigl/discussions/1920

By changing the interface to accept Eigen::MatrixBase instead of Eigen::PlainObjectBase, it becomes possible to pass data to libigl through Eigen::Map-ped data. It allows the user more flexibility in representing mesh data in their code while not requiring unnecessary copies to Eigen matrices, when a Map can be used.

#### Checklist

- [x ] All changes meet [libigl style-guidelines](https://libigl.github.io/style-guidelines/).
- [ ] Adds new .cpp file.
- [x ] This is a minor change.

### DIFF
--- a/include/igl/arap.cpp
+++ b/include/igl/arap.cpp
@@ -22,20 +22,24 @@
 #include <cassert>
 #include <iostream>
 
+template <typename Scalar>
+using MatrixXX = Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic>;
+
 template <
   typename DerivedV,
   typename DerivedF,
   typename Derivedb>
 IGL_INLINE bool igl::arap_precomputation(
-  const Eigen::PlainObjectBase<DerivedV> & V,
-  const Eigen::PlainObjectBase<DerivedF> & F,
+  const Eigen::MatrixBase<DerivedV> & V,
+  const Eigen::MatrixBase<DerivedF> & F,
   const int dim,
-  const Eigen::PlainObjectBase<Derivedb> & b,
+  const Eigen::MatrixBase<Derivedb> & b,
   ARAPData & data)
 {
   using namespace std;
   using namespace Eigen;
   typedef typename DerivedV::Scalar Scalar;
+  typedef typename DerivedF::Scalar Integer;
   // number of vertices
   const int n = V.rows();
   data.n = n;
@@ -55,8 +59,8 @@ IGL_INLINE bool igl::arap_precomputation(
   assert(data.dim <= V.cols() && "solve dim should be <= embedding");
   bool flat = (V.cols() - data.dim)==1;
 
-  DerivedV plane_V;
-  DerivedF plane_F;
+  MatrixXX<Scalar> plane_V;
+  MatrixXX<Integer> plane_F;
   typedef SparseMatrix<Scalar> SparseMatrixS;
   SparseMatrixS ref_map,ref_map_dim;
   if(flat)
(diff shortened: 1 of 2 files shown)


include/igl/arap.cpp
```diff
@@ -22,20 +22,24 @@
 #include <cassert>
 #include <iostream>
 
+template <typename Scalar>
+using MatrixXX = Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic>;
```
Poupine, line 26: Note that this is only necessary because libigl officially supports Eigen version 3.3.7. I'm not sure from which version, but in 3.4.0, Eigen has a typedef already for MatrixX. Then, this should be removed in favour of the Eigen defined typedef.

include/igl/arap.cpp
```diff
@@ -169,9 +173,9 @@ template <
   typename Derivedbc,
   typename DerivedU>
 IGL_INLINE bool igl::arap_solve(
-  const Eigen::PlainObjectBase<Derivedbc> & bc,
+  const Eigen::MatrixBase<Derivedbc> & bc,
   ARAPData & data,
-  Eigen::PlainObjectBase<DerivedU> & U)
+  Eigen::MatrixBase<DerivedU> & U)
```
Poupine, line 178: @alecjacobson I know in our discussion you said outputs should be defined as Eigen::PlainObjectBase in order to allow resizing. However, in this case, I think MatrixBase makes the most sense since the algorithm is not supposed to resize U, and it is convenient again to have the ability to pass an Eigen::Map here.

include/igl/arap.cpp
alecjacobson: good to know!

include/igl/arap.cpp
alecjacobson: This is simply the convention of libigl. The caller does not need to resize outputs in advance. The cost of this is that we can't use maps for outputs and must copy in situations where the output needs to be written in a caller defined address.

alecjacobson (changes requested):
outputs should be plainobjectbase

include/igl/arap.cpp
Poupine:
```cpp
if(U.size() == 0)
{
// terrible initial guess.. should at least copy input mesh
#ifndef NDEBUG
    cerr<<"arap_solve: Using terrible initial guess for U. Try U = V."<<endl;
#endif
    U = MatrixXd::Zero(data.n,data.dim);
}else
{
    assert(U.cols() == data.dim && "U.cols() match data.dim");
}
```

This snippet of code is why I think PlainObjectBase is doing more harm than good here. It forces the user to provide an object that can be resized and the only case where it needs to be resized is something that is non-sensical algorithmically (or at the very least a terrible initial guess as pointed out by the debug message). So why not instead make the case of U.size() == 0 an error instead? And then there is no requirement for U to be resizable anymore and the user is free to pass in a map to an existing buffer in memory.
I understand if you feel strongly about this in order to make the interface consistent with the rest of libigl but I just thought I'd point this out for reference :)

alecjacobson:
Ah, that's a good point. I'd forgotten that arap_solve is actually already
expecting there to be data in U.

I'm OK with this change (adding assertions on the size of U and clarifying
the documentation in the header).

Thanks for pointing that out.

>

Poupine:
I believe my latest changes address your comments @alecjacobson

Poupine:
Do I need to do anything specific to let GitHub know I've addressed the requested changes? :)

alecjacobson (approved):

PR #1926, opened 2021-11-29
Fixed thread unsafe code in marching_cubes.cpp

stoperro:
marching_cubes used mutable static variable causing race condition in multi-threaded code. Removed static keyword to make it work properly.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/marching_cubes.cpp
+++ b/include/igl/marching_cubes.cpp
@@ -52,8 +52,8 @@ IGL_INLINE void igl::marching_cubes(
     const unsigned i = xyz2i(x,y,z);
 
     //Make a local copy of the values at the cube's corners
-    static Eigen::Matrix<Scalar,8,1> cS;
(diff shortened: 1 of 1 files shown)


alecjacobson:
lgtm I don't really remember why I had static there. Maybe it was a misguided idea that it'd be faster. In any case, thread safety is more important.

stoperro:
Static means the matrix there is allocated only once, so there might be small performance boost due to less runtime allocations... hence at least in one version I moved it outside loop so allocates only once and then overwrites values inside only (for the other I worried it would be error prone, but could likely move outside too).

jdumas:
Those matrices have a fixed size, so they would be stack-allocated. The `static` keyword here will not give any performance boost.

alecjacobson (approved):

PR #1962, opened 2022-01-19
Add +1 offset in writeOBJ for hybrid mesh

Shuenhoy:
This would keep the behavior the same as the non-hybrid situation.

Fixes # .

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/writeOBJ.cpp
+++ b/include/igl/writeOBJ.cpp
@@ -144,7 +144,7 @@ IGL_INLINE bool igl::writeOBJ(
 
     for(const auto& vi : face)
     {
-      s<<" "<<vi; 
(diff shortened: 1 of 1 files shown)


alecjacobson (approved):
shocking that this bug exists. seems like no one could have been using it. vector-of-vector is super convenient. We should also consider adding the vectorized polygon-mesh version (see, e.g., https://github.com/libigl/libigl/blob/main/include/igl/readOBJ.h#L125)

Shuenhoy:
Hi @alecjacobson, the CI failed for the initial commit. I just merged the upstream, please re-approve the CI, thanks.

PR #1985, opened 2022-02-24
Compile GMP and MPFR from source

alecjacobson:
Previously we relied on gmp and mpfr to be installed system wide. This (work in progress) PR uses cmake's `ExternalProject_add` to download and compile gmp and mpfr from source in the build.

gmp and mpfr use `configure`/`make` style build which seems well supported by `ExternalProject_add`. I've also borrowed some configure flags from homebrew's build of these libraries (otherwise they were not easy to compile following the gmp/mpfr documentation).

This affects (and appears to work correctly) on Mac and Linux. A nice feature of compiling these from source, is that it's more straightforward to compile for a different architecture (e.g., compiling for x86_64 and using Rosetta on an M1 Mac).

I don't recall why we _weren't_ compiling gmp/mpfr from source. Is the a reason besides this ExternalProject_add not existing/being written?

Windows continues to use precompiled binaries downloaded from CGAL.

> I really banged my head on getting the `ExternalProject_add` paths working correctly. I think @jdumas had put some thought into the organization of subdirectories in the `build/` folder. Would love to know if this can be similarly cleaned up. Currently it makes these `build/gmp-install`, `build/gmp-build`, etc. folders...


cmake/recipes/external/gmp.cmake
```diff
@@ -3,12 +3,46 @@ if(TARGET gmp::gmp)
...
+    --disable-shared
+  BUILD_COMMAND make -j${Ncpu}
+  INSTALL_COMMAND make -j${Ncpu} install
+  TEST_COMMAND ""
+  BUILD_BYPRODUCTS ${gmp_LIBRARY}
+  )
+  ExternalProject_Get_Property(gmp SOURCE_DIR)
+  set(gmp_LIBRARIES ${gmp_LIBRARY})
+  add_library(gmp::gmp INTERFACE IMPORTED GLOBAL)
+  file(MAKE_DIRECTORY ${gmp_INCLUDE_DIR})  # avoid race condition
```
jdumas, line 40: You can still get race condition if you try to compile a file that includes `<gmp.h>` before the external project is downloaded.

jdumas (commented):
The main reason I've avoided using `ExternalProject_Add` for gmp/mpfr on Linux/macOS is that `ExternalProject_Add` is really more suited for a two-stage CMake build. The typical approach to using `ExternalProject_Add` with CMake is as follows:
1. Run a "meta" CMakeLists.txt that uses `ExternalProject_Add` to download, compile & install all third-party libs into a desired target folder.
2. Run your own project's CMakeLists.txt and use `find_package()` to import the target you have just installed in the previous step.

This requires the user to run `cmake && make` two times for the project, and I really don't like this approach.

The problem with trying to use `ExternalProject_Add` like you're doing right now, is that it's not easy to express the dependencies between building the external lib, and building the other targets that depend on it. Let's imagine for a moment that building libigl is super fast, and you attempt to link your libig_copyleft_cgal.a against gmp.a, but your external project hasn't finished building gmp. Then you're gonna get an error. Given how fast it is to build gmp/mpfr, and how slow it is to build libigl, this is not gonna happen very often. But looks like you've had similar issues with the include headers.

I'm pretty sure it's possible to express the dependency between `gmp::gmp` and the external project, to make sure the external lib is downloaded & compiled before anyone tries to use it. But this will require some additional logic from what I can tell.

cmake/recipes/external/gmp.cmake
alecjacobson: This is not for a race condition at `make` time, but rather at `cmake` time. The following line:
```
target_include_directories(gmp::gmp INTERFACE ${gmp_INCLUDE_DIR})
```
will fail if that include directory doesn't exist (it doesn't matter if the actual headers are there yet. just need the path to exist).

alecjacobson:
I haven't run into that issue yet. Are you sure the dependencies aren't handled correctly?

Consider this mini project https://github.com/alecjacobson/gmp-mpfr-externalproject_add

If I run `make -j8`, then it still waits for gmp and mpfr to build before compiling and linking the executable.

cmake/recipes/external/gmp.cmake
alecjacobson: (this comment `# avoid race condition` is from a copy-paste. I don't really agree that it's a race condition. It has nothing to do with threads AFAICT. It's rather just that the directory doesn't exist yet)

jdumas:
Ah you're right, it looks like the line `add_dependencies(gmp::gmp gmp)` does the trick of expressing the dependency properly at CMake level. I think this should work then! Guess the reason we didn't have this before is because nobody did it :)

alecjacobson:
Thanks. Do you have any suggestions about the build sub-directories organization? I got super confused trying to use `PREFIX` / `DOWNLOAD_DIR` / `SOURCE_DIR` etc. but it seems possible to clean this up or move everything into the `_deps` folder that gets created.

jdumas:
If you `include(FetchContent)`, you can use `${FETCHCONTENT_BASE_DIR}` to retrieve the location of the `_deps/` folder.

alecjacobson:
fwiw, this worked successfully for my immediate use case: compile libigl-based mex functions and running them on an M1 mac running Matlab using Rosetta.

jdumas:
Make sure you remove gmp/mpfr's install for Linux/macOS on GitHub Actions before merging this:
https://github.com/libigl/libigl/blob/1f44d1e025040df06f82dde648741909b17afae1/.github/workflows/continuous.yml#L52
https://github.com/libigl/libigl/blob/1f44d1e025040df06f82dde648741909b17afae1/.github/workflows/continuous.yml#L60

PR #1997, opened 2022-03-07
Use URL to download CGAL source rather than git

alecjacobson:
CGAL's git repository is ginormous (> 1GB). cmake's FetchContent has the `GIT_SHALLOW` but this still results in a huge download. If we're using a release of cgal then there's not much advantage to use git anyway. Hence, switch to url fetch → significantly smaller download ~35MB.


jdumas:
`GIT_SHALLOW` only works if the `GIT_TAG` points to a proper release tag rather than a SHA1. In this case, the amount of data downloaded should be about the same as with a tar.gz URL (unless they prune additional stuff when packaging releases).

PR #2001, opened 2022-03-14
Reduce dynamic allocations for AABB queries

svenpilz:
Regarding discussion https://github.com/libigl/libigl/discussions/1959.

Reduces dynamic allocations which would reduce performance on Emscripten/WebAssembly builds when multi-threading is enabled. Enabling multi-threading on that platform makes `malloc`/`free` atomic. 

#### Checklist

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


include/igl/ray_mesh_intersect.cpp
```diff
@@ -12,6 +12,41 @@ extern "C"
 #include "raytri.c"
 }
 
+namespace igl {
+template <
+  typename DerivedV,
+  typename DerivedF>
+IGL_INLINE bool ray_triangle_intersect(
+  Eigen::Vector3d s_d,
```
alecjacobson, line 20: should these be const references rather than non-const pass-by-copy?

include/igl/ray_mesh_intersect.cpp
```diff
@@ -12,6 +12,41 @@ extern "C"
 #include "raytri.c"
 }
 
+namespace igl {
+template <
+  typename DerivedV,
+  typename DerivedF>
+IGL_INLINE bool ray_triangle_intersect(
```
alecjacobson, line 19: should document this in the header.

include/igl/ray_mesh_intersect.cpp
```diff
@@ -62,21 +90,32 @@ template <
...
   igl::Hit & hit)
 {
-  std::vector<igl::Hit> hits;
-  ray_mesh_intersect(source,dir,V,F,hits);
-  if(hits.size() > 0)
+  if (F.rows() == 1)
   {
-    hit = hits.front();
-    return true;
+    // Shortcut for AABB based queries.
```
alecjacobson, line 101: would it make more sense for the AABB code to call this one directly?

alecjacobson (changes requested):
Thanks! Overall, this is a good change. I left a couple notes, it'd be nice to hear your thoughts on.

include/igl/ray_mesh_intersect.cpp
svenpilz: Yes. I cannot remember why I did it like that. Will fix that.

I was also not sure about your usage of `auto` and if you like named variables to give some context, even if they are basically just temporaries. Like `auto d = v.cast<double>(); foo(d)` vs. `foo(v.cast<double>())`.

include/igl/ray_mesh_intersect.cpp
svenpilz: Guess it makes sense to also have the vectors's scalar type as a template parameter?

include/igl/ray_mesh_intersect.cpp
svenpilz: That is a good question. I haven't really looked at the `AABB` code. Do you think people would use this function here directly, could this be a common use case? If the `AABB` is the only caller with one triangle, guess adapting there would be the better place.

include/igl/ray_mesh_intersect.cpp
svenpilz: Added to the header.

alecjacobson:
hopefully the builds and tests work and we can merge this

svenpilz:
Could not reproduce the failing test on Linux. Tried GCC 10, and clang 13 in Debug with address and undefined sanitizer. @alecjacobson, do you have an idea here?

PR #2011, opened 2022-03-25
eigen → v3.4.0

### LINKED ISSUE #1974: Bump Eigen to 3.4
We're currently using Eigen 3.3.7 (Dec 2018). Eigen's newest tagged release 3.4.0 from Aug 2021 has a bunch of nice features. License-wise it's also quite nice that the SimplicialCholesky solvers [have changed from LGPL to MPL2 ](https://gitlab.com/libeigen/eigen/-/commit/1dc1677d525b1df3cc32b3df5cd816e150b07185).

I believe another great feature is slicing as discussed in https://github.com/libigl/libigl/issues/1485. Maybe we should enumerate some great features (reasons to do this besides obviously staying up to date) and any identified problems (we have been burned in the past by Eigen releases containing bugs, but admittedly that was in the distant past and Eigen seems much more stable).

alecjacobson:
Fixes #1974 

We've just released our v2.4 so now is a good time to bump eigen and get a lot of testing in before our next release. I'm particularly worried about performance gotchas that might be hard to catch. There are also probably plenty of new features of Eigen v3.4.0 that we could refactor current libigl functions to take advantage of.


jdumas:
FYI I have updated the github action stuff to fix the windows issues. You may want to update your PRs to merge with the main branch.

alecjacobson:
What can we do about this static build on windows? It's always running out of heap space...

jdumas:
It could be that even with our separate job pools, ninja may be running two parallel jobs (1 compilation + 1 link) at the same time. I've opened #2018 to limit global parallelism with ninja, let's see if it helps.

alecjacobson:
I'm starting to wonder if the bump in eigen is somehow exacerbating the Windows CI issues.

Issue #1539, opened 2020-06-04
Unexplainable deformation after using ARAP (tutorial 405)


mzanin:
#### Describe your question

The issue I'm having is that the resulting geometry after ARAP does not produce a shape that makes sense to me. I'm at a loss as to what the the reason for this could be and would really appreciate any input on what I could be doing wrong. Note I used tutorial 405 as a starting point on understanding how to make use of ARAP. 

Below is an image of the state of the source geometry before deformation with the selection indices on the geometry shown in purple (the strip on the rim of the geometry) and the boundary conditions shown as green points (if there is a better way to visualize this in igl please let me know).

![image](https://user-images.githubusercontent.com/1013551/83790116-f93cb800-a697-11ea-8580-73b4c660f876.png)

and below are some images of the state after deformation. Note the unwanted bulge which is not present in the original source geometry. 

![image](https://user-images.githubusercontent.com/1013551/83790207-1b363a80-a698-11ea-9cea-55d973739ee4.png)

![image](https://user-images.githubusercontent.com/1013551/83790285-402aad80-a698-11ea-84ee-f2fbcf070071.png)

Below is the simplified source code used to obtain these results. Some additional details -
* None of the energy types improve the final shape. 
* Lowering the maximum iterations to below 10 results in a better shape. The bulge is gradually introduced the more iterations there are.
* Enabling with_dynamics also improves the shape but I don't really understand what this does. All I know is that the boundary conditions are no longer used.

```
int main(int argc, char *argv[])
{
    Eigen::MatrixXd vertices, initial_guess_vertices;
    Eigen::MatrixXi faces;
    Eigen::VectorXi selected_indices;
    igl::ARAPData arap_data;

    // the mesh which has already been transformed by a procrustes
    igl::readOFF(TUTORIAL_SHARED_PATH "/template.off", vertices, faces);
    
    // the mesh used for the initial guess which has undergone a procrustes transform
    // and a laplacian

    igl::readOFF(TUTORIAL_SHARED_PATH "/deformed.off", initial_guess_vertices, faces);
    igl::readDMAT(TUTORIAL_SHARED_PATH "/selection.dmat", selected_indices);
   
    // the boundary conditions are already defined in the deformed.off geometry so just 
    // use the selected indices to populate the boundary_conditions
    Eigen::MatrixXd boundary_conditions(selected_indices.size(), initial_guess_vertices.cols());
    for (int i = 0; i < selected_indices.size(); i++)
    {
[...]

PR #2019, opened 2022-03-30
Avoid converting exact input to inexact when detecting intersections

alecjacobson:
The previous `remesh_self_intersections.cpp` and `intersect_other.cpp` had a branch

```
if detect_only
  use Epick to find intersections
else
  use Epeck to find and resolve intersections
```

This _only_ depended on the boolean flag `detect_only`. Unfortunately, that meant if the caller's input vertices were in the exact type `CGAL::Epeck::FT` and `detect_only=true`, then the input would be converted (i.e., rounded) to `CGAL::Epick::FT` (aka `double`) before looking for intersections. This leads to **false positives**.

This PR replaces the branch with `if detect_only and not Epeck input`.

Hopefully this merges easily so I can get it in before finishing #1539 1895


alecjacobson:
Added unit test.

PR #2030, opened 2022-05-01
point GMPXX_INCLUDE_DIR to locally built gmp

### LINKED ISSUE #2029: cmake using -isystem for cgal includes and finding homebrew install instead?
On my local mac, I noticed that the cgal tests were failing. After some long debugging I traced it down to the compilation paramters:

```
-isystem /Users/ajx/Repos/libigl/build-header-only/_deps/cgal-src/include
```

if I replaced this with 

```
-I /Users/ajx/Repos/libigl/build-header-only/_deps/cgal-src/include
```

Everything works. My _guess_ is that [because isystem includes come later](https://stackoverflow.com/a/2579600/148668), the compiler was finding the cgal headers in `/usr/local/include/` where homebrew had put them. This compiled fine, but resulted in memory issues and crashes at runtime. I imagine due to a CGAL mismatch.

The easy fix for me was to uninstall cgal from homebrew. However, I wonder if this is something we could be doing better to prevent via our cmake scripts?

I'm not entirely sure why we use `isystem` for cgal or if that's something we're choosing or cgal is choosing. I believe `isystem` vs `I` also affects whether compilation warnings are displayed, but if that's the only reason then I'd rather see warnings than this hard to debug crash.

alecjacobson:
Fixes #2029

gmpxx is part of gmp and we're building it via cmake. Previously we weren't telling cgal where to find it and on some machines (mine at least) it was finding it at /usr/local/include. This caused /usr/local/include to be added to include paths which caused #2029 to happen (cgal was also found there and confused).

This should tell cgal where to find gmpxx and subsequently _not_ add /usr/local/include.

### DIFF
--- a/cmake/recipes/external/cgal.cmake
+++ b/cmake/recipes/external/cgal.cmake
@@ -34,8 +34,10 @@ function(cgal_import_target)
     include(boost)
 
     ignore_package(GMP 5.0.1)
-    set(GMP_INCLUDE_DIR "")
+    set(GMP_INCLUDE_DIR ${gmp_INCLUDE_DIR})
     set(GMP_LIBRARIES gmp::gmp)
+    set(GMPXX_INCLUDE_DIR ${GMP_INCLUDE_DIR})
+    set(GMPXX_LIBRARIES ${GMP_LIBRARIES})
 
     ignore_package(MPFR 3.0.0)
     set(MPFR_INCLUDE_DIR "")
--- a/cmake/recipes/external/cgal.cmake
+++ b/cmake/recipes/external/cgal.cmake
@@ -34,9 +34,10 @@ function(cgal_import_target)
     include(boost)
 
     ignore_package(GMP 5.0.1)
-    set(GMP_INCLUDE_DIR ${gmp_INCLUDE_DIR})
+    set(GMP_INCLUDE_DIR "")
     set(GMP_LIBRARIES gmp::gmp)
-    set(GMPXX_INCLUDE_DIR ${GMP_INCLUDE_DIR})
+    ignore_package(GMPXX 5.0.1)
+    set(GMPXX_INCLUDE_DIR "")
     set(GMPXX_LIBRARIES ${GMP_LIBRARIES})
 
     ignore_package(MPFR 3.0.0)
--- a/cmake/recipes/external/cgal.cmake
+++ b/cmake/recipes/external/cgal.cmake
@@ -36,7 +36,6 @@ function(cgal_import_target)
     ignore_package(GMP 5.0.1)
     set(GMP_INCLUDE_DIR ${gmp_INCLUDE_DIR})
     set(GMP_LIBRARIES gmp::gmp)
-    ignore_package(GMPXX 5.0.1)
     set(GMPXX_INCLUDE_DIR ${GMP_INCLUDE_DIR})
     set(GMPXX_LIBRARIES ${GMP_LIBRARIES})
 
--- a/cmake/recipes/external/cgal.cmake
+++ b/cmake/recipes/external/cgal.cmake
@@ -34,10 +34,10 @@ function(cgal_import_target)
     include(boost)
 
     ignore_package(GMP 5.0.1)
-    set(GMP_INCLUDE_DIR "")
+    set(GMP_INCLUDE_DIR ${gmp_INCLUDE_DIR})
     set(GMP_LIBRARIES gmp::gmp)
     ignore_package(GMPXX 5.0.1)
-    set(GMPXX_INCLUDE_DIR "")
+    set(GMPXX_INCLUDE_DIR ${GMP_INCLUDE_DIR})
     set(GMPXX_LIBRARIES ${GMP_LIBRARIES})
 
     ignore_package(MPFR 3.0.0)
--- a/cmake/recipes/external/gmp.cmake
+++ b/cmake/recipes/external/gmp.cmake
@@ -17,7 +17,10 @@ else()
   set(prefix ${FETCHCONTENT_BASE_DIR}/gmp)
   set(gmp_INSTALL ${prefix}/install)
   set(gmp_LIB_DIR ${gmp_INSTALL}/lib)
-  set(gmp_LIBRARY ${gmp_LIB_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}gmp${CMAKE_STATIC_LIBRARY_SUFFIX})
+  set(gmp_LIBRARY 
+    ${gmp_LIB_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}gmp${CMAKE_STATIC_LIBRARY_SUFFIX}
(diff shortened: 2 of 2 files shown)


cmake/recipes/external/cgal.cmake
```diff
@@ -34,8 +34,10 @@ function(cgal_import_target)
     include(boost)
 
     ignore_package(GMP 5.0.1)
-    set(GMP_INCLUDE_DIR "")
+    set(GMP_INCLUDE_DIR ${gmp_INCLUDE_DIR})
     set(GMP_LIBRARIES gmp::gmp)
+    set(GMPXX_INCLUDE_DIR ${GMP_INCLUDE_DIR})
```
jdumas, line 39: I wonder if we should add `ignore_package(GMPXX 5.0.1)` as well just to be safe?

cmake/recipes/external/cgal.cmake
```diff
@@ -34,8 +34,10 @@ function(cgal_import_target)
     include(boost)
 
     ignore_package(GMP 5.0.1)
-    set(GMP_INCLUDE_DIR "")
+    set(GMP_INCLUDE_DIR ${gmp_INCLUDE_DIR})
```
jdumas, line 37: This is not necessary, since our target `gmp::gmp` should carry the include dir already.

alecjacobson: I think CGAL's cmake is looking at `GMPXX_INCLUDE_DIR` so it needs to be explicitly set.

cmake/recipes/external/cgal.cmake
alecjacobson: Do you know what this is doing? If it makes sense to add, then happy to.

cmake/recipes/external/cgal.cmake
alecjacobson: Confirmed, if `GMPXX_INCLUDE_DIR` is set to `""` then CGAL determines that `CGAL_WITH_GMPXX:BOOL=OFF`. By setting `GMPXX_INCLUDE_DIR` explicitly then we keep `CGAL_WITH_GMPXX:BOOL=ON`.

cmake/recipes/external/cgal.cmake
alecjacobson: (adding this doesn't seem to hurt)

jdumas: `ignore_package()` is a function I wrote in this `cgal.cmake`. Basically it will add an empty `GMPXXConfig.cmake` file, thus preventing CGAL from using its own `FindGMPXX.cmake` file (and subsequently adding `/usr/local/include` to your include paths).

cmake/recipes/external/cgal.cmake
jdumas: If you add `ignore_package(GMPXX 5.0.1)`, you can probably leave `GMP_INCLUDE_DIR` and `GMPXX_INCLUDE_DIR` set to `""`, since the fake `GMPXXConfig.cmake` will set `GMPXX_FOUND` to `ON`, and `CGAL_WITH_GMPXX` should subsequently be set to `ON`.

alecjacobson: ah, got it. Didn't realize these were related. I pushed the change.

alecjacobson: augh... that broke Windows. Is there anything wrong with using the previous version with GMPXX_INCLUDE_DIR explicitly set? At least windows was building...

alecjacobson: alecjacobson/gmpxx-include-dir

alecjacobson: augh, that broke windows. Anything wrong with explicitly setting GMP_INCLUDE_DIR and GMPXX_INCLUDE_DIR ? That was at least working on windows...

alecjacobson: hmm, actually seems that the ignore_package is causing trouble on windows...

jdumas: Well that's because gmpxx is **not** supported on Windows. So actually we might want to disable GMPXX directly by using [CMAKE_DISABLE_FIND_PACKAGE_<PackageName>](https://cmake.org/cmake/help/v3.16/variable/CMAKE_DISABLE_FIND_PACKAGE_PackageName.html#variable:CMAKE_DISABLE_FIND_PACKAGE_%3CPackageName%3E). I.e. don't add `ignore_package(GMPXX ...)`, do not set `GMPXX_INCLUDE_DIR` or `GMPXX_LIBRARIES`, but add the following line before including CGAL:
```cmake
set(CMAKE_DISABLE_FIND_PACKAGE_GMPXX ON)
```

alecjacobson: so, when the windows CI was succeeding was CGAL just not enabling gmpxx? Why this extra code to explicitly disable it then?

jdumas: Nope. GMPXX is an optional dependency of CGAL, so if CGAL doesn't find GMPXX it continues with its own internal wrapper around gmp objects. If we force CGAL into pretending we found GMPXX (via our `ignore_package(GMPXX)`) then it will fail when it actually tries to include `gmpxx.h`.

alecjacobson: Why would the CI succeed then?

alecjacobson: We're back to the state of the original PR which passes CI. If there's a more elegant way to handle GMPXX that maintains passing for Windows then let's merge it in a new PR. I'll merge this for now.

jdumas: What do you mean by why would the CI succeed? With your changes you are not preventing CGAL's FindGMPXX from executing. But because you set the `GMPXX_INCLUDE_DIR` variable in advance, CGAL is using those locations. On Windows, it will still go through this [find_path()](https://github.com/CGAL/cgal/blob/a8e87526348d2d1faa2ea893747314853ee6868d/Installation/cmake/modules/FindGMPXX.cmake#L20) command and silently fail to find GMPXX. On Linux/macOS, I am not sure whether it find them or not.

In the end I suppose your setup kinda works, though you may think you're using GMPXX on Windows when you are not.

alecjacobson: Right. In the merge version `GMPXX_INCLUDE_DIR` is set to point to the gmp build for all systems, and I guess on windows cgal will realize that gmpxx should be off so it's ignored. Using the `ignore_package` caused windows build to fail out of the box. If there's a solution that avoids pretending that gmpxx is really there then we should merge it, but I'm not sure what that is and I end up pulling my hair out trying to debug the windows build.

jdumas: Well the solution is simple:
```cmake
if(WIN32)
    set(CMAKE_DISABLE_FIND_PACKAGE_GMPXX ON)
else()
    ignore_package(GMPXX 5.0.1)
    set(GMPXX_INCLUDE_DIR "")
    set(GMPXX_LIBRARIES ${GMP_LIBRARIES})
endif()
```

The end result should be basically the same though...

alecjacobson: 😊 Want to PR this and merge?

PR #2031, opened 2022-05-01
Templates in CGAL module

alecjacobson:
Windows can be temperamental with cgal templates so I'm using this PR to run CI (already passing on my local mac and linux).


alecjacobson:
The CI cache seems to work really well for Linux (Release, Static) but doesn't seem to make a dent in the others... How come? It's a shame that adding a template or two in a few files triggers hours worth of recompiling.

jdumas:
Yeah the cache doesn't seem to be saved/loaded properly on macOS and Windows. I have no idea why this happens, and I don't really have time to investigate at the moment... :/

PR #2057, opened 2022-07-12
Fix `blue_noise` and `random_points_on_mesh` reproducibility

### LINKED ISSUE #2053: blue_noise reproductibility
`blue_noise` produces different results for consecutive calls.

#### Example
```
#include <igl/blue_noise.h>

Eigen::MatrixXd computeBlueNoisePositions(const Eigen::MatrixXd& V, const Eigen::MatrixXi& F)
{
	Eigen::MatrixXd P_blue;
	const double r = 5.0;
	Eigen::MatrixXd B;
	Eigen::VectorXi I;
	igl::blue_noise(V,F,r,B,I,P_blue);
	return P_blue;
}

{
	Eigen::MatrixXd V;
	Eigen::MatrixXi F;
	igl::read_triangle_mesh( "bunny.obj",V,F);

	Eigen::MatrixXd P_blue1 = computeBlueNoisePositions(V, F);
	Eigen::MatrixXd P_blue2 = computeBlueNoisePositions(V, F);

	std::cout << "P_blue1\n" << P_blue1 << std::endl;
	std::cout << "P_blue2\n" << P_blue2 << std::endl;
	// P_blue1 and P_blue2 are different but expected to be the same

	return 1;
}
```
Result
```
P_blue1
  1.19629    6.2217   0.42864
  -1.7073   1.92196 -0.344443
 -4.39213   6.43616 0.0762281
P_blue2
 -4.24073   6.04523 -0.120284
  1.60464   4.71455   2.04362
 -1.05383   2.19985   -1.3946
```
It would be expected to have the same result between two successive calls, but they are different, even when forcing the seed of `rand()` with `srand` (which is not desirable on itself).
However, the result is the same between two executions, eg first call of the execution is always the same, second call is always the same but first and second are different.

In my example I used this bunny.obj model (https://pastebin.com/nnPB5ZZb ) but the problem is appears with other models as well.

I found that a similar issue was reported here https://github.com/libigl/libigl/pull/1597#pullrequestreview-473050272 and here https://github.com/libigl/libigl/issues/1613, but the PR was merged eventually, 

Is there a way to circumvent this problem and get consistent results ?

Thank you

#### Platform
Linux Debian 11.3, libigl 2.4.0 header only, clang++-14.0.5

seb-tourneux:
Add new overloads for `blue_noise` and `random_points_on_mesh` for that accept an UnformRandomBitGenerator as input. The default signature (ie without URBG) still works, and uses by default `std::minstd_rand` initialized with random seed generated by `std::rand()`. This is following the existing behaviour of `randperm`.
Corrects the previous uses of `rand()` and `VectorXs::Random` that forced the use of `srand()` and thus preventing the use of a local seed.

Add reproducibility test case for `blue_noise` and `random_points_on_mesh`

Fixes #2053 .

#### Checklist

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Thanks! Honestly I would drop the overload with the `std::rand()` initialization. People shouldn't be tempted to use non-deterministic functions.

seb-tourneux:
I didn't know about introducing breaking changes, but yes, I agree.

jdumas:
We can leave an overload that uses a default-initialized RNG.

alecjacobson:
Transferring to https://github.com/libigl/libigl/pull/2235

PR #2069, opened 2022-08-13
Tetgen: fix bug when faces don't have markers

alecjacobson:
I'm not sure why this `if` was here in the first place. This function doesn't extract markers. (seems either qnzhou or daneielpanozzo wrote the original line).  Before this fix, if the face list existed but markers did not, then it would crash. Now, it ignores the possibility of markers and extracts existing faces.

(Properly extracting markers is left for a future PR).


qnzhou:
Looks good to me.  I have no clue about the reason for `if`.

PR #2079, opened 2022-09-07
fix nan bug in gaussian_curvature when vertex coincidence

n-jing:
Fixes # .

fix nan bug in gaussian_curvature when vertex coincidence

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
It sounds like nans are happening for meshes with degenerate triangles. In this case, κ=0 seems arbitrary and hides the issue. If a caller wants to ignore these nans, my proposal is that they handle it on their end. Hitting nans is actually reasonable and will help track the issue. Would be happy to merge a PR that improves the documentation in the header.

PR #2090, opened 2022-10-05
Update715_MeshImplicitFunction

### LINKED ISSUE #2088: 715_MeshImplicitFunction sample  is outdated
#### Describe the bug
I'm not sure if this should be consider a bug, but it is not a request or something like that too, since it works but it is not correct either.

715_MeshImplicitFunction sample has in its code:

` printf("  %5f secs dual contouring\n",t_Gf+tictoc());
      // Could use igl::marching_cubes once
      // https://github.com/libigl/libigl/pull/1687 is merged
      tictoc();
      igl::copyleft::marching_cubes(Gf,GV,GI,mcV, mcF);
      printf("  %5f secs marching cubes\n",t_Gf+tictoc());`

Since that branch has been merged probably that code should be updated to use igl::marching_cubes

MotivaCG:
Replace the copyleft marching_cubes function with native the libigl one.

Fixes #2088 .

It works as before but without the copyleft dependency

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/tutorial/715_MeshImplicitFunction/main.cpp
+++ b/tutorial/715_MeshImplicitFunction/main.cpp
@@ -1,4 +1,4 @@
-#include <igl/copyleft/marching_cubes.h>
+#include <igl/marching_cubes.h>
 #include <igl/unique_simplices.h>
 #include <igl/dual_contouring.h>
 #include <igl/get_seconds.h>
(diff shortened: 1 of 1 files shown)


FrankEscobar (commented):
Changes from copyleft marching_cubes to igl marching_cubes

PR #2117, opened 2022-11-07
glfw viewer - high dpi: handling both width and height.

fghoussen:
Fixes # .

From what I know of OpenGL, both width and height must be handled to resize window to frame.
Not 100% sure!...I PR that as a proposal: I let you decide if it's a good move

#### Checklist

`make all test` runs OK at my side (debian)

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/include/igl/opengl/glfw/Viewer.cpp
+++ b/include/igl/opengl/glfw/Viewer.cpp
@@ -47,7 +47,8 @@
 
 // Internal global variables used for glfw event handling
 static igl::opengl::glfw::Viewer * __viewer;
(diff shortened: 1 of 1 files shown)


alecjacobson:
Is this fixing a known issue or something that theoretically could happen? Wouldn't that mean that someone's display has a different resolution in height vs width?

fghoussen:
I have not the HW to test this, so, I can not say this has been tested.
But this appears in many OpenGL snippets, tutorials, stackoverflow questions I've seen so I reported it here.

PR #2131, opened 2022-12-19
Support find_package

justend29:
## Desires

1. provide the ability to install and then use all libigl targets through CMake's `find_package` command. Reasons for this include:
   - allowing users to separately build & install libigl
   - after the above step, reuse the installation where appropriate
2. provide the ability to acquire libigl's dependencies through CMake's `find_package` command. Reasons for this include:
    - allowing users to separately build dependencies or use previously built dependencies
    - remove obligation to fetch and rebuild all dependencies for every build directory of libigl, even if it's built with the exact same build options.
    - improve support for package managers. You can see [here](https://github.com/microsoft/vcpkg/tree/master/ports/libigl) how vcpkg has made patches for libigl to incorporate a subset of an old version of libigl into its ecosystem. I would like to upgrade the port after this.

## Existing Points of Friction

I'm just trying to be detailed here, not discount anything.

### WRT Desire 1

- The existing CMake system in libigl provided installation commands and exported the target *igl::core*. However, all other targets, such as *igl_copyleft::cgal*, could not be used via installation. Furthermore, there existed certain issues with this installation, preventing consuming the package through `find_package`. These include:
 - libigl's package config files were installed into the directory `.../igl/` instead of one named `.../libigl`. CMake's [search procedure](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) searches paths including the package name, so it could not find `.../igl/`
 - Eigen was automatically installed, and installed in the wrong path (eigen, not eigen3). Library installations rarely install dependencies, as this causes many duplicates of the library all across the system. Instead, they're commonly handled with `find_dependency`. The incorrect install path also prevented CMake from finding the installed Eigen, in the same manner as the mark above for libigl.
 -  The CMake package configuration file [libigl-config.cmake](https://github.com/libigl/libigl/blob/c35f0fee39195e72ea92d28ab7c10dfc0425b5d4/cmake/igl/libigl-config.cmake.in) didn't actually include any installed & exported targets. So, even when a consuming project finds libigl-config.cmake, no targets are available to use.
 - Header files with `.hpp` extension were disregarded by CMake and therefore weren't installed.
 
### WRT Desire 2

- Dependencies were invariably fetched and built from source as subprojects. No options existed to change this behaviour.

## Fixes

I tried to conform to the existing structure and behaviour of libigl, and not make sweeping or breaking changes.

- Other libigl targets are installed with appropriate CMake package config & targets files.
  - I took the approach of creating one root package config file (libigl-config.cmake) and a respective package config and target file for each library component. This makes opting into and out of components and defining internal dependencies (igl_copyleft::core requires igl::core) by simply including the appropriate config file ([within other config files](https://github.com/libigl/libigl/compare/main...justend29:cmake-package?expand=1#diff-7c8ad35788eab48ce90be3c776da6513c8dedce4d020265907fdea96a5a4dec3R4-R6)).
- files are installed into the appropriate directory (`.../libigl/`); those within CMake's search paths.
- Removed installation of Eigen, limited to [this commit](https://github.com/libigl/libigl/commit/7ea0b9d271330387ae132c75ebbe118b8aaab209)
- `.hpp` files are considered and installed
- Another option, `LIBIGL_FIND_PACKAGES`, has been added. Its default value is `OFF`, such that existing builds expecting *FetchContent* don't break.
  - When set, it will cause dependencies to be acquired through `find_package()` calls.
  - It doesn't first try `find_package()` and then resort to `FetchContent`. When `find_package()` fails, I personally expect a warning to indicate I made a typo/error, I don't want a surprise version to be downloaded. 
  - CMake 3.24 has new [variables on FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html#variables) that will first try `find_package()` then resort to `FetchContent`. These changes do not explicitly support/test that, but will not prevent it in the future, and fundamentally provide different behaviour.
   - Exceptions exist, and will still be downloaded:
     - **Catch2** A fairly old version is used in libigl, and since it's only a test dependency, and will only have PRIVATE linking against targets that don't get installed, it's probably no biggie. A version range could alternatively be specified in a `find_package()` call.
     - **libigl_tests_data** This project doesn't currently export its targets. This can change in the future.

## Testing

Automated tests have been added to install libigl within the current binary directory, then consume the installed package from a separate project. Both of these steps are registered as libigl tests.
The separate project is housed within libigl's tests, but is configured, built, and tested by separate CMake & CTest instances, invoked by the tests registered with libigl. This keeps the consuming project isolated to replicate how consuming projects would use an installed version of libigl.

You'll [notice](https://github.com/libigl/libigl/compare/main...justend29:cmake-package?expand=1#diff-712ee8013bd51d459ee5735de47eae7c6cd10c5150d3b8d3bc144512d49454eeR50-R52) that only `igl::core`, `igl::embree`, and `igl_copyleft::core` are consumed at this point. The internal tests within the separate project steal existing unit testing source code from libigl's tests.
I've run the tests locally and they all succeeded.

## Remaining Tasks

- None of the *restricted* components have been considered but need to be.
- `igl::predicates` has not been considered.
- Document using *libigl* through `find_package()`  beyond the automated test mentioned above.
- Install licenses.

## To the Reviewer

Before moving on to the remaining work, I would like to know if these changes are something the maintainers would even consider bringing upstream. If not, I will keep them to myself, but I believe they will have benefits beyond me (like vcpkg).

I would also like to know if a versioned installation is preferred instead of the single one, to support installing and using multiple versions of *libigl* within a single installation prefix. An example of this would be installing *libigl* globally on a system (`/usr`, `/opt`).

Should I bump the libigl version? SemVer style?

What remaining tasks are necessary as part of this PR as opposed no a separate one?

#### Checklist

- [ ] Adds new .cpp file.
- [ ] This is a minor change.


jdumas:
Hi. I'm not opposed to having such changes in libigl, however, I have a few concerns:

1. Please don't submit PR that mix CMake formatting and adding new features. This is very hard to review. The current CMake style doesn't adds space between `if` statements and parenthesis. This is consistent with the style in official CMake files, and I see no reason to change it.
2. Please add all the `-config.cmake.in` file in a new subfolder. I am concerned about maintainability of those files alongside with the usual FetchContent approach (which should be the default behavior).
3. I don't want to add a `if(LIBIGL_FIND_PACKAGES)` to the `cmake/recipes/external` files. Those files should be self-contained and independent of any libigl option by design, so it's easy to reuse between projects. I would rather have separate individual files in `cmake/recipes/modules/` that calls `find_package()` (and only that) if that is the mode that is selected (kinda like how we currently support Hunter for the core libigl module).
4. Have you looked into CMake's new [dependency providers](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html) mechanism to switch between `find_package()` and `FetchContent` provided dependencies? It's something I've been planning to investigate but haven't had the time to do so yet.
5. Any question about semantic versioning and version bump should be deferred to @alecjacobson, as he's the boss. However, libigl official tagged released are very infrequent in general.

justend29:
Thanks for taking a look. On your  concerns:

1. Sorry about that. I briefly opened libigl in Clion last night which has autoformatting for CMake. I'll remove those small changes.
2. I can totally add them to a separate folder. They can also be generated. The main differences are the dependencies each component requires, as defined by `fetch_dependency()` and to each other with `include(<path-to-other>-config.cmake)`. A simple generation function arguments for each could handle that.
3. Sure thing
4. I have looked into it, but haven't directly used these features. Originally, my changes were made in June, before it was a thing. 
In the case of libigl, we could create `find-package_provider.cmake` provider files (or whatever name) within the `cmake/recipes/` directory, which users could opt into. The only issue I see is the existing files in `cmake/recipes/external` perform additional steps following their calls to `FetchContent_MakeAvailable`. I'm not currently sure if these would be compatible if the dependency came pre-built or how to deal with those, yet. I will more concretely investigate this avenue with libigl.

jdumas:
For (2) let's try to get those files generated then. We can have a CMake function inspect the dependencies of a given libigl target and generate the files accordingly.

Also, for the sake of the reviewing process, I'd say we can try to separate the changes in two stages:

1. Making libigl dependencies retrievable via find_package() rather FetchContent.
2. Making libigl targets properly installable for packaging.

justend29:
I'll close this draft PR, make the changes we discussed (hopefully soon), and open the separate PRs you requested.

PR #2138, opened 2023-01-04
don't crash on brew update

alecjacobson:
CI is failing because I think `brew update` is sending a failing exit code. https://github.com/libigl/libigl/actions/runs/3832188849/jobs/6522212468

It seems everything is continuing to install fine but it doesn't like that it can't create python links (which I don't believe we're using).

This fix ([suggested here](https://github.com/actions/setup-python/issues/577)) just forces the update to return true. It seems that many others just run `brew install ccache` without `brew update`. Not sure if we still need `brew update` or if that was from when we used brew to get gmp and mpfr.

Another option would be to install ccache in some more reproducible way, but I don't have the bandwidth to figure that out at the moment.

### DIFF
--- a/.github/workflows/continuous.yml
+++ b/.github/workflows/continuous.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Dependencies (macOS)
         if: runner.os == 'macOS'
         run: |
-          brew update
+          brew update || true
           brew install ccache
 
       - name: Cache Build
--- a/.github/workflows/continuous.yml
+++ b/.github/workflows/continuous.yml
@@ -56,7 +56,7 @@ jobs:
         if: runner.os == 'macOS'
         run: |
           brew update || true
-          brew install ccache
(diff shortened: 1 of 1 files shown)


alecjacobson:
huh. It's still failing. I guess I don't understand what's being run when. Is it failing on `brew update` or `brew install ccache`?

alecjacobson:
Ok, so after a bunch of experiments, it seems like the easiest fix is to just avoid running `brew update`. I'm not sure what the pitfalls of that are. `brew update` was pretty slow, too, so it's nice if we can avoid doing it.

The [other PR](https://github.com/libigl/libigl/pull/2139) is trying to download a ccache binary. But even if that works it seems inferior to using a package manager in some respects.

alecjacobson:
I'm going to go ahead and merge this for now so our CI continues. If I can grab @jdumas 's ear at some point maybe we can do something better.

jdumas:
I think this solution is as good as any tbh.

PR #2139, opened 2023-01-04
dont use brew; download binary from github

alecjacobson:
Makes https://github.com/libigl/libigl/pull/2138 obsolete. Download precompiled and hosted ccache binary directly.


alecjacobson:
oof. I sure don't understand those compilation errors.

PR #2144, opened 2023-01-09
add , 0 to index into area which supports different version of Eigen

fwilliams:
The local version of Eigen I am using does not support indexing Nx1 Matrices as `M(idx)` and requires `M(idx, 0)`. The fast winding number code breaks because of this. This PR adds changes `A(idx)` to `A(idx, 0)` in fast_winding_numbers.cpp. 

This shouldn't require any new tests and should be mergeable assuming unit tests pass.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/fast_winding_number.cpp
+++ b/include/igl/fast_winding_number.cpp
@@ -7,12 +7,12 @@
 #include <cassert>
 
 template <
-  typename DerivedP, 
-  typename DerivedA, 
+  typename DerivedP,
+  typename DerivedA,
   typename DerivedN,
-  typename Index, 
-  typename DerivedCH, 
-  typename DerivedCM, 
+  typename Index,
+  typename DerivedCH,
+  typename DerivedCM,
   typename DerivedR,
   typename DerivedEC>
 IGL_INLINE void igl::fast_winding_number(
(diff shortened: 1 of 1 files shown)


alecjacobson:
Is that a new version of Eigen?

I don't know if it's important here, but one reason to use `M(idx)` instead of `M(idx,0)` is that `M(idx)` will work for both Nx1 and 1xN. A lot of times our templates are generous enough to accept both.

fwilliams:
It fails on both Eigen 3.4 and 3.3.7

alecjacobson:
oh no. aren't we using 3.4.0? Is this not being caught be CI?

alecjacobson:
Could you post a little example with your compilation error?

alecjacobson:
I think I hit something similar using numpyeigen and I wonder if it's related to your issue. https://github.com/fwilliams/numpyeigen/blob/96c6356102a9f6c6eca5dd5446c4513552c12da9/src/npe.h defines maps only on Matrix types (e.g., `Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>`) but some libigl functions are expecting that an input argument is a `Eigen::*Vector*` type (i.e., at _compile time_ one of the dimensions is `1`).

alecjacobson:
And if so, then it seems an easy fix is if your input is `X` then call the function with `X.reshaped()`

alecjacobson:
I'm closing this for now. I think the correct solution is to change NPE to support compile-time "vector" types and pass them as expected to libigl. This way libigl continues to work with both "row vectors" and "column vectors".

PR #2155, opened 2023-01-26
Shadow Mapping in igl::opengl::glfw::Viewer

alecjacobson:
Shadows!

**Before:**
Screen Shot 2023-01-25 at 6 18 39 PM
**After:**
Screen Shot 2023-01-25 at 6 18 34 PM

Pretty straightforward shadow mapping implementation for directional lights. I'm guessing this could be optimized, but I'd like to have something in there for visualization of things where close contact is important, like, contact simulations.

I left a few notes about the funny coordinate system for our current positional lights. I didn't change that yet, but I think if we upgrade this to handle point lights then it'd worth reconsidering the coordinate system lights live in.

 For now lights are turned off by default and can be turned on with `'S','s'`. That'll also switch to directional lights.

I kinda like the floor in those renderings, but for now it's just ad hoc for the new tutorial entry. (actual tutorial entry to follow in that repo). Try it out in `113_Shadows`

### DIFF
--- a/include/igl/opengl/MeshGL.cpp
+++ b/include/igl/opengl/MeshGL.cpp
@@ -264,17 +264,29 @@ R"(#version 150
   out vec4 Kai;
   out vec4 Kdi;
   out vec4 Ksi;
+  uniform mat4 shadow_view;
+  uniform mat4 shadow_proj;
+  uniform bool shadow_pass;
+  uniform bool is_shadow_mapping;
+  out vec4 position_shadow;
 
   void main()
   {
     position_eye = vec3 (view * vec4 (position, 1.0));
-    normal_eye = vec3 (normal_matrix * vec4 (normal, 0.0));
-    normal_eye = normalize(normal_eye);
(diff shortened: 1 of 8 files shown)


alecjacobson:
**Before:**
Screen Shot 2023-01-27 at 12 50 29 AM
**After**
Screen Shot 2023-01-27 at 12 50 33 AM

In general, raw matcaps won't look very good with hard shadows. _If_ the matcap has an well-approximate-able light-direction _and_ has a near constant "dark-side", then it can look alright. For now, I don't have C++ code to approximate light directions or filter matcaps, but if someone has those, then this PR's shader will now handle them alright. The example above uses `light_direction = -0.2003    0.3390    0.9192` and 

matcap-1-shadow

Another example.

https://user-images.githubusercontent.com/2241689/215019706-3a2e7e9f-85d8-4284-a0f8-b7403b0dcd0b.mov

Using `0.3016    0.2469    0.9209` and ok-45-shadow

PR #2109, opened 2022-10-24
Fast mesh-to-mesh intersection and mesh self intersection without CGAL

vfonov:
Added code for fast triangle-triangle intersection checking and function for fast detection of mesh self-intersections and mesh-to-mesh intersections without CGAL.

Triangle-triangle intersection code is based on https://github.com/erich666/jgt-code/blob/master/Volume_08/Number_1/Guigue2003/tri_tri_intersect.c 

Functions for fast mesh self-intersection detection and mesh-mesh intrersections is using AABB and triangle-triangle intersection functions. 

I added unit tests for mesh-mesh intersection and mesh-self intersection.
Also added two tutorial programs. 

#### Checklist

- [X] All changes meet [libigl style-guidelines](https://libigl.github.io/style-guidelines/).
- [X] Adds new .cpp file.
- [X] Adds corresponding unit test.
- [ ] This is a minor change.

PR #2161, opened 2023-02-04
Update continuous.yml to use -j2 on Windows

alecjacobson:
@jdumas , git blame indicates you original added this line. Would love your review to know if using just one core is deliberate and important. [This github page](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) seems to indicate that 2 cores are available.

This might help solve the windows compilation issues for this pending PR: https://github.com/libigl/libigl/pull/2109


jdumas:
Well I've had a bunch of time where the linking step gets out of memory on Windows due to the two concurrent linking processes. Even with the job pools defined previously this happened from time to time. If you get a stable CI with 2 cores on Windows I'd say go for it though, it will be faster indeed...

alecjacobson:
Thanks, I guess let's see how it goes.

PR #2175, opened 2023-02-23
Allow use of external Eigen3 build indicated by Eigen3_DIR

KeithBallard:
In projects where Eigen3 is built externally (for other dependencies also linked in), it is preferrable to use the same source. Specifying the CMake variable Eigen3_DIR at configuration will trigger a search for Eigen3 in that directory. If successful, Eigen will not be downloaded as a dependency. If unsuccessful, the download will still occur.  If Eigen3_DIR is blank, no search occurs.

Fixes # .

#### Checklist

- [ ] Adds new .cpp file.
- [ ] This is a minor change.

### DIFF
--- a/cmake/recipes/external/eigen.cmake
+++ b/cmake/recipes/external/eigen.cmake
@@ -2,6 +2,16 @@ if(TARGET Eigen3::Eigen)
     return()
 endif()
 
+if(Eigen3_DIR)
+    message(STATUS "Looking for Eigen in: ${Eigen3_DIR}")
+    find_package(Eigen3 CONFIG HINTS ${Eigen3_DIR})
+    message(STATUS "Found Eigen in given directory: ${Eigen3_FOUND}")
+
+    if(TARGET Eigen3::Eigen)
+        return()
+    endif()
+endif()
+
 message(STATUS "Third-party: creating target 'Eigen3::Eigen'")
 
 include(FetchContent)


jdumas:
Sorry if you want to use an external build of Eigen, you need to provide your own `Eigen3::Eigen` target before including libigl. libigl itself should not try to call `find_package()`.

KeithBallard:
@jdumas I respectively disagree that this is possible in all cases.

I would like to build IGL as a static library as a standalone project.  Then I would like multiple of my projects to be able to reuse that installation, and it should be noted that both projects use libraries that also depend on Eigen.  Consequently, to make everything play nice, I build/install Eigen.

I see that in v2.4 the build system has been revamped.  Was it the intent to preclude external builds of IGL that could be reused?

jdumas:
Hmm I see. Install support has never really been a strong suit of libigl. There was a previous attempt at thisby @justend29 in #2131, but it never really went anyway. I think that the correct solution would be to use CMake's [dependency provider](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html) feature. Alternatively, we could use a different subfolder in `cmake/recipes`, similar to how we [find Eigen with Hunter](https://github.com/libigl/libigl/blob/90464ffbc3534578d521f06f490d0f61c4a0973d/cmake/recipes/hunter/eigen.cmake#L2), but "without" the Hunter part. In any case I think this setting should be [global](https://github.com/libigl/libigl/blob/90464ffbc3534578d521f06f490d0f61c4a0973d/CMakeLists.txt#L48-L52) (either _all_ dependencies are fetched via FetchContent, or none of them are). Relying on the presence of a `Eigen3_DIR` variable for that switch seems too brittle.

justend29:
I haven't forgotten about #2131, I just haven't had the time for the changes you requested. We've been using the fork I made with the changes proposed in #2131 at work, so there's been no dire need to circle back. I'm still up for them, though - I'll get there!

KeithBallard:
@jdumas First, I appreciate the response.  I do agree that it is too brittle to treat one dependency in a special manner.  Eigen3_DIR was chosen as the variable since find_package() creates a <package>_DIR variable.  However, I do think in general that find_package should be attempted first before a call to FetchContent.

Regarding the install support...I realized after this pull request that the targets are not exported on install.  #2131 is surprisingly close to what would be needed!  Clearly, I should have searched the issues first.

I do hold the opinion that FetchContent should be a fallback when find_package() fails, though it seems you (@jdumas) hold a dissimilar opinion.  I am curious to know why, though it does make sense to split the PR.

jdumas:
The reason to prefer FetchContent over find_package() is fairly simple, I want to [build everything from source, with the same flags, at the same time](https://youtu.be/IY8tHh2LSX4?t=1518). Building from source is the only way to ensure a reliable and repeatable build, and imho it's critical when sharing research code based on libigl, since those projects are typically published once and then never touched again.

In any case, wanton usage of `find_package()` without a proper package manager (such as vcpkg, hunter or conan) does more harm than good, which is why I do not want it to be the _default_.

Now that I look at @justend29's PR again I think it wasn't too bad actually. In particular I revised my initial statement about [these lines](https://github.com/justend29/libigl/blob/01a06e3bbf8b2ffa35c5f46da9c2eaad863dd313/cmake/recipes/external/eigen.cmake#L1-L3). I think they could work and those files are still somewhat self-contained. If we just separated the CMake style changes and put the `.in` files in a separate subfolder I think it'd be more viable.

KeithBallard:
@jdumas You have a very good argument, and I actually really appreciated the links.  To be honest, my change was half-baked to get my situation working, and I would be good to follow your philosophy.  However, I have a counter argument as well.  If you are tying dependencies to releases (not specific commits that may not be a release), then the [find_package() allows specifying a single version](https://cmake.org/cmake/help/latest/command/find_package.html#version-selection).  This would allow the best of both worlds in my view if you have a find_package call with the version you want with a FetchContent for that version upon failure.

As an aside, a package manager undoubtedly makes installing dependencies easier, but it is not a requirement for finding the version of an installed dependency.  I would argue that find_package(<package> <version>) works fine if one really wanted to manually install that package but is also super convenient if they use vcpkg.

KeithBallard:
PS: I promise I am not arguing on the internet for no reason.  You are very skilled in managing this project, and I think you might be receptive to some users who prefer to use installations due to multiple/many projects that require the same dependencies.  I am sincere in my view that I think there might be solution here.  Plus, I would be happy to contribute to it.

jdumas:
I am ok with allowing people to use `find_package()` to retrieve dependencies, but I just don't want it to be the default for reasons specified above.

Outside of ODR issues when you don't build from source, `find_package()` often only specifies a min version of a dependency, not an exact match, which can make reproducibility a challenge. E.g. imagine the following scenario:
1. You have some research project from 10y ago using libigl 1.x.
2. It uses a dependency `Foo` version 2.3.0.
3. The next major version of `Foo` 3.0 completely changes the way it's packaged/installed.
4. On your current Ubuntu version system libraries only provide `Foo` starting from version 6.0 or smth.
5. Basically you have no way of compiling said research code from 10y ago because you were using `find_package()` and your dependencies are all over the place.

That's the kind of scenario we're avoiding with `FetchContent`. Anyway if we want to add support for `find_package()`, PR #2131 was probably closer to what we need so we can try to resurrect that approach.

KeithBallard:
@justend29 You echoed many of my thoughts with far more eloquent language, and I appreciate your tact for making the case rather than attempt persuasion (as I naively do sometimes).  I would be happy to contribute, though I unfortunately won't have significant bandwidth until March 13th.

jdumas:
> Although, as you said, it's not as easy as just checking out and rebuilding.

Exactly. It's not easy, and if you have to rebuild the dependencies manually sometimes it's darn impossible due to evolving build instructions, etc. To properly use `find_package()` we'd need to 1) use version range to ensure things don't break with new major versions of each dependency, and 2) ensure each provided dependency comes with proper install rules & exported CMake targets. Some dependencies like STB are header-only and don't come with any CMake at all. Others like Schewchuk's predicate are also manually wrapped by us. GMP and MPFR are currently build without CMake entirely, etc.

At the end of the day we just don't have the luxury to properly support using `find_package()` for every dependency. And it is my opinion that building everything from source via FetchContent is a superior alternative anyway (see ODR issues mentioned above, out-of-the-box compilation experience, etc.).

It's fine if you disagree with it, and again I'm not trying to prevent anyone from contributing a `find_package()` alternative to libigl. I just want it to be done correctly, which means it's no little amount of work.

justend29:
CMake provides the concept of find-modules in addition to config-file packages. The former allows finding binaries from projects that don't offer their usage requirements through config-file packages.

I don't fully understand how `find_package()` inherently introduces ODR violations, since CMake uses the first target, essentially ignoring subsequent calls, thereby not linking against multiple different libraries. This behaviour extends to FetchContent. I know some people want everything to be built in one invocation, and that's still to be supported.

The out-of-the box compilation w/ FetchContent concept is great, but is a limited in its applicability. Existing projects that already use a package manager, system packages, or some other form of dependency management can experience libigl's dependencies as painful or incompatible.
As a different perspective, the original reason I made the find_package changes to libigl is because my usage of libigl is one where it's built on servers without any network connection. Additionally, I manage dependencies, both open-source and private, through a package manager. Also, I've worked at companies that demand all dependencies to be provided from internally hosted registries such that third-party hosting never influenced uptime. None of these are currently supported out-of-the box.

I realize there are an infinite number of use cases, and there's no way they could all be supported. Nevertheless, using package managers, and external dependencies is a large, common usage. Precluding doing these, or even building on isolated build-servers seems significant. 

As for the amount of work, I agree with you. However, everything requires work. Luckily, we have the power of open-source (free labour, lol), and I'm willing to do the work. `find_package()` support can be considered a second-class citizen, which is totally fine, and it can be implemented slowly. I think the idea of supporting both is a good idea, and only enhances libigl. We're all on the same team :smile:

justend29:
Ah, ODR violations from differing transitive binary dependencies. I thought you were saying different versions of the same direct dependency would be found and linked in as a consequence of `find_package()`, pardon me.

Yes, the possibility of every consumer providing libigl's dependencies _for it_, before using libigl, is an option, it just isn't out-of-the-box. That talk you link even recommends `find_package()`.

I may not prioritize this effort any longer, since this support clearly isn't valuable to the libigl authors, and that's totally ok. Sorry, Keith.

jdumas:
This talk is from 2017. FetchContent was introduced in CMake 3.11, which was released in March 2018 ;)

Look, just stop trying to convince me to that `find_package()` is preferable to `FethcContent`. It's ok to disagree on this. As I said I'm not opposed to a `find_package()` alternative for libigl, as long as it's not the _default_ behavior.

PR #2188, opened 2023-04-04
Fix CMake package installation for igl_core

### LINKED ISSUE #2186: v2.4.0 top-level CMake installation is broken
Observed when trying out the new port via vcpkg, see https://github.com/microsoft/vcpkg/issues/30621. Relevant details repeated below:

**CMake config files are broken**

1. Naming inconsistent 
image
2. Main config file not including target file, i.e., missing `include(${CMAKE_CURRENT_LIST_DIR}/LibiglConfigTargets.cmake)`.
   * It means the include folder of igl cannot be found on a simple test case, because it cannot find the target igl::igl_core

**Modules are not installed**

Upstream repository does not install the targets containing the options, only the core target, see https://github.com/search?q=repo%3Alibigl%2Flibigl%20igl_install&type=code. Moreover, installing does not include copyleft targets https://github.com/libigl/libigl/blob/a05865e2650f22057fcdc0b77c40d2458922f5d8/CMakeLists.txt#L150

- [x] Windows
- [x] macOS
- [x] Linux

FabienPean:
Fixes #2186 (or a part of it)

Fix the packaging of igl so that the core library is correctly found. Tested on WSL-Ubuntu-20.04 with GCC 9.4.0. 
```cmake
find_package(libigl CONFIG REQUIRED)
target_link_libraries(main PUBLIC igl::igl_core)
```

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/cmake/igl/libigl-config.cmake.in
+++ b/cmake/igl/libigl-config.cmake.in
@@ -1,3 +1,7 @@
 @PACKAGE_INIT@
 
+include(CMakeFindDependencyMacro)
+find_dependency(Eigen3 REQUIRED)
+find_dependency(Threads REQUIRED)
+include("${CMAKE_CURRENT_LIST_DIR}/LibiglConfigTargets.cmake")
 check_required_components(Libigl)


jdumas (approved):
Lgtm.

PR #2190, opened 2023-04-06
Support libigl in other projects that export targets via add_subdirectory

KeithBallard:
If libigl is included in another project that exports its own targets, an issue arises in that libigl does not export its targets (besides igl_core) nor are several third party libraries exported.  This PR is my attempt to export all components of libigl and relevant 3rd party libraries.  

This is part 1 of 2 for supporting the CMake installation interface.  If the targets are exported correctly, that resolves most of the work.  Part 2 would include ensuring all files are installed correctly.  This capability has been discussed across many issues.  Any input would be welcome.  I have developed the CMake code to the point that my own project needs at this time, but I would be happy to help get it the rest of the way.

To be clear: I am not advocating changing the integration method recommended by igl.  I have come around to the argument by @jdumas made in https://github.com/libigl/libigl/pull/2175.  However, I do think supporting an installation interface is necessary for some uses cases, like my own.  If library A integrates libigl and exports itself for use downstream, then libigl must correctly export its own targets for A to be consumed. 

I have tested it by setting up a project with libigl as a submodule (in external/libigl) with the following CMakeLists.txt:
```cmake
set(LIBIGL_BUILD_TESTS           OFF CACHE BOOL "" FORCE)
set(LIBIGL_BUILD_TUTORIALS       OFF CACHE BOOL "" FORCE)
set(LIBIGL_INSTALL               ON  CACHE BOOL "" FORCE)

set(LIBIGL_PREDICATES            ON  CACHE BOOL "" FORCE)
set(LIBIGL_EMBREE                ON  CACHE BOOL "" FORCE)
set(LIBIGL_GLFW                  ON  CACHE BOOL "" FORCE)
set(LIBIGL_IMGUI                 ON  CACHE BOOL "" FORCE)
set(LIBIGL_OPENGL                ON  CACHE BOOL "" FORCE)
set(LIBIGL_PNG                   ON  CACHE BOOL "" FORCE)
set(LIBIGL_XML                   ON  CACHE BOOL "" FORCE)
set(LIBIGL_COPYLEFT_CGAL         ON  CACHE BOOL "" FORCE)
set(LIBIGL_COPYLEFT_COMISO       ON  CACHE BOOL "" FORCE)
set(LIBIGL_COPYLEFT_TETGEN       ON  CACHE BOOL "" FORCE)
set(LIBIGL_RESTRICTED_MATLAB     ON  CACHE BOOL "" FORCE)
set(LIBIGL_RESTRICTED_MOSEK      ON  CACHE BOOL "" FORCE)
set(LIBIGL_RESTRICTED_TRIANGLE   ON  CACHE BOOL "" FORCE)

add_subdirectory(external/libigl)

add_library(test_igl_linkage main.cpp)

target_link_libraries(test_igl_linkage PRIVATE igl_core igl_predicates igl_png igl_xml)
target_link_libraries(test_igl_linkage PRIVATE igl_embree igl_glfw igl_opengl igl_imgui)
target_link_libraries(test_igl_linkage PRIVATE igl_copyleft_cgal igl_copyleft_comiso )
target_link_libraries(test_igl_linkage PRIVATE igl_copyleft_tetgen)
target_link_libraries(test_igl_linkage PRIVATE igl_restricted_triangle)
```

Important notes:
* I do not have Matlab or Mosek available to test those targets
* The targets are exported correctly for build interface, but I have not tested the install interface. It is quite likely that relevant 3rd party header files are not marked to be copied to the installation include directory.
* This PR currently points to my forks of some of projects mirrored by igl since I had to make changes in those projects as well (just the CMakeLists).  Before this PR is accepted, the 3rd party mirror PRs need to be accepted and the URLs updated for fetchcontent in this project.

#### Checklist

- [ ] Adds corresponding unit test.  (I don't know how to create an automated test for this)
- [x] [predicates mirror PR](https://github.com/libigl/libigl-predicates/pull/2) accepted
- [x] [glad mirror PR](https://github.com/libigl/libigl-glad/pull/1) accepted
- [x] [comiso mirror PR](https://github.com/libigl/CoMISo/pull/13) accepted
- [ ] FetchContent URLs updated
Are there other tasks for this?

### DIFF
--- a/cmake/igl/igl_install.cmake
+++ b/cmake/igl/igl_install.cmake
@@ -6,8 +6,10 @@ function(igl_install module_name)
     # Check if category is `copyleft` or `restricted`
     if(${module_name} MATCHES "^igl_copyleft")
         set(suffix "_copyleft")
+        set(LIBIGL_INSTALLED_COPYLEFT TRUE CACHE INTERNAL "")
     elseif(${module_name} MATCHES "^igl_restricted")
         set(suffix "_restricted")
+        set(LIBIGL_INSTALLED_RESTRICTED TRUE CACHE INTERNAL "")
     else()
         set(suffix "")
     endif()
@@ -61,3 +63,10 @@ function(igl_install module_name)
         )
     endforeach()
 endfunction()
+
+# These two variables keep track whether a restricted or copyleft target
+# was installed, so that the corresponding install/export call can be made
+# at the time of project installation.
+# See top level CMakeLists.txt in root folder of the libigl repo.
+set(LIBIGL_INSTALLED_RESTRICTED  FALSE CACHE INTERNAL "Tracks whether a restricted target was installed")
+set(LIBIGL_INSTALLED_COPYLEFT    FALSE CACHE INTERNAL "Tracks whether a copyleft target was installed")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,7 +138,6 @@ endif()
 ################################################################################
 # Install CMake config files
 ################################################################################
-
 if(LIBIGL_INSTALL)
     include(GNUInstallDirs)
     set(project_config_in "${PROJECT_SOURCE_DIR}/cmake/igl/libigl-config.cmake.in")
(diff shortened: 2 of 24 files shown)


cmake/igl/igl_install.cmake
```diff
@@ -61,3 +63,10 @@ function(igl_install module_name)
         )
     endforeach()
 endfunction()
+
+# These two variables keep track whether a restricted or copyleft target
+# was installed, so that the corresponding install/export call can be made
+# at the time of project installation.
+# See top level CMakeLists.txt in root folder of the libigl repo.
+set(LIBIGL_INSTALLED_RESTRICTED  FALSE CACHE INTERNAL "Tracks whether a restricted target was installed")
```
jdumas, line 71: I don't think these will work the way you intend. Maybe we should create an empty export set `LibiglTargets${suffix}` for each possible suffix, and simply unconditionally export that at the end of the root CMakeLists? That way if the set is empty, nothing will be exported anyway?

KeithBallard: In my tests, it worked as expected, but I could very well be blind to a side effect.  I tried something similar to what you suggest, but it seems CMake will throw if you try to call install/export with an export list that does not exist.  I didn't know how to create an empty export list, but that seems like a better solution if possible.  I will do some searching to figure that out.

jdumas: The problem I see with using a CACHE variable is that once `LIBIGL_INSTALLED_COPYLEFT` is set to `FALSE`, subsequent calls to `set(LIBIGL_INSTALLED_COPYLEFT ...)` will not change the value to `TRUE`. When you check your `CMakeCache.txt` after running CMake I suspect the value of this variable should still be `FALSE`.

KeithBallard: I did check my CMakeCache and it does show them as TRUE (when I have a copyleft or restricted option turned on respectively). It appears to be working as I expect with v3.25.1. However, I have not tested it with older versions.

KeithBallard: But...I agree that an empty export set would be cleaner.

alecjacobson: Is this PR still in progress?

Issue #2098, opened 2022-10-12
Why I don't have 'igl_glfw.lib'?


KMushaL:
#### Describe your issue

After make this program, I tried to debug(MSVS 2022) the tutorial 109, but it failed to find 'igl_glfw.lib'. Also, I couldn't find 'igl_glfw.lib' in the corresponding folder, there only exists 'igl_glfw.pdb'.
Here are the error and my folder:
![image](https://user-images.githubusercontent.com/47875865/195372623-d7fd148a-7185-41c5-a8e9-79d9f3905db9.png)
![image](https://user-images.githubusercontent.com/47875865/195374412-623a09ad-43f3-442b-a2d1-d60d06cdbc90.png)

But when I swith to RELEASE mode it works successfully, so the problem is why I don't have 'igl_glfw.lib'?

#### Checklist

- [x] I have read the [bug report](https://libigl.github.io/bug-report/) page
- [x] CMake issue: I have tried with a fresh clone/empty build directory

#### Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

zih-an:
I just met the same problem and solved it. In file `./include/igl/opengl/glfw/viewer.cpp`, delete the line 188: `printf("OpenGL Version %d.%d loaded\n", GLVersion.major, GLVersion.minor);`, then this file can be compiled correctly and igl_glfw.lib can be generated.

alecjacobson:
"Fixed" by 10002b6cf12fe3d8cb3b8a6ddcfd1362682ec924

PR #2198, opened 2023-05-08
Bug fix in GLVersion field Viewer.cpp

ning2510:
When I build libigl, I also encountered the problem of  #2098 and reported an error:
image

In `libigl/include/igl/opengl/glfw/viewer.cpp`, when I delete line 188  `printf("OpenGL Version %d.%d loaded\n", GLVersion.major, GLVersion.minor);` , I can compile normally

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/opengl/glfw/Viewer.cpp
+++ b/include/igl/opengl/glfw/Viewer.cpp
@@ -185,7 +185,7 @@ namespace glfw
       return(-1);
     }
     #if defined(DEBUG) || defined(_DEBUG)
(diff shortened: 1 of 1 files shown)


alecjacobson:
This one line continues to create all sorts of annoyances I think only on
windows. I _thought_ that glad was supposed to be avoiding this.
I don't really mind getting rid of this debug output but it seems odd to me
that there's not a reliable way to get this info.
-Alec

PR #2215, opened 2023-06-25
switch to gmp mirror

alecjacobson:
GMP .zip is [failling to download repeatedly during CI](https://github.com/libigl/libigl/actions/runs/5365140085/jobs/9741350204). When I try locally the link works fine, so I guess it's getting unlucky with the gmp server?

This PR switches to [an up-to-date mirror I found](https://github.com/alisw/GMP).


alecjacobson:
I'd much rather use gmp's official source but it's holding up CI. Hopefully we'll remember to revisit this from time to time to see if gmplib.org has become more reliable.

PR #2216, opened 2023-06-25
spectra module

alecjacobson:
This introduces a [spectra](https://github.com/yixuan/spectra) module. I'm still not 100% convinced that spectra is reliable. This uses a shift-invert mode with Eigen's SparseLU as a solver.

Matlab uses Arpack + umfpack. Eigen does provide a wrapper for umfpack, which would mean further including a SuiteSparse dependency (I'd say this is out of scope for now). So, in the meantime, I've templated over the solver so if a user really wanted to use umfpack they could.

One thing that makes me uncomfortable is that spectra did not find the correct modes for a basic unit test without really increasing the shift value `sigma`.  Ideally this would work with `sigma=0` or some other reliable procedure that is hidden from the caller. I'm not sure if we should blame spectra or SparseLU or the way I'm calling spectra for this issue.

As a use-case, I included Spectral Conformal Parametrization. Another great addition would be the skinning eigenmodes from https://arxiv.org/abs/2303.11886.


alecjacobson:
```-- Third-party: creating target 'spectra::spectra'
CMake Error at build/_deps/spectra-src/CMakeLists.txt:24 (find_package):
  Could not find a package configuration file provided by "Eigen3" with any
  of the following names:

    Eigen3Config.cmake
    eigen3-config.cmake

  Add the installation prefix of "Eigen3" to CMAKE_PREFIX_PATH or set
  "Eigen3_DIR" to a directory containing one of the above files.  If "Eigen3"
  provides a separate development package or SDK, be sure it has been
  installed.

```

Huh. I'm not seeing this problem locally. If anyone knows a quick fix I'd be grateful, I won't have time to return to this for a few days.

jdumas:
> Huh. I'm not seeing this problem locally. If anyone knows a quick fix I'd be grateful, I won't have time to return to this for a few days.

I guess that's because you have a system-wide version of Eigen that Spectra picks up. Spectra is looking for Eigen via [find_package()](https://github.com/yixuan/spectra/blob/8c7242e08f0fb7f6a0022cfe232e3dc5b5bd4eb4/CMakeLists.txt#L24) without checking that the target already exist first. I think you can set `Eigen3_FOUND` to true before including Spectra to trick it into thinking it has found Eigen (since it uses the target `Eigen3::Eigen` later on this will work).

jdumas:
Yes it would definitely solve the issues we're having. I don't understand why it's been opened for 2y...

jdumas:
If they don't merge such a change soon I can work on a libigl-side workaround later this week, I've done this a million times for misbehaving projects...

alecjacobson:
Thanks, that'd be great. Using my fork is not a great solution. Though it seems spectra is not changing very frequently.

PR #2236, opened 2023-08-17
Cmake option to delete tutorials as built

alecjacobson:
Attempt to get that windows header only tutorial build to fail less often on github actions.


alecjacobson:
Works locally on mac os but doesn't seem to actually delete as they build on windows. Hopefully #2237 and #2238 are fixing this issue anyway.

PR #2254, opened 2023-08-26
Fix a bunch of warnings

### LINKED ISSUE #1637: Compiler warnings
#### Describe your question
Hello, when I compile libigl with CMake, I get a lot of compiler warnings from libigl. It would be nice if you fix the code that throws the warnings.

You can replicate these warnings by cloning this repo [https://github.com/ivanpuhachov/ift6113_2020.git](url). Install the prerequisites for the repo (for a mesh processing course). Then go to the folders ending with `_cpp` and follow the inner readme.md to compile an executable. 

An excerpt:
```
C:\Users\Song\Documents\UDEM\ift6113\hw\ift6113_2020\libigl\include\igl\read_triangle_mesh.cpp(147): note: see reference to function template instantiation 'bool igl::readSTL<Eigen::Matrix<double,-1,-1,0,-1,-1>,Eigen::Matrix<int,-1,-1,0,-1,-1>,Eigen::Matrix<double,-1,-1,0,-1,-1>>(FILE *,Eigen::PlainObjectBase<Eigen::Matrix<double,-1,-1,0,-1,-1>> &,Eigen::PlainObjectBase<Eigen::Matrix<int,-1,-1,0,-1,-1>> &,Eigen::PlainObjectBase<Eigen::Matrix<double,-1,-1,0,-1,-1>> &)' being compiled
C:\Users\Song\Documents\UDEM\ift6113\hw\ift6113_2020\libigl\include\igl\readSTL.cpp(105): warning C4996: 'sscanf': This function or variable may be unsafe. Consider using sscanf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
C:\Users\Song\Documents\UDEM\ift6113\hw\ift6113_2020\libigl\include\igl\readSTL.cpp(133): warning C4996: 'sscanf': This function or variable may be unsafe. Consider using sscanf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
```

- [x] Windows
- [ ] macOS
- [ ] Linux

alecjacobson:
Fixes #1637 as much as I'm willing to.

This introduces the CMake flag `LIBIGL_WARNINGS_AS_ERRORS` which doesn't really turn on all warnings but the ones I was willing to fix:

```cmake
target_compile_options(${module_name} PRIVATE -Wall -Wextra -Wpedantic -Wno-sign-compare -Werror -Wno-gnu -Wno-unknown-pragmas)
```

I also wasn't willing to fix the warnings in the COMISO package and for some reason the compiler was finding warnings in the tetgen.h dependence even though that was supposed to be a system include and I thought they would be ignored. 

The strong majority of the warnings were unused variables. If there were unused _parameters_ I tried to remove those parameters (which will likely break some code; in a rather innocuous way).

Finally, there were some variables that show up just for assertions. It seems that C++ compilers don't have a good solution for this common problem. So I used the suggestion on [SO](https://stackoverflow.com/a/985807/148668).


include/igl/ASSERT.h
```diff
@@ -0,0 +1,9 @@
+// https://stackoverflow.com/a/985807/148668
+#include <cassert>
+#ifndef ASSERT
+#ifdef NDEBUG
+#define ASSERT(x) do { (void)sizeof(x);} while (0)
```
jdumas, line 5: I would suggest to prefix this with `IGL_`. You never know what other codebase might be defining their own `ASSERT` macro, in ways that could be incompatible with the libigl one...

alecjacobson:
Happy to merge a pr fix

PR #2255, opened 2023-08-28
Expose cutoff parameter for CGAL intersections; better default

### LINKED ISSUE #270: libigl's boolean take very large memory

alecjacobson:
Fixes #270

CGAL weirdly uses a pretty bad default for the somewhat mysterious `cutoff` parameter in box_intersections_d and box_self_intersections_d

I tired 1,10,100,1000,10000,100000 and √n for all the tutorial data models (not a particularly good test set) and 1000 was the clear winner.

So, this now uses that as the default and exposes the parameter for users to tweak if needed.

The speedups are modest (like 1.1-1.2x).

### DIFF
--- a/include/igl/copyleft/cgal/RemeshSelfIntersectionsParam.h
+++ b/include/igl/copyleft/cgal/RemeshSelfIntersectionsParam.h
@@ -29,16 +29,22 @@ namespace igl
         bool stitch_all;
         /// whether to use slow and more precise rounding (see assign_scalar)
         bool slow_and_more_precise_rounding;
+        /// cutoff parameter for box_self_intersection_d (based on some casual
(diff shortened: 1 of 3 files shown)


kbongort:
> the somewhat mysterious cutoff parameter

If you're curious, you can read about what this parameter is in [Fast Software for Box Intersections](https://dl.acm.org/doi/pdf/10.1145/336154.336192). Basically it determines how early the algorithm switches from tree recursion to scanning. The authors do a bunch of experimentation with the parameter themselves. It's odd that CGAL chose 10 as the default since the paper seems to point toward values in the 1k-10k range as best, generally.

PR #2256, opened 2023-08-29
Avoid conflicting template parameters in AABB

### LINKED ISSUE #436: igl::hausdorff does not work with MatrixX3d
You have to copy the vertices into a MatrixXd, otherwise Eigen spews YOU_MIXED_MATRICES_OF_DIFFERENT_SIZES errors.

alecjacobson:
Fixes #436

This uses class template specialization to avoid ever calling AABB with conflicting template params (e.g., MatrixX3D and DIM=2). The old code was all dynamic so it lead to conflicting compilation paths.

I claim this is a reasonable C++11 solution. If we were using C++17 then we could use `if constexpr` which would be significantly less code and more readable.

Unless I notice something broken with this, I'll also fix the nasty dynamic situation in `signed_distance` similarly. And then we can add [to AABB] a 
```cpp
static_assert(DerivedV::ColsAtCompileTime == DIM || DerivedV::ColsAtCompileTime == Eigen::Dynamic,"V's #columns must either match DIM or be Dynamic.");
```

So that bogus compilation paths get caught.

### DIFF
--- a/include/igl/point_mesh_squared_distance.cpp
+++ b/include/igl/point_mesh_squared_distance.cpp
@@ -9,6 +9,114 @@
 #include "AABB.h"
 #include <cassert>
 
+namespace 
+{
+  // Sigh. This is a somewhat elaborate way of making
+  // igl::point_mesh_squared_distance build a correctly dimensioned AABB tree
+  // regardless of whether it has #columns known at compile time or not.
+  //
+  // The pattern below uses a class which can be partially specialized on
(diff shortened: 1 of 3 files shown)


alecjacobson:
If I can do this right, then I can get rid of this junk: https://github.com/libigl/libigl/blob/7a84503c8e3f7f11576c23c19f0d020e1ec96fa2/include/igl/AABB.cpp#L994

PR #2295, opened 2023-09-28
tata -> data

alecjacobson:
Seems like a typo, but if @jdumas had a reason for calling these "tata" instead of "data" please let me know.


jdumas:
No reason lol. Maybe I was trying to see if reviewers were paying attention :D

PR #2299, opened 2023-10-06
fix aassertions, add templates to compile in debug

### LINKED ISSUE #2298: libigl version in CMakeLists.txt
#### Describe the bug
The version string within the root CMakeLists.txt still says it's 2.4.0.

This is the case for
* current main:
https://github.com/libigl/libigl/blob/7991ee81d81bd8a58df261f2b6cc782a77d57ec2/CMakeLists.txt#L40

* release tag v2.5.0:
https://github.com/libigl/libigl/blob/v2.5.0/CMakeLists.txt#L40

With each release I would expected the version to match that release version e.g. for the latest release:
```
project(libigl VERSION 2.5.0)
```

#### Platform

- [x] Windows
- [x] macOS
- [x] Linux

alecjacobson:
Release v2.5.0 (and stable) were not compiling in debug mode. Also bumps version to fix #2298


alecjacobson:
I guess I'll also merge this into `main` when it passes CI?

PR #2302, opened 2023-10-13
prefer std::u?int[0-9]+_t and include <cstdint> else include <stdint.h>

alecjacobson:
Fixes this kind of problem https://www.reddit.com/r/Fedora/comments/16issng/fedora_38_cant_run_simple_libigl_code/

`#include <cstdint>` or `#include <stdint.h>` is often missing (and accidentally included by some other path) or a global namespace type is used (`uint64_t`) with the assumption that `#include <cstdint>` will define it (it doesn't have to). So, prefer to use `std::` namespace types and `#include <cstdint>`. In some cases, where the code was ported from elsewhere I used `#include <stdint.h>` to match the global namespace types. (also remove some unnecessary `size_t` types along the way)


jdumas:
I don't understand why you need the `std::` namespace? Isn't it enough to include `<cstdint>`? I've never had any issue without the `std::` prefix for these types whatsoever.

alecjacobson:
It was news to me too. But apparently global namespace is only guaranteed with the c headers

jdumas:
Yikes... Well I'm choosing to close my eyes until someone complains and I cannot do otherwise but to change my code...

IslamOmar-360Imaging:
@alecjacobson  https://github.com/libigl/libigl/blob/main/include/igl/path_to_executable.cpp#L25 gives error as https://github.com/libigl/libigl/blob/main/include/igl/path_to_executable.cpp#L17 doesn't provide the namespace for `uint32_t` and  you are adding `using namespace std`, so

I think you should either add `<cstdint>` or use `uint32_t` without `std::`.

PR #2328, opened 2024-01-07
fix compile error

wtholliday:
Fixes a compile error when checking for not returning a value.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


wtholliday:
Note that CI failures are unrelated to this PR. The downloaded archive of boost is failing a checksum.

PR #2330, opened 2024-01-08
use our boost-cmake which uses slow, non-jfrog url

alecjacobson:
Boost's jfrog artifactory servers went bad somehow over the new year. It seems that if we ever bump boost to a modern version we should just use their github releases. I also don't really understand what role this extra `boost-cmake` repo is playing for us now. Is it needed? In any case, for now I forked it and changed the 1.71.0 url to something that doesn't crash.

### DIFF
--- a/cmake/recipes/external/boost.cmake
+++ b/cmake/recipes/external/boost.cmake
@@ -7,8 +7,8 @@ message(STATUS "Third-party: creating targets 'Boost::boost'...")
 include(FetchContent)
 FetchContent_Declare(
     boost-cmake
-    GIT_REPOSITORY https://github.com/Orphis/boost-cmake.git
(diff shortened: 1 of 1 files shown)


jdumas:
Boost doesn't officially support CMake integration, so this boost-cmake repo is a set of handwritten CMAke code to compile boost from source. There are CMake scripts in the official boost repo though, and although it's not officially supported we can it use with caveats. You can take a look at Lagrange's boost.cmake recipe for an example on how to compile boost from source using their official CMake code (but beware that the boost source repo is huge compared to the release zip). Since is the CMake code is stripped from the official boost releases there's unfortunately no other way to compile recent boost versions from source (and thus one might want to use CPM or Hunter to amortize the huge initial download).

PR #2338, opened 2024-01-16
enable mixed polygons using convention that negative indices are ignored

### LINKED ISSUE #2337: Change in API in tetrahedralize
#### Describe your issue

In this [commit](https://github.com/dyollb/libigl/commit/1b07ebcf9e18c33ba342fd5c945dedbda055df78) the API of tetrahedralize was changed probably to reduce maintenance.

The `std::vector` allowed to mesh polygons with different numbers of vertices (quads, triangles, ...) and segments.
This is how I previously created meshes with embedded skeleton bones/joints for running animations with BBW/LBS/DDM.

I would be happy to add something that works. Here are some ideas
- re-add std::vector api
- use a convention, e.g. process vertices of polygon until a `-1` is found (for quads, triangles and segments F would have shape Nx4)
- add extra argument(s), e.g. `Eigen::MatrixXi S` for segments

#### Checklist

- [x] I have read the [bug report](https://libigl.github.io/bug-report/) page
- [x] CMake issue: I have tried with a fresh clone/empty build directory

#### Platform

- [x] Windows
- [x] macOS Intel
- [x] macOS Arm (e.g., M1, M2)
- [x] Linux

dyollb:
Fixes #2337.

This change adds support for mixed polygons, e.g. triangles and segments.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/copyleft/tetgen/mesh_to_tetgenio.cpp
+++ b/include/igl/copyleft/tetgen/mesh_to_tetgenio.cpp
@@ -65,10 +65,10 @@ IGL_INLINE void igl::copyleft::tetgen::mesh_to_tetgenio(
     f->numberofholes = 0;
     f->holelist = NULL;
     tetgenio::polygon * p = &f->polygonlist[0];
-    p->numberofvertices = F.cols();
+    p->numberofvertices = (F.row(i).array() >= 0).count();
     p->vertexlist = new int[p->numberofvertices];
     // loop around face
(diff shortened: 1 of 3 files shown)


alecjacobson:
Do you mind adding a unit test? Maybe one for a polygon face and one for a segment?

This way if someone (me) gets overzealous about refactoring again it'd get cought.

PR #2344, opened 2024-01-22
Fix split_nonmanifold

alecjacobson:
The previous method (and the old one in gptoolbox) was cool cause it wouldn't just create boundaries at the non-manifold edge chains. It would identify big manifold components across non-manifold edges. Unfortunately, it was just getting lucky. It would find those if they were _in order_ in the list of faces, and fail to correctly split the mesh oetherwise. (Some of the new tests will fail the old code).

At one point I set out to implement "Cutting and Stitching: Converting Sets of Polygons to Manifold Surfaces" [Guéziec et al. 2001] but its details are a bit confusing and it's not quite written with proofs so I got nervous I'd make an elaborate implementation and still have failure cases.

Instead, I implemented the "obvious" thing. Split all the edges, then try to stitch them back. Only allow a stitch if the resulting mesh continues to be manifold. The order still matters to get nice components, so this implementation does a depth-first traversal on stitched edges as it goes.

This is purely combinatorial. It'd be cool to use geometry to prefer outputting a "solid" mesh. 

Finally, "manifold" doesn't necessarily mean orientable, but this code is assuming the output should also be orientable (I claim the common case). It shouldn't be too hard to add a flag that allows non-orientable output.

### DIFF
--- a/include/igl/split_nonmanifold.cpp
+++ b/include/igl/split_nonmanifold.cpp
@@ -6,10 +6,13 @@
 // v. 2.0. If a copy of the MPL was not distributed with this file, You can 
 // obtain one at http://mozilla.org/MPL/2.0/.
 #include "split_nonmanifold.h"
+#include "unique_edge_map.h"
+#include <cassert>
 #include "connected_components.h"
-#include "remove_unreferenced.h"
-#include "find.h"
-#include "ismember_rows.h"
+
+#include "is_vertex_manifold.h"
+#include "matlab_format.h"
+#include <iostream>
 
 template <
   typename DerivedF,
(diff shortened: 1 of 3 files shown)


alecjacobson:
Some thoughts on the geometric version. Ideally, _if_ a solid extraction exists, then it should be found. It's possible for this to still be ambiguous (the bottom steps of this model could have been two components or one):
split-nonmanifold-stairs

This extraction is essentially the same as the _nearly_ combinatorial `extract_cells`. Where just the cyclic order around the edge (pointing into the screen) matters. So in this case, we _must_ connect B→C and A→D (because B→C appears in cyclic order before B→D).
```
  B
  ↓
A→o→C
  ↓
  D
```

But in this case, we could either do B→C,D→A or B→A,D→C because both have no skips in cyclic order. We could further make a geometric rule to prefer the most convex out or something based on dihedral angle difference.
```
  B
  ↓
A←o→C
  ↑
  D
```

alecjacobson:
Another thought, in "Progressive Shell Quasistatics for Unstructured Meshes" we jiggle nearly colliding vertices into intersection free configurations. Seem unprincipled but I convinced myself that the problem is sufficiently hard that a randomized algorithm is a reasonable response. I think a similar approach would work well for finding a non-self-intersecting, manifold mesh that's close(est) to the input:

 0. Resolve any self-intersections, so they occur at (possibly non-manifold) edges and vertices
 1. split into manifold but overlapping at vertices and edges
 2. jiggle vertices by ε until self-intersection free
 
 Would also be a good use case for the [dynamic AABB that just merged](https://github.com/libigl/libigl/pull/2301).

PR #2360, opened 2024-03-04
fix debug bug: ndef instead of def

alecjacobson:
🙃


alecjacobson:
Merging this despite the failing unrelated test in main.

PR #2364, opened 2024-03-19
hybrid mass matrix for tets

alecjacobson:
Been wanting to code this up for... 13 years?

"Discrete Differential-Geometry Operators for Triangulated 2-Manifolds" [Meyer et al.] describes a method for computing lumped/diagonal mass-matrix entries by using Voronoi area for non-obtuse triangles (splitting each triangle into dual cell portions using the circumcenter) and using "safe" positive contributions computed by snapping the circumcenter to the nearest edge-midpoint. The formulation in the paper is cool cause it's intrinsic, but it's only defined for triangle meshes.

After some thought, I'm proposing this generalization to tetrahdral meshes (implemented for extrinsic case) and n-simplices (left for future/never).

The core idea is the same, compute Voronoi areas, but snap the circumcenter if it's outside the element.

For each edge (aka, line segment, aka 1-simplex) compute its center as its circumcenter which is its midpoint (always inside).

For each triangle (2-simplex), compute its center as its circumcenter but if it's outside, snap it to the center of the boundary edge opposite the corner corresponding to the most negative barycentric coordinate of the circumcenter.

For each tetrahedron, …

_General recursive rule_: for each n-simplex, compute its center as its circumcenter but if outside, snap to the center of the (n-1)-subsimplex opposite the corner corresponding to the most negative barycentric coordinate of the circumcenter.

The theorem we're relying on is that as a non-obtuse n-simplex morphs into an obtuse one, its circumcenter _exits_. At the moment the circumcenter is on the boundary, it _must_ be a circumcenter of some boundary (n-1)-simplex. 

In this way, Meyer et al. or this general construction if continuous in morphs from meshes with all non-obtuse elements to a messed up mesh.

I could probably make it intrinsic using [Cayley-Menger](https://mathworld.wolfram.com/Cayley-MengerDeterminant.html) matrix, for the circumradius.
https://math.stackexchange.com/a/4187480/35376


alecjacobson:
"Properties of Laplace Operators for Tetrahedral Meshes" [Alexa et al. 2020] proposes to use circumcentric mass matrices (alongside their dual construction of the Laplace operator for tet meshes). However, they make a different choice for what to do about negative volumes:

> Moreover, negative volumes can appear in the mass matrix (Figure 5  (center)) and negative coefficients wij can lead to a violation of the maximum principle and an indefinite Laplacian (see Section 7 ). This might not only produce visible artifacts in practice and prevent the use of certain linear solvers but can also hinder the applicability of theoretical results that require certain properties of the Laplacian or a positive definite mass matrix. If this is not desired, we propose to alter the construction for boundary elements. To this end, we replace the circumcenters of the tetrahedra and the faces with centroids if the circumcenters are not  located  in  the  interior  of  their  primal  objects  (see  Figure  5  ).This strategy is reminiscent of the mixed finite element/finite vol-ume strategy described by Meyer et al. [MDSB03] and guarantees a positive mass matrix and positive coefficients wij.

I don't agree that this is reminiscent of Meyer et al. Using centroids, means that the mass matrix of Alexa et al. is a discontinuous function of vertex positions.

PR #2382, opened 2024-04-27
Add and fix test for fast_find_self_intersections

### LINKED ISSUE #2381: fast_find_self_intersections false positives
```cpp
Eigen::MatrixXd V(15,3);
V<< 
  0.280039012432098,0.82766300439834595,0.46377500891685502,
  0.33890101313590998,0.78804200887680098,0.48812699317932101,
  0.31805300712585399,0.76836401224136397,0.49308800697326699,
  0.30947001278400399,0.80785250663757346,0.47595100104808802,
  0.299046009778976,0.79801350831985496,0.47843150794506101,
  0.32847701013088199,0.77820301055908248,0.490607500076294,
  0.29475451260805097,0.81775775551795971,0.46986300498247152,
  0.28954251110553697,0.81283825635910045,0.47110325843095802,
  0.32418551295995701,0.79794725775718722,0.48203899711370451,
  0.33368901163339598,0.78312250971794173,0.48936724662780751,
  0.30854950845241502,0.78318876028060946,0.485759757459164,
  0.32326500862836799,0.77328351140022322,0.4918477535247805,
  0.30425801128148999,0.80293300747871421,0.47719125449657451,
  0.31897351145744302,0.79302775859832797,0.48327925056219101,
  0.31376150995492902,0.78810825943946872,0.4845195040106775;
  Eigen::MatrixXi F(16,3);
F<< 
  2,11,10,
  1,8,9,
  0,7,6,
  3,12,13,
  5,14,11,
  3,13,8,
  4,12,7,
  4,14,12,
  4,10,14,
  5,9,13,
  3,6,12,
  5,13,14,
  14,10,11,
  13,9,8,
  12,6,7,
  14,13,12;
```

Reports `Found 5 self intersecting pairs`, but there are none (this is a twice subdivided triangle).

alecjacobson:
Fixes #2381 

We certainly can't expect non-exact tests to get everything right, but it should at least handle this case of two co-planar triangles not remotely close to each other. I've added an explicit test of coplanarity early, rather than determining it by exclusion.

cc @vfonov


alecjacobson:
sigh. Another false positive:

```cpp
Eigen::MatrixXd V(6,3);
V<< 
  0.39234799146652199,0.38443601131439198,0.44744500517845198,
  0.38924700021743752,0.385637506842613,0.45762349665164948,
  0.38700349628925301,0.38789276033639897,0.45634675025939975,
  0.39079749584197976,0.38503675907850249,0.45253425091505073,
  0.38769650459289529,0.38623825460672351,0.46271274238824822,
  0.39279749989509577,0.37299175560474374,0.45553924888372399;Eigen::MatrixXi F(2,3);
F<< 
  0,3,2,
  1,5,4;
```

vfonov:
Interesting observation:

this test fails (using numbers from your SO comment):
```
TEST_CASE("tri_tri_intersection_test_3d breaks Delliver's code", "[igl]")
{
  // test that breaks Delliver's method using doubles
  using RowVector3d = Eigen::Matrix<double, 1, 3>;

  RowVector3d  p1(0.38924700021743752,0.38563750684261300,0.45762349665164948); 
  RowVector3d  q1(0.39279749989509577,0.37299175560474374,0.45553924888372399); 
  RowVector3d  r1(0.38769650459289529,0.38623825460672351,0.46271274238824822); 
  
  RowVector3d  p2(0.39234799146652199,0.38443601131439198,0.44744500517845198); 
  RowVector3d  q2(0.39079749584197976,0.38503675907850249,0.45253425091505073); 
  RowVector3d  r2(0.38700349628925301,0.38789276033639897,0.45634675025939975);

  RowVector3d s,t;
  bool coplanar;
  REQUIRE( !igl::tri_tri_intersection_test_3d(p1,q1,r1,p2,q2,r2, coplanar, s, t) );

  REQUIRE( !coplanar );
}
```


However replacing `double` with `long double` makes it pass:
```
TEST_CASE("tri_tri_intersection_test_3d long double works", "[igl]")
{
  // test that doesn't break Delliver's method using long doubles
  using RowVector3ld = Eigen::Matrix<long double, 1, 3>;

  RowVector3ld  p1(0.38924700021743752,0.38563750684261300,0.45762349665164948); 
  RowVector3ld  q1(0.39279749989509577,0.37299175560474374,0.45553924888372399); 
  RowVector3ld  r1(0.38769650459289529,0.38623825460672351,0.46271274238824822); 
  
  RowVector3ld  p2(0.39234799146652199,0.38443601131439198,0.44744500517845198); 
  RowVector3ld  q2(0.39079749584197976,0.38503675907850249,0.45253425091505073); 
  RowVector3ld  r2(0.38700349628925301,0.38789276033639897,0.45634675025939975);

  RowVector3ld s,t;
  bool coplanar;
  REQUIRE( !igl::tri_tri_intersection_test_3d(p1,q1,r1,p2,q2,r2, coplanar, s, t) );

  REQUIRE( !coplanar );
}
```

alecjacobson:
Thanks, @vfonov. That's interesting. These missed cases are "easy" in the sense that they're not close calls or nearly overlapping. My sense is that there's a really basic flaw with the Delliver's floating point tests against `0.0`.

The smaller tests came from trying to check for self-intersections on an upsampled low-res model: so there are tons of coplanar but otherwise very nice and easy cases. I've added that as a test case because it seems like fixing them one-by-one was turning into whac-a-mole.

I'm tempted to try to switch the triple product tests with shewchuck's `orient3D` (which would mean moving these functions to `igl/predicates/`.

alecjacobson:
meanwhile, for the SO counterexample, I found that `tri_tri_intersection_test_3d` is wrong but `tri_tri_overlap_test_3d` is correct.

alecjacobson:
Calling `tri_tri_overlap_test_3d` before calling `tri_tri_intersection_test_3d` helps but does not get the correct answer on all easy cases unfortunately.

vfonov:
Another experiment:
```c++
TEST_CASE("tri_tri_intersection_test_3d x10 works", "[igl]")
{
  // test that breaks the Delliver's method using doubles
  using RowVector3d = Eigen::Matrix<double, 1, 3>;

  RowVector3d  p1(0.38924700021743752,0.38563750684261300,0.45762349665164948); 
  RowVector3d  q1(0.39279749989509577,0.37299175560474374,0.45553924888372399); 
  RowVector3d  r1(0.38769650459289529,0.38623825460672351,0.46271274238824822); 
  
  RowVector3d  p2(0.39234799146652199,0.38443601131439198,0.44744500517845198); 
  RowVector3d  q2(0.39079749584197976,0.38503675907850249,0.45253425091505073); 
  RowVector3d  r2(0.38700349628925301,0.38789276033639897,0.45634675025939975);

  RowVector3d  p1_=p1*10;
  RowVector3d  q1_=q1*10;
  RowVector3d  r1_=r1*10;

  RowVector3d  p2_=p2*10;
  RowVector3d  q2_=q2*10;
  RowVector3d  r2_=r2*10;

  RowVector3d s,t;
  bool coplanar;
  REQUIRE( !igl::tri_tri_intersection_test_3d(p1_,q1_,r1_,p2_,q2_,r2_, coplanar, s, t) );

  REQUIRE( !coplanar );
}
```

this passes. So, maybe we could just make a test that checks if some triangle parameters are too close to 0 for `double` precision?

alecjacobson:
I tried a really simple epsilon on the inequalities and it failed other
ones. I'm a little skeptical that the fix will be simple. But worth a short!

vfonov:
i am thinking about condition on the input parameters for which algorithm will work. I.e I think for a give precision (float/double/long double) it will always be possible to find combination of parameters that would make it fail.

alecjacobson:
Yeah I agree it can't be 100%. But I claim it'd failing on really "easy"
cases where the minimum distance between the triangles is relatively large.

vfonov:
I checked this two triangles on moller97's code ( https://github.com/erich666/jgt-code/blob/master/Volume_08/Number_1/Shen2003/tri_tri_test/include/Moller97.c ), it fails too when data is in `double` and in `long double`.

alecjacobson:
Thanks.

> I'm tempted to try to switch the triple product tests with shewchuck's orient3D (which would mean moving these functions to igl/predicates/.

Some rough timing estimates are that for _only_ detection, the `igl::copyleft::cgal::remesh_self_intersections` is 2-3× slower than `igl::fast_find_self_intersections` for 100K-face models. That leaves some room for spending flops on making `fast_find_self_intersections` work correctly. And of course, many users would rather not use cgal.

I'm not sure what to do in the meantime. Part of me thinks we should move the `igl::fast_find*` functions from main into a branch. I'm also worried I don't personally have time to fix them (for my personal project I think I'll switch to the cgal ones and suffer the 2-3x slow down).

alecjacobson:
The igl::predicates::find_* routines that I just pushed seem to work. They're detection only and duplicate most of the code in the igl::find_* versions. If they're still faster than cgal, then perhaps these are worth keeping (and trying to optimize).

alecjacobson:
Unsurprisingly the speed is closer to CGAL. If I take the self-intersection free decimated-knight and subdivide it 6 times so that it has 4M faces then I see:

| Method | Time | Note |
|-:|-:|:-|
| igl::fast_find_self_intersections | 4.1 secs | False positive |
| igl::predicates::find_self_intersections | 8.4 secs | Correct |
| igl::copyleft::cgal::remesh_self_intersections | 13.2 secs | Correct |

with all the methods set to `detect_only=true,first_only=true`.

So, still a bit faster than cgal, giving the right answer here and of course a more permissive license.

alecjacobson:
I separated the detection from resolution. Both are passing tests and work with tutorials. I think the resolution could be performance optimized a bunch. I also deleted the `igl::fast_find*` functions.

vfonov:
looks good, wouldn't it make sense to port the code  for finding intersecting lines too?

alecjacobson:
I didn't port the tri_tri_intersect code for that. I just wrote a simpler
thing for now. I'm not convinced that the tri_tri_intersect is doing
something particularly smart for that (it has to be inexact with floats
anyway). But I could be wrong!
One of the nice things of the new code is that it explicitly avoids
duplicate vertices: segment endpoints are shared into a graph.

alecjacobson:
I'm going to merge this for now. If @vfonov or others would like to further optimize things, I'm very welcome to it!

Issue #2383, opened 2024-04-29
Linker error on mac os continuous integration tests


alecjacobson:
```
Assertion failed: (extras.otherInstrOffset != 0 && "Kind::arm64_adrp_ldr missing extra info"), function applyFixup, file Fixup.cpp, line 793.
```

Seems to be this: https://forums.developer.apple.com/forums/thread/739993

Looks like we need a bugfix on xcode on the github action servers. I'm not sure how to do that.

jdumas:
The new linker that comes with Xcode 15.0 is seriously bugged. I suggest using Xcode 15.1 or 15.2 in GitHub Actions. Here is what I use [in Lagrange](https://github.com/adobe/lagrange/blob/930f894ccac1157ba318f6029aa3c7d4afc7fd1c/.github/workflows/continuous.yaml#L92-L96) to switch:

```yaml
      - name: Setup Xcode version
        if: runner.os == 'macOS'
        uses: maxim-lobanov/setup-xcode@v1
        with:
          xcode-version: '15.2'
```

PR #2384, opened 2024-04-29
Fix CI Build: avoid test on windows, remove Comiso module, specific xcode on github

### LINKED ISSUE #2363: ray_mesh_intersect: corner-case fails on windows CI update
One of these tests just started failing on Windows. I'm not sure what changed on the CI machines between:

❌ https://github.com/libigl/libigl/actions/runs/8147735595/job/22270837016
✅ https://github.com/libigl/libigl/actions/runs/7976689950/job/21777837779

The code for libigl is identical, so something in the compilation. I'm not a windows user so this is really difficult for me to debug. @eolivi-fy  or @jnordin20 would you be able to help me ?

_Originally posted by @alecjacobson in https://github.com/libigl/libigl/issues/2271#issuecomment-1977726806_

alecjacobson:
Fixes #2363


tests/include/igl/ray_mesh_intersect.cpp
```diff
@@ -30,6 +30,8 @@ TEST_CASE("ray_mesh_intersect: one_triangle", "[igl]")
   IGL_POP_FPE;
 }
 
+// https://github.com/libigl/libigl/issues/2363
+#ifndef _WIN32
```
jdumas, line 34: Rather than not compile the test on Windows, I would mark it as `[!mayfail]` so it is compiled but not run (and can easily be run manually). Something like this:

```c++
#ifdef _WIN32
#define IGL_MAYFAIL_WIN "[!mayfail]"
#else
#define IGL_MAYFAIL_WIN
#endif

TEST_CASE("ray_mesh_intersect: corner-case", "[igl]" IGL_MAYFAIL_WIN)
{
...
}
```

alecjacobson: oh nice! definitely let's do that.

alecjacobson:
Fixes #2383 too (hopefully)

alecjacobson:
I'm going to take this moment to remove the comiso module. It was put on notice with the last release. I will not be maintaining it. If someone else would like to, then happy to accept a PR adding it back.

PR #2419, opened 2024-10-14
Fix "Assertion failed" when calling igl::heat_geodesics_precompute using a V with fixed column count

### LINKED ISSUE #2418: "Assertion failed" when calling `igl::heat_geodesics_precompute` using a `V` with fixed column count.
#### Describtion

The following code works fine:

```cpp
#include <Eigen/Core>
#include <igl/heat_geodesics.h>

{
  // Inline mesh of a cube
  const Eigen::MatrixXd V = (Eigen::MatrixXd(8,3)<<
    0.0,0.0,0.0,
    0.0,0.0,1.0,
    0.0,1.0,0.0,
    0.0,1.0,1.0,
    1.0,0.0,0.0,
    1.0,0.0,1.0,
    1.0,1.0,0.0,
    1.0,1.0,1.0).finished();
  const Eigen::MatrixXi F = (Eigen::MatrixXi(12,3)<<
    0,6,4,
    0,2,6,
    0,3,2,
    0,1,3,
    2,7,6,
    2,3,7,
    4,6,7,
    4,7,5,
    0,4,5,
    0,5,1,
    1,5,7,
    1,7,3).finished();

  igl::HeatGeodesicsData<double> data;
  igl::heat_geodesics_precompute(V, F, data);
}
```

but changing the line

```cpp
const Eigen::MatrixXd V = (Eigen::MatrixXd(8,3)<<
```

to

```cpp
const Eigen::MatrixX3d V = (Eigen::MatrixX3d(8,3)<<
```

makes `igl::heat_geodesics_precompute` fail with the following error message:

```
Assertion failed: (!(RowsAtCompileTime!=Dynamic) || (rows==RowsAtCompileTime)) && (!(ColsAtCompileTime!=Dynamic) || (cols==ColsAtCompileTime)) && (!(RowsAtCompileTime==Dynamic && MaxRowsAtCompileTime!=Dynamic) || (rows<=MaxRowsAtCompileTime)) && (!(ColsAtCompileTime==Dynamic && MaxColsAtCompileTime!=Dynamic) || (cols<=MaxColsAtCompileTime)) && rows>=0 && cols>=0 && "Invalid sizes when resizing a matrix or array.", file {{EIGEN_DIR}}\Eigen\src/Core/PlainObjectBase.h, line 277
```

#### Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

#### Investigation

The error seems to be in this line:
https://github.com/libigl/libigl/blob/fac5d4a01da9ddc83b96036c4bd544c550092a15/include/igl/heat_geodesics.cpp#L85

When calling `igl::heat_geodesics_precompute` with a `V` of dimension Nx3,  `M.diagonal().transpose()` will be of dimension NxN, but `M_diag_tr` will have `ColumnsAtCompileTime` set to 3.

#### Open questions

A possible solution would be to change the type of `M_diag_tr` to `Eigen::Matrix<DerivedV::Scalar, DerivedV::RowsAtCompileTime, DerivedV::RowsAtCompileTime>` or `Eigen::Matrix<DerivedV::Scalar, Eigen::Dynamic, Eigen::Dynamic>`, but its probably better to avoid the allocation of a dense NxN matrix here altogether?

How about changing
https://github.com/libigl/libigl/blob/fac5d4a01da9ddc83b96036c4bd544c550092a15/include/igl/heat_geodesics.cpp#L85-L86

to 

```cpp
const Eigen::SparseMatrix<Scalar> Aeq = M.diagonal().transpose().sparseView();
```

or

```cpp
const Eigen::SparseMatrix<Scalar> Aeq = M.diagonal().sparseView();
```
?

The call to `.diagonal()` and thus the creation of `Aeq` should be redundant too, but maybe its better to keep it for clarity and in case someone uses a different type of mass matrix in the future?

Nevsor:
Fixes #2418.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
This looks good to me. (I think long ago Eigen was really unreliable when chaining sparse function calls.)

alecjacobson:
Unfortunately this is causing a runtime error on windows. It could actually be due to the change. Not sure why. I don't have cycles to investigate right now.

Issue #2458, opened 2025-04-07
Embree causes trouble during compilation


odedstein:
#### Describe your issue

I have trouble building the libigl example project with embree on my Mac after the newest update for my toolchain. I have Apple clang version 17.0.0 (clang-1700.0.13.3), Target: arm64-apple-darwin24.3.0, and I am using cmake version 3.26.3.

Compiling the libigl-example-project with embree turned on gives me the following compilation errors:

```
In file included from /Users/ostein/Desktop/libigl-example-project/build/_deps/embree-src/kernels/common/device.cpp:9:
In file included from /Users/ostein/Desktop/libigl-example-project/build/_deps/embree-src/kernels/common/scene_curves.h:10:
/Users/ostein/Desktop/libigl-example-project/build/_deps/embree-src/kernels/common/../subdiv/bezier_curve.h:138:56: error: no member named 'u' in 'QuadraticBezierCurve<V>'
  138 |         return cout << "QuadraticBezierCurve ( (" << a.u.lower << ", " << a.u.upper << "), " << a.v0 << ", " << a.v1 << ", " << a.v2 << ")";
      |                                                      ~ ^
/Users/ostein/Desktop/libigl-example-project/build/_deps/embree-src/kernels/common/../subdiv/bezier_curve.h:138:77: error: no member named 'u' in 'QuadraticBezierCurve<V>'
  138 |         return cout << "QuadraticBezierCurve ( (" << a.u.lower << ", " << a.u.upper << "), " << a.v0 << ", " << a.v1 << ", " << a.v2 << ")";
      |                                                                           ~ ^
In file included from /Users/ostein/Desktop/libigl-example-project/build/_deps/embree-src/kernels/common/accelset.cpp:5:
In file included from /Users/ostein/Desktop/libigl-example-project/build/_deps/embree-src/kernels/common/scene.h:14:
In file included from /Users/ostein/Desktop/libigl-example-project/build/_deps/embree-src/kernels/common/scene_curves.h:10:
/Users/ostein/Desktop/libigl-example-project/build/_deps/embree-src/kernels/common/../subdiv/bezier_curve.h:138:56: error: no member named 'u' in 'QuadraticBezierCurve<V>'
  138 |         return cout << "QuadraticBezierCurve ( (" << a.u.lower << ", " << a.u.upper << "), " << a.v0 << ", " << a.v1 << ", " << a.v2 << ")";
      |                                                      ~ ^
[...]

PR #2459, opened 2025-04-07
embree v4 ; add missing template

alecjacobson:
It's likely that the CI build might break on unrelated embree issue #2458


alecjacobson:
Fixes #2458

PR #2466, opened 2025-04-25
Use box-drawing characters in comments.

jdumas:
To avoid compiler warnings of the style:

```
include/igl/AABB.cpp:549:3: warning: multi-line comment [-Wcomment]
549 |   //            /  \
```

I've replaced comments using forward/backward slashes with unicode box-drawing characters. This avoids ending a comment with a trailing backslash, which can trigger some annoying compiler warnings.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.

### DIFF
--- a/include/igl/AABB.cpp
+++ b/include/igl/AABB.cpp
@@ -59,11 +59,11 @@ namespace
       using Scalar = typename DerivedV::Scalar;
       const Scalar epsilon = igl::EPS<Scalar>();
       static_assert(
-        DerivedV::ColsAtCompileTime == 2 || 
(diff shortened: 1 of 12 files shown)


alecjacobson:
Thanks. Should probably compile, but the hard-coded ubuntu-20.4 is retired. Do you remember why the version is hard-coded rather than ubuntu-latest?

jdumas:
> Do you remember why the version is hard-coded rather than ubuntu-latest

Hmm no particular reason. I think I may have wanted to test compatibility on the "oldest" available ubuntu image.

Issue #2480, opened 2025-06-20
ambiguous symbol for `Matrix` in several files


xarthurx:
#### Describe your issue

Some of the header files, including but not limited to:
- massmatrix_intrinsic.cpp
- gaussian_curvature.cpp

are using `Eigen::Matrix` type without the namespace `Eigen`.
This in some cases create compilation issue with other libraries as `Matrix` is a quite common type.

e.g.:
https://github.com/libigl/libigl/blob/73a2e0de9b0bcfac83dc5bb661049c516a73650f/include/igl/massmatrix_intrinsic.cpp#L49
https://github.com/libigl/libigl/blob/73a2e0de9b0bcfac83dc5bb661049c516a73650f/include/igl/gaussian_curvature.cpp#L21

**Suggestion**
Parse the whole lib and add `Eigen::` to all such cases.

#### Checklist

- [x] I have read the [bug report](https://libigl.github.io/bug-report/) page
- [x] CMake issue: I have tried with a fresh clone/empty build directory

#### Platform

- [x] Windows
- [ ] macOS Intel
- [ ] macOS Arm (e.g., M1, M2)
- [ ] Linux

Issue #2132, opened 2022-12-26
Avoid using namespace std and use std::vector instead


Nosenzor:
#### Describe your issue

In some files (eg faces_first.cpp; triangle_triangle_adjacency.cpp; unique_edge_map.cpp...) the usage of std::vector are not prefixed by "std::" leads to a compile error for ambiguous naming (vector is too common in geometry)
The problem appears when you only include a very small subset of libIGL (for now i'm including only "igl/triangle_triangle_adjacency.h")
Also i've noted "#include <vector>" is missing in some files.

using namespace std is a bad practice ...
I suggest to also remove all usage of "using namespace Eigen"

**I can do the change and do a PR if you like ...**

#### Checklist

- [ X] I have read the [bug report](https://libigl.github.io/bug-report/) page
- [ X] CMake issue: I have tried with a fresh clone/empty build directory

#### Platform

- [X ] Windows
- [ X] macOS
- [ X] Linux

BruegelN:
Related to https://github.com/libigl/libigl/issues/1282, especially https://github.com/libigl/libigl/issues/1282#issuecomment-527158060
____

> vector is too common in geometry

Yes, thus it's a good idea to capsule your `vector` class into your own namespace and not pollute global namespace with it.

Nosenzor:
Problem is that IGL is "pseudo header-only" (When you include the header it will include the cpp), so it breaks the common rules of not using "using namespace" in headers and pollute the namespace for every caller.
In library mode that's not a big issue (although not recommended).
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-using
IMO adding "std::" in front of vector declaration is not too cumbersome.
Do you see any issue adding "std::" ?

alecjacobson:
There shouldn't be any `using namespace` calls in libigl outside of function scopes, so I don't see how this would pollute the namespace.

jdumas:
I've had issues in the past with _other_ headers introducing their own `vector` class, leading to compilation when using libigl in header-only mode. Basically the problem was the other third-party header is included _before_ a libigl header (sometimes for a reason outside of my control), and then the libigl-included .cpp header complains about some ambiguous unqualified `vector` types (one coming from the `std::` namespace, one coming from said third-party).

Unfortunately I don't have a concrete MWE at the moment. One could argue that it's the polluting third-party lib's fault as well, so it would be better to have a concrete failure case with a problematic third-party library that we could examine...

[...]

Issue #1282, opened 2019-08-30
Compilation Errors


aoabuhatab:
#### Describe your issue

I am getting compilation errors because of namespace clashes. Problem is solved when namespace is removed and direct scope is used. E.g. instead of (using namespace Eigen) use (Eigen::) where ever needed.

Because other namespaces that we have, has a matrix class, for example, defined as well.

e.g. per_vertex_normals.h

#### Check all that apply (change to `[x]`)
- [ ] Windows
- [ ] Mac OS X
- [x] Linux
- [ ] I have tried the `dev` branch and the problem persists

See https://libigl.github.io/CONTRIBUTING/#bugreport for more tips.

jdumas:
Afaik all occurrences of `using namespace Eigen` are scoped within libigl functions. So if their is a compilation clash, it's probably your fault for polluting the global namespace. If you can provide a MWE where such a clash happens without yourself polluting the global namespace, we'd be happy to look into it.

adityaghantasala:
Completely agree, stray classes are very bad and pollute the global namespace. Saying that assuming everyone is ware of that and does program accordingly is a bit dangerous. Here is a small example code 

```cpp
#include <iostream>

namespace nsp2
{
	class Matrix{
		public:
			Matrix(int ii = 0){i = ii;}
		private:
			int i;
	};
}

class Matrix{
	public:
		Matrix(double ii = 0.0){i = ii;}
	private:
		double i;
};

namespace nsp3
{
	void UseMatrix(){
		using namespace nsp2;
		Matrix A(1);
	}
}


int main() {
	std::cout << "!!!Hello World!!!" << std::endl; // prints !!!Hello World!!!
	return 0;
}
```

In namespace nsp3  `using namespace nsp2` is scoped in the function `UseMatrix`. But the presence of class `Matrix` which does not belong to any namespace will compilation failure with the error saying `error: reference to ‘Matrix’ is ambiguous`.  What works without any ambiguity and even if someone pollutes the global namespace is 

```cpp
namespace nsp3
{
	void UseMatrix(){
		nsp2::Matrix A(1);
	}
}
```
A stray class `Matrix` can occur when combining multiple projects, sometimes they can be multiple open-source projects (which happened in our case). So to avoid clashes like this, its safe to use `nsp2:: Matrix`.

jdumas:
[...]

PR #2483, opened 2025-06-23
No longer using namespace std; and using namespace Eigen; to avoid issues with other libs

alecjacobson:
Previously from #2481

> While `using namespace std;` and `using namespace Eigen;`  was limited to function scope in libigl there have been a few issues where third-parties did not play nice with it.
> I'm on par with @jdumas that's it's primarily the third-parties fault, however these issues keep popping up e.g.
> - https://github.com/libigl/libigl/issues/2480
> - https://github.com/libigl/libigl/issues/2132
> - https://github.com/libigl/libigl/issues/1282
> 
> Herby I propose to avoid `using namespace std;` and `using namespace Eigen;` completely in the sources of libigl.
> For quality of live I did not touch tests or tutorials.

### DIFF
--- a/include/igl/AABB.cpp
+++ b/include/igl/AABB.cpp
@@ -817,7 +817,6 @@ IGL_INLINE void igl::AABB<DerivedV,DIM>::init(
     const Eigen::MatrixBase<Derivedelements> & elements,
     const int i)
 {
-  using namespace std;
   using namespace Eigen;
   clear();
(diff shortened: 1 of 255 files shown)


alecjacobson:
I made a new PR because I couldn't quickly figure out how to push to the other one. Just a few more missing std:: and Eigen::

BruegelN:
Upsi and thank you @alecjacobson !

PR #2488, opened 2025-07-16
Lipschitz octree pruning

alecjacobson:
https://github.com/user-attachments/assets/462a9a1d-35c4-441c-8924-166a00e72211

Given a 1-lipschitz non-negative function (like an unsigned distance function), we can conservatively prune an octree leaving only leaf nodes that might contain the zero-levelset of the function.

The new tutorial shows how this plugs into the sparse marching cubes implementation for O(n²) extraction of level sets. Unlike the existing `igl::sparse_voxel_grid` this does not require a "seed" cell and will find all components of the extractable mesh at a given resolution. 

I've done some light optimization to avoid to many unnecessarily repeated calls to the query function (though each corner is still called 2× on average).

On my laptop, I can go up to a 4096³ grid and extract everything in less than a minute. A dense grid that big would not fit in memory.


alecjacobson:
btw, [sphere carving](https://aparis69.github.io/SphereCarving/index.html) is an alternative to this simple octree method. That paper compares to an octree approach and it's a bit inconclusive which method is faster. The octree has the advantage of very simple logic and immediate compatibility with sparse marching cubes.

PR #2497, opened 2025-09-05
Reduce mallocs in principal_curvature.

svenpilz:
Hi!

I experienced performance issues with `principal_curvature` on platforms where allocations are expensive.

For me, reducing the number of allocs in `principal_curvature` helped to improve the performance when used on Emscripten with the default allocator (atomic operations are expensive in WebAssembly).

Might also help to reduce heap fragmentation.

This only changes the internals, API stays the same. The  principal curvature test passes.

Maybe this is helpful for you guys.

Greetings from Berlin
Sven

#### Checklist

- [ x ] All changes meet [libigl style-guidelines](https://libigl.github.io/style-guidelines/).
- [ ] Adds new .cpp file.
- [ x ] This is a minor change.


include/igl/principal_curvature.cpp
```diff
@@ -900,25 +893,27 @@ IGL_INLINE void igl::principal_curvature(
   // Copy it back
   for (unsigned i=0; i<V.rows(); ++i)
   {
-    PD1.row(i) << cc.curvDir[i][0][0], cc.curvDir[i][0][1], cc.curvDir[i][0][2];
-    PD2.row(i) << cc.curvDir[i][1][0], cc.curvDir[i][1][1], cc.curvDir[i][1][2];
-    PD1.row(i).normalize();
-    PD2.row(i).normalize();
+    if (cc.curv[i].has_value()) {
```
svenpilz, line 896: This check is new, corresponding to https://github.com/libigl/libigl/pull/2497/files#diff-2419bde26b9d0d869e13e85423962dbf2a8a5349ff94facb26dc150422f5bcebL820.

PR #2511, opened 2025-10-31
snap_rounding: fix x86 MSVC build (construct EScalar from int, not size_t); add test

dg0yt:
### What actually broke

In `snap_rounding`, the centroid of a segment's intersections with a hot pixel was computed as:

```cpp
cen = Vector_2(cen.x()/EScalar(hits.size()), cen.y()/EScalar(hits.size()));
```

`EScalar` is `CGAL::Epeck::FT`, and `hits.size()` is a `std::size_t`. Constructing the exact number type from an unsigned `size_t` is ambiguous / ill-formed under **x86 MSVC** (where `size_t` is a 32-bit `unsigned int`): both `unsigned → int` and `unsigned → double` are equal-rank standard conversions. Everywhere else in this file `EScalar` is built from a plain `int` (e.g. `EScalar(i)` in `round`), which is why only this spot broke.

The `std::vector<Point_2>` itself was never the problem — it was just the source of the `size_t`. Removing it is a nice cleanup (no allocation for ≤8 points), but the load-bearing change is counting the hits with an **`int`** so the division is `EScalar(int)`, which is unambiguous.

### Changes

- Accumulate the centroid on the fly with an `int` counter (original @dg0yt change), plus a comment explaining *why* the `int` matters so it isn't "optimized" back to `size_t`.
- Rename the inner `Segment_2 * s` → `seg` (it shadowed the segment source point `s`).
- **Add the first unit test for `snap_rounding`**: two segments crossing exactly at `(5,5)` must insert that Steiner vertex and split both segments through it — exercising exactly the changed "centroid of hits" path. Verified locally (Epeck via `CGAL_DISABLE_GMP`): output is byte-identical to the pre-change code.

No behavior change; this only fixes the build and adds coverage.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
Reviewed and verified. LGTM to merge (pending green Windows CI).

## Correctness of the int-vs-size_t change

The load-bearing change is counting intersection hits with an `int` so the final centroid division is `EScalar(int)` rather than `EScalar(size_t)`.

- **Value range / no truncation:** `hits_count` is incremented at most twice per wall (a `Point_2` intersection adds 1; a collinear-overlap `Segment_2` adds its two endpoints) across 4 walls, so it is bounded by 8. It is a hit counter, not a geometric coordinate, and cannot exceed INT_MAX. `EScalar(int)` and the old `EScalar(hits.size())` therefore produce the identical exact value for every reachable input. The refactor from `std::vector<Point_2> hits` to on-the-fly summation is a pure equivalent rewrite (same points summed, same count, same divisor), so no behavior change.
- **Why only this spot broke:** `EScalar` is `CGAL::Epeck::FT`, whose relevant constructors are from `int` and `double`. On x86 MSVC `size_t` is a 32-bit `unsigned int`, so `unsigned -> int` and `unsigned -> double` are equal-rank standard conversions and the call is ambiguous/ill-formed. Every other `EScalar(...)` construction in this file already uses a plain `int` (e.g. `EScalar(i)` in `round`), consistent with this being the only affected line. On LP64 (this box) `size_t` is `unsigned long`, which does not trigger the ambiguity — matching the fact that the failure is x86-MSVC-only.
- The `Segment_2 * s` -> `seg` rename removes shadowing of the segment source point `s`; harmless and a genuine readability improvement.

## Local build/test result

CGAL 6.0.1 headers are available here but without GMP/MPFR, so I compiled the new test **header-only** with `-DCGAL_DISABLE_GMP=1` (Epeck) on gcc/Linux, exactly as the author did:

- Test TU compiles cleanly, links, and **runs green: "All tests passed (21 assertions in 1 test case)".**

## Added test assessment

The test is meaningful and exercises the changed path: two segments crossing exactly at `(5,5)` force a hot-pixel intersection so the "centroid of hits" code runs. It checks the right invariants — 5 output vertices with the rounded input copies first, the lone Steiner vertex at `(5,5)`, 4 output edges each incident on vertex 4 and non-degenerate, and `J` mapping to two children per parent segment. Good first coverage for `snap_rounding`.

## Caveat

The x86-MSVC compile failure itself is not reproducible on this Linux/gcc box (LP64 `size_t`), so the MSVC-specific fix is validated by reasoning about the overload set + value range above, plus the clean local header-only build. The Windows CI job should be run/confirmed green to close the loop; given the reasoning that is a formality.

---
🤖 This review was written by [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8) on behalf of the maintainer.

PR #2513, opened 2025-11-13
Thread pool parallel for

### LINKED ISSUE #2412: fast_winding_number creates too many threads
AFAICT, on a big enough mesh, `fast_winding_number` will generate enough threads that the OS (actually iPadOS in this case, but probably applies to all others) will eventually not allow more threads. 

image

image

#### Platform

- [x] Windows
- [x] macOS
- [x] Linux

alecjacobson:
Fixes #2412 

Vibe coded with chatgpt including the tests.


alecjacobson:
## Hardened the pool backend + made backends pluggable (pushed to this branch)

Building on the original thread-pool version, I reworked `parallel_for.h` so the shared pool is safe as the long-term default, added opt-in TBB/OpenMP backends with **no dependency leak into the default build**, and exposed the choice through one CMake option. Summary of what changed and why:

### Default backend (internal `std::thread` pool)
- **Fixed a real data race (intermittent SIGSEGV).** The completion counter was a lone atomic, so the caller could wake — even spuriously — see `remaining==0`, and destroy the stack-local mutex/condvar while the last worker was still `lock/notify`-ing them. Now the decrement/test/notify happen under one mutex that the waiter's predicate also uses. (Reproduced the crash before, 8/8 clean after.)
- **Leak-on-exit**: the pool is heap-allocated and never destroyed, so process/library teardown never blocks joining idle/in-flight workers. Joining threads from a static destructor is a classic shutdown hang (DLL/plugin unload) and breaks thread-restricted platforms (WASM, iPadOS — the platform in #2412). The OS reaps the blocked workers at exit.
- **Lazy creation**: workers are spawned on the first *parallel* region, so merely linking libigl costs no background threads.
- **Caller participates**: the calling thread runs one chunk instead of parking a core.
- **Uniform nested→serial guard** via a *scoped* thread-local flag (restored on exit, so a thread is reusable for later non-nested regions). This is the same guard as #2514 and is what actually caps the thread count.

### Pluggable backends (exactly one; headers included ONLY when selected)
```
LIBIGL_PARALLEL_FOR_BACKEND = POOL   (default, dependency-free)
                            | OPENMP (experimental)
                            | TBB    (experimental; version-robust, no this_task_arena)
                            | SERIAL
```
Wired in `core.cmake` via `${IGL_SCOPE}` (PUBLIC for static / INTERFACE for header-only) so the define — plus `-fopenmp` / TBB link — propagate to every consumer of the header-only `parallel_for.h`. A requested-but-missing OPENMP/TBB **warns and falls back to POOL**, so the build never breaks on a missing optional dep. The internal pool remains the default and the fallback.

### Bugs surfaced while testing the backends
- `IGL_PARALLEL_FOR_FORCE_SERIAL` didn't compile (the pool `#else` branch, referencing `parallel_for_pool`, was still compiled). Restructured so forced-serial compiles *only* the serial path.
- The suite wasn't backend-honest: `used_parallel == true` assertions fail under serial. Gated them on a helper so the full suite passes under every backend.

### Tests
Added #2412 regressions: distinct thread count stays bounded under deep nesting, and pool reuse across many calls doesn't grow the thread count. Documented how to run the suite under each backend.

### Validation (128-core box)
Built and ran the whole `parallel_for` suite under all four backends:
- **POOL**: passes at 128 threads (pool created once).
- **TBB**: passes at 128 threads (work-stealing).
- **OpenMP**: passes with `IGL_NUM_THREADS` capped — notably raw libgomp *hangs* on repeated 128-thread regions here, which the pool/TBB backends don't (empirical support for the pool as the default).
- **SERIAL**: passes.

CMake propagation confirmed by reconfiguring the real build: OPENMP → `-fopenmp`+define, SERIAL → force macro, TBB → links system oneTBB, POOL → nothing.

### Relationship to #2514
The nested→serial guard from #2514 is the *correctness* fix and is included here; the pool is the *performance* fix (it also removes the ~ms per-call thread-spawn overhead that shows up across libigl's ~59 `parallel_for` call sites). So this PR is effectively #2514's guard + a hardened engine, rather than a competing approach.

PR #2531, opened 2026-02-17
Added ability to change rotation up axis when ViewerCore::rotation_type == ROTATION_TYPE_TWO_AXIS_VALUATOR_FIXED_UP

villains:
This change allows you to set the direction of the up axis when `ViewerCore::rotation_type` is set to `ROTATION_TYPE_TWO_AXIS_VALUATOR_FIXED_UP`. By default, the y-axis (0,1,0) is used as the up axis. However there are cases where it is more natural to work with the z-axis as the up axis. For example, when working with geoscientific data, the general convention is that x and y are the geographic coordinates, and z is the elevation. When viewing 3D datasets of this type, the most natural tendency is to rotate about the z-axis.

This PR implements the ability to choose the up axis when required.

Together with `ViewerCore::camera_eye` and `ViewerCore::camera_up` the initial 3D view can be configured to fit more naturally with geoscientific data.

Thank you for creating an excellent and easy to use (and integrate) 3D mesh library!

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
Seems fine to me. If CI passes, will merge.

PR #2536, opened 2026-03-25
fix: correct min_t semantics in AABB::intersect_ray

### LINKED ISSUE #2518: Clarifying `min_t` in AABB ray intersections
#### Describe the bug
The description of the `min_t` input parameter in all the different AABB::intersect_ray overloads is "min_t  minimum t value to consider". I take this to mean "we consider intersections that happen along the ray, at times _after_ min_t". However, the opposite seems to be happening; instead, it seems like `min_t` is the _maximum_ time value considered (intersections are considered only if they happen _before_ min_t). 

As an example, the following code

```cpp
#include <igl/AABB.h>

{
  // Inline mesh of a cube
  const Eigen::MatrixXd V= (Eigen::MatrixXd(8,3)<<
    0.0,0.0,0.0,
    0.0,0.0,1.0,
    0.0,1.0,0.0,
    0.0,1.0,1.0,
    1.0,0.0,0.0,
    1.0,0.0,1.0,
    1.0,1.0,0.0,
    1.0,1.0,1.0).finished();
  const Eigen::MatrixXi F = (Eigen::MatrixXi(12,3)<<
    0,6,4,
    0,2,6,
    0,3,2,
    0,1,3,
    2,7,6,
    2,3,7,
    4,6,7,
    4,7,5,
    0,4,5,
    0,5,1,
    1,5,7,
    1,7,3).finished();

  igl::AABB<Eigen::MatrixXd, 3> tree;
  tree.init(V, F);

  Eigen::Vector3d origin(0.5,-1,0.5);
  Eigen::Vector3d direction(0,1,0);

  igl::Hit<double> hit, hit_with_min_t, hit_with_big_min_t;
  bool is_hit = tree.intersect_ray(V, F, origin.transpose(), direction.transpose(), hit);
  std::cout << "Ray intersection: " << (is_hit ? "true" : "false") << std::endl;

  bool is_hit_with_small_min_t = tree.intersect_ray(V, F, origin.transpose(), direction.transpose(), 0.00001, hit_with_min_t);
  std::cout << "Ray intersection: " << (is_hit_with_small_min_t ? "true" : "false") << std::endl;

  bool is_hit_with_big_min_t = tree.intersect_ray(V, F, origin.transpose(), direction.transpose(), 100.0, hit_with_big_min_t);
  std::cout << "Ray intersection: " << (is_hit_with_big_min_t ? "true" : "false") << std::endl;
}
```
outputs
```
Ray intersection: true
Ray intersection: false
Ray intersection: true
```

#### Platform

- [ ] Windows
- [X ] macOS
- [ ] Linux

nmlucarelli:
Fixes #2518 .
**Breaking fix for users who have adapted to the inverted behavior when supplying `min_t`**

Fixes the inverted behavior of `min_t` argument in `AABB::intersect_ray`.

The `min_t` value passed in is used as `t1` when calling `ray_box_intersect` which sets the upper limit of t range considered for intersections. That mixup is part of a larger issue with a mismatch in intended behavior and usage of `min_t` between the public `intersect_ray` and private `intersect_ray_opt` called recursively to do the search and return the nearest hit.

L1686 in AABB.cpp is attempting to back out the nearest hit value with comparisons to `min_t`. That's consistent with the defined behavior for that specific `intersect_ray_opt` overload, but clashes with the expected behavior of `min_t` in the public method `intersect_ray`.

This patch introduces a `max_t` parameter to the `intersect_ray_opt` method. Internally BVH culling is now done with comparisons to `max_t` and the user input of `min_t` is properly passed as the lower limit, `t0`, of the `ray_box_intersect` call. The PR also includes a separate fix for a hardcoded cast to `float` in `ray_triangle_intersect`.

The existing unit test AABB: intersect_tet is updated to pass `min_t = 0` instead of `min_t = std::numeric_limits<double>::infinity()` which follows the old, inverted behavior. A new unit test (AABB: intersect_cube_min_t) is added following the example given in the bug report (#2518)

#### Checklist
- [ ] Adds new .cpp file.
- [ ] This is a minor change.


nmlucarelli:
Just checking in to see if there's anything I need to address before this is ready to merge.

alecjacobson:
I don't think there's anything on your end, I just need to find time to review and don't know when that will be.

fallenmi (changes requested):
Reviewed with OpenAI Codex assistance under the account owner's authorization. I independently verified the exact revisions, reproducer, test outputs, and conclusion.

The new `min_t` value is used to prune each node's bounding box, but the leaf path still returns the raw `ray_mesh_intersect()` result without checking the resulting `hit.t` against `min_t`. A primitive can intersect before the threshold while its non-zero-width AABB continues beyond the threshold, so the box test alone is not sufficient.

On exact head `1a346b7ea74df25836501cfbcf032983f51081e0`, I used one triangle with vertices `(0,0,0)`, `(2,0,2)`, `(0,2,2)`, a ray from `(0.25,0.25,-1)` along `+z`, and `min_t=2`. The primitive hit is `t=1.5`, while its AABB occupies ray interval `[1,3]`; `AABB::intersect_ray()` incorrectly returns `true` with `t=1.5`. An eight-threshold direct-primitive oracle improves from 1/8 matches on the recorded base to 3/8 on the head, but every tested threshold above the hit (`1.500001`, `1.75`, `2`, `2.5`, `3`) still returns the forbidden `t=1.5` hit. The clean merge with current `main`, `635f086e9de0ebbeccdf65a37a756feac58ca135`, has the same 3/8 result.

Please filter the primitive hit against the active lower bound before returning it (using the intended documented boundary semantics), and add a regression where the triangle hit is before `min_t` but the triangle's AABB extends past it. The existing cube test cannot cover this because the intersected face's box is zero-width along the ray. The exact head and current merge both pass all seven `AABB:*` test cases (1,170,450 assertions) and all three `ray_mesh_intersect*` cases (12 assertions), confirming the gap is not covered.

PR #2555, opened 2026-08-31
Return infinity for unreachable exact geodesic targets

### LINKED ISSUE #2554: exact_geodesic returns 0.0 instead of inf for disconnected source and target
```cpp
#include <igl/exact_geodesic.h>

{
  Eigen::MatrixXd V(6,3);
  V<<
    0,0,0,
    1,0,0,
    0,1,0,

    2,0,0,
    3,0,0,
    2,1,0;
  Eigen::MatrixXi F(2,3);
  F<<
    0,1,2,
    3,4,5;
  Eigen::VectorXi VS(1); VS<< 0;
  Eigen::VectorXi VT(1); VS<< 5;
  Eigen::VectorXd D;
  igl::exact_geodesic(
      V,F,VS,Eigen::VectorXi(),VT,Eigen::VectorXi(),D);
  std::cout<<"D: "<<D.transpose()<<std::endl;
}
```

An assertion will fire: https://github.com/libigl/libigl/blob/4d5829f48dcf7c3d3766b52740a73cd42194946e/include/igl/exact_geodesic.cpp#L1067

And the comments around it (in this copy-paste job from googlecode) make it seem like it expected the input to be a single component.

This fix might just be to document this precondition, though returning infs (if it will always be found without crashing) would be better.

SCW5370:
Fixes #2554.

I found that `exact_geodesic` rejected disconnected meshes in Debug builds and treated an unreachable target as distance zero in Release builds because its traceback path was empty.

I removed the connectivity-only verification from the internal mesh check, mapped an empty traceback path to infinity, and documented the unreachable-target behavior. I also added a regression test with one reachable and one unreachable target across two components.

Verification:
- Debug header-only build: `exact_geodesic:*` passed (12 assertions in 2 test cases).
- Release static-library build: all core tests passed (10,605,089 assertions in 237 test cases).
- `git diff --check` passed.

I also attempted the full Debug core suite. It stopped in the unrelated `boundary_facets: single_tri` test on an Eigen comma-initializer assertion after 18 test cases; the focused Debug tests above passed.

The behavior change is limited to disconnected meshes and empty tracebacks. Existing connected-mesh coverage remains unchanged.

#### Checklist

- [ ] Adds new .cpp file.
- [x] This is a minor change.


alecjacobson:
## Review: verified correct, ready to merge

I built and ran the changed code **header-only** (no `IGL_STATIC_LIBRARY`, no prebuilt `libigl.a`) so the PR's modified `.cpp`/`.h` are actually compiled, in both configurations:

**Release** (`-O2 -DNDEBUG`):
```
All tests passed (12 assertions in 2 test cases)
```

**Debug** (`-g`, no `-DNDEBUG`, asserts active):
```
All tests passed (12 assertions in 2 test cases)
```

Both configs pass the new `exact_geodesic: disconnected components` test, which covers a reachable target (`D(0)==1.0`) and an unreachable one across two components (`isinf(D(1))`). I also ran the exact repro from #2554 (3D, source 0 -> target 5) under a Debug build: it now prints `D: inf` with **no assertion firing**, whereas previously it aborted on the connectivity assert.

### Correctness reasoning

- **Empty path <=> unreachable.** In `GeodesicAlgorithmExact::trace_back`, `path` is only left empty when `best_total_distance >= GEODESIC_INF/2.0` (the early `return`). Any reachable target always executes `path.push_back(destination)` first, so a non-empty path is guaranteed even for source==target. Therefore `path.empty()` is an exact indicator of unreachability, and mapping it to `+infinity` is the correct semantic.
- **`+inf` is the right contract** (vs NaN/-1): it matches the library's internal `GEODESIC_INF` convention and the standard geodesic/graph-distance meaning of "no path". It composes correctly under `min`/comparison, unlike NaN or a magic negative.
- **No regression on connected meshes.** The reachable branch is unchanged; `D(0)==1.0` confirms normal distances are unaffected.
- **Removed `verify()` block is safe.** It only asserted single-component edge-connectivity — a false precondition that rejected perfectly valid multi-component meshes. The loose-vertex check (the part that catches genuinely malformed input) is retained. So this removes an incorrect assumption, not a real malformed-mesh guard.

Verdict: **LGTM, approve / merge.**

---
🤖 This review was written by [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8) on behalf of the maintainer.

PR #2556, opened 2026-08-31
igl::triangle::remesh_at_points

alecjacobson:
Adds `igl::triangle::remesh_at_points`: given a set of unique points on a mesh specified by barycentric coordinates and a face index, remesh so those points become vertices.

- Vertex-points are not inserted, only tracked in `K`.
- Edge-points are inserted and the edge is split on *all* incident faces, so non-manifold input stays non-manifold and manifold input stays manifold.
- Face-points are inserted.
- Input may be in 3D (or any dimension); each face is triangulated in its own barycentric coordinate space via `igl::triangle::triangulate`.

Also adds tutorial `1005_SubtriangleGeodesicDistances`, which uses it to remesh at the sample points of an exact geodesic computation.

## Tests

`tests/include/igl/triangle/remesh_at_points.cpp` covers face-points, vertex-points, edge-points on a boundary edge, on a shared manifold edge, and on a non-manifold edge shared by three faces (including two points on the same non-manifold edge, where one incident face traverses it in the opposite direction). Two larger cases: inserting every edge midpoint of a cube, which is exactly one step of 4-1 subdivision, and 2503 interior points spread over the 1000 faces of `decimated-knight.obj`.

A shared checker validates the invariants the header promises:

- `V` is a prefix of `VV`; all of `FF`, `J`, `K` in range
- `VV.row(K(i))` is exactly the barycentric interpolation, and is actually referenced by `FF`
- each original face is *exactly tiled* by its descendants: every output triangle is coplanar with and positively oriented w.r.t. face `J(i)`, and their areas sum to the original face's area — this rules out degenerate, inverted, overlapping and missing triangles in one check

Every input is deterministic. Barycentric coordinates are dyadic rationals so each row sums to exactly `1.0` in floating point, which the test asserts, so a CI failure is reproducible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


alecjacobson:
This is a kind of cheezy way to do it but I didn't want to touch exact_geodesic "implementation" 

image

